PR #3, opened 2020-01-05
add :time module

dsvoronin:


time/src/testFixtures/kotlin/com/avito/time/FakeTimeProvider.kt
```diff
@@ -0,0 +1,20 @@
+package com.avito.time
+
+import java.util.Date
+
+class FakeTimeProvider : TimeProvider {
```
eugene-krivobokov, line 5: Do we need a default constructor with exact time for testing?

eugene-krivobokov (approved):

time/src/testFixtures/kotlin/com/avito/time/FakeTimeProvider.kt
dsvoronin: not until you asked somehow :)

PR #4, opened 2020-01-05
add :git module

dsvoronin:


git/build.gradle.kts
```diff
@@ -0,0 +1,21 @@
...
+
+val kotlinVersion: String by project
+val mockitoJunit5Version: String by project
+
+dependencies {
+    implementation(gradleApi())
+    implementation(project(":kotlin-dsl-support"))
+    // can't use logging due to cyclic dependency
+    implementation(project(":utils"))
+    implementation("org.jetbrains.kotlin:kotlin-stdlib:$kotlinVersion")
```
eugene-krivobokov, line 14: Can we use this property directly from gradle.properties? Is it available implicitly?
I'd like to omit "val <propertyName>: String by project" if possible

eugene-krivobokov (approved):

git/build.gradle.kts
dsvoronin: afaik no

PR #5, opened 2020-01-05
add :okhttp and :test-okhttp modules

dsvoronin:


okhttp/src/main/kotlin/com/avito/http/FallbackInterceptor.kt
```diff
@@ -0,0 +1,52 @@
...
+            chain.doFallback(request)
+        } else {
+            response ?: throw exception!!
+        }
+    }
+
+    private fun isFallbackNeeded(response: Response?): Boolean {
+        return when (response) {
+            null -> true
+            else -> when {
```
eugene-krivobokov: Nested "when" looks overcomplicated for reading.
I'd rather split this check and return earlier.

okhttp/src/main/kotlin/com/avito/http/RequestBodies.kt
```diff
@@ -0,0 +1,50 @@
...
+
+    return if (isPlaintext(buffer)) {
+        buffer.readString(charset!!)
+    } else {
+        null
+    }
+}
+
+/**
+ * Returns true if the body in question probably contains human readable text. Uses a small sample
```
eugene-krivobokov: "in question" ?

eugene-krivobokov (approved):

okhttp/src/main/kotlin/com/avito/http/RequestBodies.kt
dsvoronin: no idea, fixed

okhttp/src/main/kotlin/com/avito/http/FallbackInterceptor.kt
dsvoronin:
```
return when {
            response == null -> true
            response.isSuccessful -> false
            else -> response.code() in doFallbackOnTheseCodes
        }
```
idea's variant, looks readable

PR #152, opened 2020-02-05
prepare publish to bintray, optional artifactoryUrl

dsvoronin:


ci/check_onboarding.sh
```diff
@@ -0,0 +1,15 @@
...
+# shellcheck disable=SC2086
+USER_ID=$(id -u ${USER})
+
+docker run --rm \
+    --volume "$(pwd)":/app \
+    --volume /var/run/docker.sock:/var/run/docker.sock \
+    --workdir /app \
+    --env LOCAL_USER_ID="$USER_ID" \
+    --env GRADLE_USER_HOME=/gradle \
+    dsvoronin/android-builder \
```
sergioboi, line 14: why is your name here?)

sergioboi (commented):
I was waiting this feature for years. Finally it's here!

ci/check_onboarding.sh
dsvoronin: because i haven't registered official avito dockerhub acc yet, 
will do

PR #154, opened 2020-02-07
support test impact analysis for non-default test build type

dsvoronin:
+ opportunistic refactoring


subprojects/gradle/instrumentation-impact-analysis/src/main/kotlin/com/avito/instrumentation/impact/InstrumentationTestImpactAnalysisPlugin.kt
```diff
@@ -20,66 +21,57 @@ class InstrumentationTestImpactAnalysisPlugin : Plugin<Project> {
...
-            AnalyzeTestImpact::class.java,
-            extension,
-            modifiedProjectsFinder,
-            gson
-        ).configure { analyzeTask ->
-            analyzeTask.group = TASK_GROUP
-            analyzeTask.description = "Find tests to run based on changed modules and test code analysis"
-            analyzeTask.dependsOn(bytecodeAnalyzeTask) // TODO https://docs.gradle.org/5.6.4/userguide/lazy_configuration.html#sec:working_with_task_dependencies_in_lazy_properties
-            analyzeTask.apply {
+                dependsOn("${project.path}:compile${it.testBuildType}AndroidTestKotlin")
```
dsvoronin: could be only one `testBuildType`, no need to produce multiple tasks

eugene-krivobokov: is it a build variant or just a build type? should we support build flavors?

eugene-krivobokov (approved):

dsvoronin:
> is it a build variant or just a build type? should we support build flavors?

its just a buildType, but seems like flavors tasks generated anyways, you're right

PR #156, opened 2020-02-07
Publish all modules to bintray [MERGE_WHEN_READY]

dsvoronin:
Gradle plugin portal looks kinda unreliable and unnecessary at this point.

Unreliable because i got bunch of 503 on first try, no retry mechanism, and "dirty" state: some plugins was uploaded some not, and next run stuck because versions are taken. Bad experience, and costly workaround.

Unnecessary, because we can live with module substitution in settings gradle and use plugins { } block as before.
One caveat is: we should keep our module artifactsIds same as plugin Id's , or close to them.
By close i mean that we should able to map all names with pattern.

`com.avito.android.X` -> `com.avito.android:X:version`


build.gradle.kts
```diff
@@ -263,3 +258,28 @@ fun Project.getOptionalExtra(key: String): String? {
         null
     }
 }
+
+fun Project.configureBintray(vararg publications: String) {
+    extensions.findByType<BintrayExtension>()?.run {
+        user = System.getenv("BINTRAY_USER")
+        key = System.getenv("BINTRAY_API_KEY")
```
eugene-krivobokov, line 265: What if I don't have these envs? Will I get an understandable error?

docs/content/Infrastructure.md
```diff
@@ -9,47 +9,92 @@ Monorepo of all tooling to continuously test and deliver apps to users
 
 ## Modules
 
-### Gradle modules
+### Gradle plugins
+
+`build.gradle`
```
eugene-krivobokov, line 14: Before the code, we should give a brief explanation of this procedure.

docs/content/Infrastructure.md
```diff
@@ -89,31 +134,23 @@ Shared modules between android-test and gradle.
...
-environmental argument should be set:
+1. Make sure new project version specified in develop head
+1. Manually run [Teamcity configuration](http://links.k.avito.ru/releaseAvitoTools)
 
-- `BINTRAY_USER`
-- `BINTRAY_API_KEY`
+### Local integration tests with avito
 
-`./gradlew publishToArtifactory`
+1. Choose project version that will not clash with released ones
```
eugene-krivobokov, line 146: It would be great to have an example

docs/content/Infrastructure.md
```diff
@@ -89,31 +134,23 @@ Shared modules between android-test and gradle.
...
+1. Run `./gradlew publishToMavenLocal -PprojectVersion=<Your test version>`
+1. Run integration tests of your choice in avito with specified test version
 
-environmental argument should be set:
+### CI integration tests with avito
```
eugene-krivobokov, line 150: with -> against?

docs/content/Infrastructure.md
```diff
@@ -89,31 +134,23 @@ Shared modules between android-test and gradle.
...
 
-- `ARTIFACTORY_URL`
-- `ARTIFACTORY_USER`
-- `ARTIFACTORY_PASSWORD`
+1. Run [Teamcity configuration](http://links.k.avito.ru/fastCheckIntegration). You can also change build branch if you need to test unmerged code.
+1. Pull request checks of avito would run against generated version of tools project.
 
-To specify a version, use `PROJECT_VERSION` env or `projectVersion` gradle property 
+To re-check this changes locally you need to find android tools version in build parameters. Look for `integrationVersion`.\
```
eugene-krivobokov, line 155:
```suggestion
To re-check these changes locally you need to find android tools version in build parameters. Look for `integrationVersion`.\
```

sergioboi (approved):

subprojects/gradle/instrumentation-test-impact-analysis/build.gradle.kts
```diff
@@ -12,8 +13,8 @@ val kotsonVersion: String by project
 
 dependencies {
     implementation(project(":subprojects:gradle:android"))
+    implementation(project(":subprojects:gradle:impact-shared"))
```
eugene-krivobokov, line 16: Can you please describe changes in a project layout?
It looks like a new convention or a type of module.
Is it just a shared code between gradle plugins or something more complex?

eugene-krivobokov (approved):

build.gradle.kts
dsvoronin: not sure, we should make an extension method for it

subprojects/gradle/instrumentation-test-impact-analysis/build.gradle.kts
dsvoronin: i had to rename `impact` -> `impact-plugin` because of publication quirks
and `impact` to something else
that's the only reason

PR #159, opened 2020-02-09
MBS-7669 Add buildSrc to the tech radar [MERGE_WHEN_READY]

eugene-krivobokov:


sergioboi (approved):

eugene-krivobokov:
Another overview: https://proandroiddev.com/stop-using-gradle-buildsrc-use-composite-builds-instead-3c38ac7a2ab3

PR #164, opened 2020-02-10
built infrastructure now used in test-app ui tests

dsvoronin:


sergioboi:
I gave it 5 min and i didn't realize what was it about

dsvoronin:
> I gave it 5 min and i didn't realize what was it about

We don't use latest version of built infrastructure in `:test-app` module, it is always behind and updated manually without any process

This PR is about reusing built infrastructure in separate gradle command of checking `:test-app` against latest version of infra

-------
updating still manual though

build.sh
```diff
@@ -2,6 +2,11 @@
 
 set -e
 
-source $(dirname $0)/_main.sh "build publishToMavenLocal"
+source $(dirname $0)/_main.sh
+TEMP_PROJECT_VERSION="temp-version"
+
+runInBuilder "${GIT_COMMANDS}
```
eugene-krivobokov: It's easy to forget about `GIT_COMMAND` in a new usage of `runInBuilder`
Can you run it internally always?

ci/docker/android-builder/entrypoint.sh
```diff
@@ -11,8 +11,6 @@ echo "Starting with UID : $USER_ID"
 groupadd --gid "${USER_ID}" ${BUILD_USER}
 useradd --shell /bin/bash --uid "${USER_ID}" --gid "${USER_ID}" --comment "User for container" --create-home ${BUILD_USER}
 
-setfacl -m user:$BUILD_USER:rw /var/run/docker.sock
```
eugene-krivobokov, line 14: It seems we need to rebuild images to apply changes in entrypoint.

gradle.properties
```diff
@@ -12,7 +12,7 @@ syncAndroidModules=true
 # Version is not a day of a month. We don't want to bound releases to days.
 projectVersion=2020.2.6
 # Previous stable version to be used in this project
-infraVersion=2020.2.1
+infraVersion=2020.2.6
```
eugene-krivobokov: Where will we use this version after this PR?

eugene-krivobokov (changes requested):

gradle.properties
dsvoronin: it was released, but never integrated (process still not stable)

dsvoronin: i discovered a bug with incorrect plugin declaration because didn't integrate test version correctly, now we need .7 version

ci/docker/android-builder/entrypoint.sh
dsvoronin: i did it, it points on `:latest` for now

eugene-krivobokov (approved):

PR #165, opened 2020-02-10
Delete overlapping detection. Add in process click docs [MWR]

sergioboi:


docs/content/test/InProcessClick.md
```diff
@@ -0,0 +1,35 @@
...
+
+Espresso click use IPC. It leads to problems:
+- Flakiness clicking on moving objects. Because click takes some time to touch after coordinates calculating.
+- "misinterpret clicks as long clicks"
+
+## Implementation details
+
+Click happens via dispatch touch event on root view
+
+### Knowing issues
```
dsvoronin: known

docs/content/test/InProcessClick.md
```diff
@@ -0,0 +1,35 @@
+---
+title: In process click
+type: docs
+---
+
+# In process click
+
+It is click without IPC.
```
dsvoronin: IPC
It is click without IPC.
??

Screenshot 2020-02-10 at 16 50 23

docs/content/test/InProcessClick.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Motivation
+
+Espresso click use IPC. It leads to problems:
+- Flakiness clicking on moving objects. Because click takes some time to touch after coordinates calculating.
```
dsvoronin: `Flaky clicks on moving views?`

docs/content/test/InProcessClick.md
```diff
@@ -0,0 +1,35 @@
+---
+title: In process click
+type: docs
+---
+
+# In process click
+
+It is click without IPC.
+
+## Motivation
```
dsvoronin: Motivation for what?
If it's about our motivation to implement custom click logic(that's what i'd expect), then this section body has only motivation why standard one's are problematic, and no conclusion

docs/content/test/InProcessClick.md
```diff
@@ -0,0 +1,35 @@
...
+
+### Knowing issues
+
+- Can click through any system elements on the screen. It applies clicks directly on root
+ view of our application. Because of it, crash or permission dialogs can be ignored by
+ tests.
+- Can click through separate decor view of our application. Sometimes we have multiple
+ decor view in application (for example, when we have toolbar overflow menu). And that
+ kind of click implementation can click through it.
+- When clicked view is overlapped we can't click it. But now we haven't error. Click behaves like a successful.
```
dsvoronin: `But now we haven't error` - haven't what? missed verb

dsvoronin: `we can't click, but we can, but without error` - this is misleading

dsvoronin (dismissed):

docs/content/menu/index.md
```diff
@@ -43,6 +43,7 @@ headless: true
     - [Test minimized]({{< ref "/test/TestMinimized.md" >}})
     - [Troubleshooting]({{< ref "/test/TroubleshootingUI.md" >}})
     - [Кастомный запуск]({{< ref "/test/DynamicConfig.md" >}})
+    - [Кастомный клик]({{< ref "/test/InProcessClick.md" >}})
```
eugene-krivobokov, line 46: no one is interested in implementation details on this level in menu.
Let's wrap it in more generic structure:

```
Testing
  ...
  - Custom running
  - Actions
```

docs/content/test/InProcessClick.md
```diff
@@ -0,0 +1,35 @@
...
+- Flaky clicks on moving views. Because click takes some time to touch after coordinates calculating.
+- "misinterpret clicks as long clicks"
+
+## Implementation details
+
+Click happens via dispatch touch event on root view
+
+### Known issues
+
+- Can click through any system elements on the screen. It applies clicks directly on root
```
eugene-krivobokov, line 22: Can we show this in a test?
I'm worried about duplicating implementation details in the documentation.

docs/content/test/InProcessClick.md
```diff
@@ -0,0 +1,35 @@
+---
+title: In process click
+type: docs
+---
+
+# In process click
```
eugene-krivobokov, line 6: We shouldn't start from this low level implementation details.
For starters, we should explain briefly that we have click action and it is a wrapper above preconditions, click, ...
Only after that, we could explain these details.

You can add it to the first paragraph and restructure the document.

eugene-krivobokov (changes requested):

docs/content/menu/index.md
eugene-krivobokov: Generally, it's better to start from common pages and split them into multiple when they come too big.
I'm thinking of other actions - swipe, text, ... For the beginning, one page for all actions looks enough.

docs/content/test/InProcessClick.md
sergioboi: I don't know. I tired of this task. Everything i do is move text from source to documentation.

docs/content/menu/index.md
sergioboi: I agree that nobody except us need those details. But i don't understand where do you want to see that page.

eugene-krivobokov: I suggest to copy structure of documentation from Espresso:
https://developer.android.com/training/testing/espresso/basics

When it becomes bigger, we will split pages.

eugene-krivobokov (approved):

PR #174, opened 2020-02-14
ability to disable performance step

sergioboi:
Discussed with @eugene-krivobokov 
Root task describes a problem to suppress fails on build on target. But cheeper to support disable whole step. It's give us opportunity to win all problems and more clear expectation


subprojects/gradle/cd/src/main/kotlin/com/avito/ci/steps/PerformanceTestCheck.kt
```diff
@@ -14,8 +14,11 @@ open class PerformanceTestCheck(context: String) : SuppressibleBuildStep(context
 
     var configuration: String = ""
 
+    var enabled = true
```
eugene-krivobokov, line 17: We also need documentation about this feature.

eugene-krivobokov (changes requested):

subprojects/gradle/cd/src/main/kotlin/com/avito/ci/steps/PerformanceTestCheck.kt
```diff
@@ -14,8 +14,11 @@ open class PerformanceTestCheck(context: String) : SuppressibleBuildStep(context
...
+
     override fun registerTask(project: Project, rootTask: TaskProvider<out Task>) {
         if (useImpactAnalysis && !project.internalModule.isModified()) return
+        if (!enabled) return
```
eugene-krivobokov, line 20: it's a bit cheaper to check this flag first.

eugene-krivobokov (approved):

PR #175, opened 2020-02-14
MBS-7688 Fail on first unexpected error in retry

eugene-krivobokov:


subprojects/android-test/test-app/src/androidTest/kotlin/com/avito/android/ui/test/VisibilityTest.kt
```diff
@@ -30,7 +29,6 @@ class VisibilityTest {
         rule.runOnUiThread {
             rule.activity.findViewById<View>(R.id.text).visibility = View.INVISIBLE
         }
-        exception.expect(UITestFrameworkException::class.java)
```
eugene-krivobokov, line 33: It doesn't matter

subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/runner/InHouseInstrumentationTestRunner.kt
```diff
@@ -268,13 +268,6 @@ abstract class InHouseInstrumentationTestRunner :
             }
 
             onWaiterRetry = { }
-
-            /**
-             * Большинство ексепшенов от espresso оборачивается в UITestFrameworkException
-             */
-            waiterAllowedExceptions = waiterAllowedExceptions.plus(
-                UITestFrameworkException::class.java
-            )
```
eugene-krivobokov, line 320: We lost any unknown error in retry and go into infinite retry.

"unknown" = else case in `ReportFriendlyFailureHandler.handle`

Example:

- Test: `click()` 

It runs in retry loop
- Activity: 
  - handles click
  - startActivity() -> fail ! (This is the original error)
- Espresso failure handler (`ReportFriendlyFailureHandler`) wraps any unknown error to `UITestFrameworkException`
- Retry loop skips it as allowed

We go into the second loop, but we are in another state on the screen. 
The original error is lost. Happy debugging!

eugene-krivobokov: @elya-a, it was the cause of the lost error

sergioboi (approved):

PR #178, opened 2020-02-14
use nebula integration tests plugin to separate slack integration tests

dsvoronin:
These tests depends on slack env variables
now we're able to run whole unit tests suite locally

https://plugins.gradle.org/plugin/nebula.integtest
https://docs.gradle.org/current/userguide/java_testing.html#sec:configuring_java_integration_tests


subprojects/gradle/slack/build.gradle.kts
```diff
@@ -3,6 +3,7 @@ plugins {
     id("java-test-fixtures")
     `maven-publish`
     id("com.jfrog.bintray")
+    id("nebula.integtest")
```
eugene-krivobokov, line 6: Can you please point to documentation or examples?
I haven't got it from https://github.com/nebula-plugins/nebula-project-plugin#nebula-integtest-plugin

eugene-krivobokov (approved):

subprojects/gradle/slack/build.gradle.kts
dsvoronin: it is only in provided links (nebula plugin and official docs with concept)

PR #180, opened 2020-02-15
MBS-7669 Fix lost search engine in docs [MERGE]

eugene-krivobokov:
Add a lost file for searching in docs.

Also, I've tried to change search results but it requires more frontend skills.
https://github.com/nextapps-de/flexsearch (see themes/assets/search...).


dsvoronin (approved):

docs/local.sh
```diff
@@ -33,4 +33,4 @@ docker run --rm \
         -w="/app" \
         -p 1313:1313 \
         android/docs/local:latest \
-        sh -c "hugo server --cleanDestinationDir --i18n-warnings --minify --theme book --bind 0.0.0.0"
+        sh -c "hugo server --cleanDestinationDir --i18n-warnings --minify --bind 0.0.0.0 --baseURL http://localhost:1313/avito-android"
```
dsvoronin, line 36: `--theme book` - how it will work without specified theme?

dsvoronin (approved):

docs/local.sh
eugene-krivobokov: The theme is already defined in config
https://github.com/avito-tech/avito-android/blob/develop/docs/config.toml#L12

PR #182, opened 2020-02-16
in-house-runner is highly tight with implementation details.

sergioboi:
It challenges runner reuse

This is first step. Remove logically wrong files


subprojects/android-test/ui-testing-core/build.gradle.kts
```diff
@@ -24,8 +25,10 @@ dependencies {
     api("androidx.appcompat:appcompat:$androidXVersion")
     api("androidx.recyclerview:recyclerview:$androidXVersion")
     api("com.google.android.material:material:$androidXVersion")
+    api("com.google.android.gms:play-services-base:17.0.0")
```
dsvoronin, line 28: why? it looks wrong to have such dependency in core ui testing library
you should create new module

dsvoronin (changes requested):

subprojects/android-test/ui-testing-core/build.gradle.kts
sergioboi: It needs for PlayServices file. checkPlayServices is called from AbstractLaunchRule.
We already have had problems with dependency leak to inhouse from ui-testing-maps.
Now my goal is clear in-house module after we should refactor whole ui-testing-core

dsvoronin: it is better to have a mess in inhouse than in ui-testing core
make a module with this check and add dependency in inhouse, or even in its implementations.
better in inhouse, because we want to control that every client uses it.
to make it truly optional we need to implement a new way to make sure it is used

sergioboi: I don't understand intention of your message. 
Where is a problem? We had playservices check AbstractLaunchRule and didn't have direct dependency. In future we can discuss where we need AbstractLaunchRule.
I don't understand what we could get with AbstractLaunchRule in Inhouse runner.
Now i'm use principle if you use something u need it in your dependency.

dsvoronin: one could use ui testing core without inhouse runner, why core comes with play services dependency?

sergioboi: because it needs PlayServices check.
I agree with that in future we should extract google deps to own module. 'ui-google-rule' for example. But now my focus on clearing inhouse runner.

dsvoronin: but ui testing core doesn't use play services as a dependency, you introduced it with this PR

sergioboi: but inhouse runner don't use play services but has a dependency.
I can do new module if u want.

PR #188, opened 2020-02-17
fixed package name

DEADMC:
Pulling was broken, because I had to use testVariant.applicationId instead of just applicationId. Screenshots were requested from unexisting folder and that led pulling to fail

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -10,7 +10,7 @@ syncAndroidModules=true
 # Artifacts' versions
 # Scheme: year.month[.<version>]
 # Version is not a day of a month. We don't want to bound releases to days.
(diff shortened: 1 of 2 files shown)


gradle.properties
```diff
@@ -10,7 +10,7 @@ syncAndroidModules=true
 # Artifacts' versions
 # Scheme: year.month[.<version>]
 # Version is not a day of a month. We don't want to bound releases to days.
-projectVersion=2020.2.7
+projectVersion=2020.2.8
```
sergioboi: IT'IS UNACCEPTEABLE

sergioboi (approved):

PR #189, opened 2020-02-17
fix api breaking changes

sergioboi:


sboishtyan:
https://tmct.avito.ru/viewLog.html?buildId=13211339&tab=queuedBuildOverviewTab

dsvoronin:
> https://tmct.avito.ru/viewLog.html?buildId=13211339&tab=queuedBuildOverviewTab

let's not publish internal links, even if we already got one on status

dsvoronin (approved):
i think we should also consider simple static logger

PR #190, opened 2020-02-17
gradlew 6.2

dsvoronin:
https://docs.gradle.org/6.2/release-notes.html

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -11,10 +11,6 @@ import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 
 buildscript {
     val r8Version: String by project
-
-    repositories {
-        maven { setUrl("http://storage.googleapis.com/r8-releases/raw") }
-    }
     dependencies {
         classpath("com.android.tools:r8:$r8Version")
     }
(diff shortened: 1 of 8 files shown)


sergioboi (approved):

build.gradle.kts
```diff
@@ -54,7 +50,27 @@ subprojects {
 
     repositories {
         jcenter()
-        google()
+        exclusiveContent {
+            forRepository {
+                google()
+            }
+            filter {
+                includeModuleByRegex("com\\.android.*", "(?!r8).*")
```
eugene-krivobokov, line 58: can we use `excludeModule` for r8 to avoid regexp?

eugene-krivobokov (approved):

build.gradle.kts
dsvoronin: yep, forgot about this one. thanks

dsvoronin:
```
    ExclusiveContentRepository filter(Action<? super InclusiveRepositoryContentDescriptor> config);
```

only inclusive one available here :confused: 
i understand the intent, but seems like non necessary limit

PR #193, opened 2020-02-18
validate instrumentation configuration that can break build

sergioboi:


subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/configuration/InstrumentationConfiguration.kt
```diff
@@ -66,6 +66,8 @@ open class InstrumentationConfiguration(val name: String) {
     }
 
     fun data(parentInstrumentationParameters: InstrumentationParameters): Data {
+        require(kubernetesNamespace.isNotBlank())
```
dsvoronin: `InstrumentationConfiguration` has special `fun validate()`, can this check go there?

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/configuration/InstrumentationParameters.kt
```diff
@@ -7,12 +7,18 @@ data class InstrumentationParameters(
     val initialParameters: Map<String, String> = emptyMap()
 ) : Map<String, String> by initialParameters, Serializable {
 
-    fun applyParameters(newParameters: Map<String, String>): InstrumentationParameters =
-        InstrumentationParameters(
+    fun applyParameters(newParameters: Map<String, String>): InstrumentationParameters {
+        newParameters.forEach { key, value ->
```
dsvoronin: extracted precondition function looks better here

dsvoronin: it's not clear to me that instrumentation args are immutable from this point,
i'd expect check as close as possible to runner, but before any reservation job (to prevent resource waste)
but fail fast is also good idea. 
Maybe both? two checks?

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/configuration/target/TargetConfiguration.kt
```diff
@@ -40,17 +40,21 @@ open class TargetConfiguration(val name: String) : Serializable {
...
-        rerunReservation = rerunScheduling?.data()?.reservation ?: scheduling.data().reservation,
-        deviceName = deviceName,
-        instrumentationParams = parentInstrumentationParameters
-            .applyParameters(instrumentationParams)
-            .applyParameters(
-                mapOf("deviceName" to deviceName)
-            )
-    )
+    fun data(parentInstrumentationParameters: InstrumentationParameters): Data {
+        require(deviceName.isNotBlank())
```
dsvoronin: `fun validate()`?

subprojects/gradle/runner/service/src/main/kotlin/com/avito/runner/service/worker/device/adb/AdbDevice.kt
```diff
@@ -272,8 +272,8 @@ data class AdbDevice(
             command = listOf(
                 "am",
                 "instrument",
-                "-w",
-                "-r",
+                "-w", // wait for instrumentation to finish before returning.  Required for test runners.
+                "-r", // raw mode need for parsing
```
dsvoronin: 👍 for comments

`raw mode (is required/is necessary) for further output parsing`

dsvoronin (commented):
added some thoughts, take a look pls

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/configuration/InstrumentationParameters.kt
sergioboi: I don't want to know about it in many places.
My view is we have a structure InstrumentationParameters and it will guarantee that parameters isn't blank. If we had problems in future with incorrect reusing we will fix this.
Now is constructor parameters validation is missing i will add it

dsvoronin: we modify params once more at least here : https://github.com/avito-tech/avito-android/blob/12b0fdff3962329c8814112a726abd24efc15d3d/subprojects/gradle/runner/service/src/main/kotlin/com/avito/runner/service/worker/device/adb/AdbDevice.kt#L80
and it's already too late to prevent resources reservation

But i got your point, it's ok enough for now

dsvoronin (approved):

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/configuration/target/TargetConfiguration.kt
```diff
@@ -40,25 +40,25 @@ open class TargetConfiguration(val name: String) : Serializable {
...
-        reservation = scheduling.data().reservation,
-        rerunReservation = rerunScheduling?.data()?.reservation ?: scheduling.data().reservation,
-        deviceName = deviceName,
-        instrumentationParams = parentInstrumentationParameters
-            .applyParameters(instrumentationParams)
-            .applyParameters(
-                mapOf("deviceName" to deviceName)
-            )
-    )
+    fun data(parentInstrumentationParameters: InstrumentationParameters): Data {
```
YapTeam: Было же красивше?

PR #196, opened 2020-02-18
Use read only dependencies cache

dsvoronin:
https://docs.gradle.org/6.2/userguide/dependency_resolution.html#sub:ephemeral-ci-cache

- [x] push new docker image (we need versioning now)
- [x] describe in docs


_main.sh
```diff
@@ -33,68 +23,84 @@ function clearDockerContainers() {
...
-clearGradleLockFiles
+# https://docs.gradle.org/6.2/userguide/dependency_resolution.html#sub:ephemeral-ci-cache
+MARKER_FILE="$GRADLE_CACHE_DIR"/marker.avito
+MOUNT_OPTIONS=""
+# in container
+GRADLE_RO_DEP_CACHE_DEST=/gradle/readonlydepcache
+function resolveDependenciesCacheOptions() {
+    if [ ! -f "$MARKER_FILE" ]; then
+        echo "[Gradle] Marker file not found. Mounting writeable dependencies cache."
+        MOUNT_OPTIONS+="--volume ${GRADLE_CACHE_DIR}:/gradle/caches "
```
eugene-krivobokov: Why do we use the whole /caches, but not `/caches/modules-<version>`? 
https://docs.gradle.org/6.2/userguide/dependency_resolution.html#sub:cache_copy

_main.sh
```diff
@@ -33,68 +23,84 @@ function clearDockerContainers() {
     fi
 }
 
-GRADLE_CACHE_DIR=$HOME/.gradle/caches
-GRADLE_WRAPPER_DIR=$HOME/.gradle/wrapper
-
-clearDockerContainers
-clearGradleLockFiles
+# https://docs.gradle.org/6.2/userguide/dependency_resolution.html#sub:ephemeral-ci-cache
+MARKER_FILE="$GRADLE_CACHE_DIR"/marker.avito
```
eugene-krivobokov, line 27: marker for what? can we add some meaning to the name?

_main.sh
```diff
@@ -33,68 +23,84 @@ function clearDockerContainers() {
...
+MOUNT_OPTIONS=""
+# in container
+GRADLE_RO_DEP_CACHE_DEST=/gradle/readonlydepcache
```
eugene-krivobokov, line 30: Just a notice

I'm thinking of prefixes for all paths. We have two options:
- In build agent
- In image

When I read the source, I always have to check whether it is inside an image or not.

subprojects/gradle/kotlin-dsl-support/src/main/kotlin/com/avito/kotlin/dsl/ProjectExtensions.kt
```diff
@@ -5,19 +5,15 @@ import kotlin.reflect.KProperty
...
 }
 
-fun Project.getOptionalStringProperty(name: String, default: String): String {
-    return if (hasProperty(name)) {
-        property(name)?.toString() ?: default
-    } else {
-        default
-    }
-}
+fun Project.getOptionalStringProperty(name: String, default: String): String =
```
sergioboi: can we have one function with default value for default parameter

sergioboi:
Can you make short description about that pr or put link to task?

subprojects/gradle/kotlin-dsl-support/src/main/kotlin/com/avito/kotlin/dsl/ProjectExtensions.kt
dsvoronin: nullable with default? sure. but it would be a breaking change.

dsvoronin:
yes, i will add an explanation in docs

subprojects/gradle/kotlin-dsl-support/src/main/kotlin/com/avito/kotlin/dsl/ProjectExtensions.kt
dsvoronin: Nope, we can't actually

We can't express not-null result for this function, and it seems have a great value

_main.sh
eugene-krivobokov: Discussed. It is used for simpler and more robust checks in bash

_main.sh
eugene-krivobokov: Discussed. It's a Gradle's ENV

_main.sh
eugene-krivobokov: @dsvoronin , ?

docs/content/menu/index.md
```diff
@@ -57,6 +57,7 @@ headless: true
     - [Custom Gradle Plugins]({{< ref "/ci/CustomGradlePlugins.md" >}})
     - [Impact analysis]({{< ref "/ci/ImpactAnalysis.md" >}})
     - [Containers]({{< ref "/ci/Containers.md" >}})
+    - [Dependencies cache]({{< ref "/ci/DependenciesCache.md" >}})
```
eugene-krivobokov: This menu becomes a blog :thinking:

docs/content/ci/DependenciesCache.md
```diff
@@ -0,0 +1,38 @@
+---
+title: Dependencies cache
+type: docs
+---
```
eugene-krivobokov: Please, check the text in https://www.grammarly.com/

eugene-krivobokov (changes requested):

_main.sh
dsvoronin: i'll check, it seems you're right

docs/content/menu/index.md
int02h: Maybe we should assemble this file automatically

eugene-krivobokov:
https://github.com/gradle/gradle/releases/tag/v6.2.1
contains a bugfix for read-only cache

eugene-krivobokov (approved):

PR #202, opened 2020-02-20
Add documentation about mocking User-Agent

int02h:


fat-fellow (approved):

docs/content/test_framework/Mocking.md
```diff
@@ -39,7 +39,18 @@ locationRule.setLocation(getTestLocation())
...
+### Mocking User-Agent
+
+It is possible to mock User-Agent for test suite or for specific test with `@UserAgent` annotation. Here's an example:
+
+```kotlin
+@UserAgent("AVITO 59.0 (Android 8.1.0; ru)")
+class MyAwesomeTests {}
+```
+
+This annotation is available in all types of tests but make sense the mostly for functional tests.
```
dsvoronin: `sense the mostly` - odd `the`

dsvoronin: `tests, but`

docs/content/test_framework/Mocking.md
```diff
@@ -39,7 +39,18 @@ locationRule.setLocation(getTestLocation())
...
 in `BaseGeoProvider`. That is technically possible that device location providers would give better location that was mocked. 
-To avoid that mock location with high accuracy. 
+To avoid that mock location with high accuracy.
+
+### Mocking User-Agent
+
+It is possible to mock User-Agent for test suite or for specific test with `@UserAgent` annotation. Here's an example:
+
+```kotlin
+@UserAgent("AVITO 59.0 (Android 8.1.0; ru)")
```
sergioboi, line 49: @dsvoronin why user-agent isn't rule?
Do i need to know user-agenet implementation details to set it?

docs/content/test_framework/Mocking.md
int02h: Actually, not comma needed before but - https://www.grammarly.com/blog/comma-before-but/

int02h: 'the' fixed

docs/content/test_framework/Mocking.md
dsvoronin: 1. we can parse it as a rule now, but should we?
it's a static data and i can't imagine actionable functions on rule for it,
or `@After` cleanup

2. Yes, you need to know implementation details
**good thing**: its generic as a string, you can use this rule on any app; (it's parsing should be implemented on client side btw, so not so generic)
**bad thing**: you can't introduce some function with internal app user agent rules, like `fun createUserAgent(someStronglyTypedArgsWithValidation): String`, because annotations only accepts static things

if bad thing is unacceptable, we should go from annotation to rule or some other implementation

sergioboi: Rule is more predictable. It has lifecycle tight to test. Rule is more expressive.
Custom annotation is a magic.
In future i will prefer rule to annotation

int02h: Create a task to migrate annotation to rule - MBS-7746 (internal task, just for the record)

dsvoronin: Rule is in fact a bit more magical in this case:

- we parse test name
- creating Class instance via reflection
- getting declared fields (looking for Rules)
- going through declared resolvers in runner
- finding match and parsing user agent value from it (oops, we don't have non-static value here)
- ... so its impossible that way
- then we need to put parsed value to instrumentation args bundle
- get it from TestApp and pass to dagger graph as actual user agent

dsvoronin: So we need to inject new user agent value dynamically during test. App should support this.

docs/content/test_framework/Mocking.md
```diff
@@ -39,7 +39,18 @@ locationRule.setLocation(getTestLocation())
 
 **KEEP IN MIND** that above rule does not mock location for the whole device but rather replace the best known location
 in `BaseGeoProvider`. That is technically possible that device location providers would give better location that was mocked. 
-To avoid that mock location with high accuracy. 
+To avoid that mock location with high accuracy.
+
+### Mocking User-Agent
+
+It is possible to mock User-Agent for test suite or for specific test with `@UserAgent` annotation. Here's an example:
```
eugene-krivobokov, line 46: I prefer to get rid of introductory words before examples. Examples are self-explanatory.

... for specific test with `@UserAgent` annotation:

```kotlin
@UserAgent(...)
```

PR #205, opened 2020-02-20
update release steps

sergioboi:


docs/content/Infrastructure.md
```diff
@@ -145,8 +145,8 @@ is version for future release.
 
 [Bintray project](https://bintray.com/avito-tech/maven/avito-android), mirroring to jcenter
 
-1. Make sure integration tests passed via `CI integration tests against avito`
-1. Make sure new project version specified in develop head
+1. Checkout branch with name equally to `projectVersion`
```
dsvoronin, line 148: why is that? just for convention? nothing seems rely on it for now

dsvoronin (approved):

docs/content/Infrastructure.md
eugene-krivobokov: @sboishtyan , ?

PR #206, opened 2020-02-20
use assumptions to ignore slack integration tests without env params

dsvoronin:
enabler for local build.sh run


subprojects/gradle/slack/src/integTest/kotlin/com/avito/slack/SlackConditionalSenderIntegrationTest.kt
```diff
@@ -87,4 +97,9 @@ internal class SlackConditionalSenderIntegrationTest {
             emoji = ":crazy-robot:"
         )
     ).onFailure { it.printStackTrace() }
+
+    private fun String?.assumeNotNullOrBlank(): String {
+        Assumptions.assumeFalse { isNullOrBlank() }
```
eugene-krivobokov, line 102: It looks like a half measure.
What are we fixing? Local checks?

1. Such optimizations give a false sense of safety in local runs.
2. We can't do a fail-fast check in CI. 
Imagine a situation when we lose these ENVs in CI. Tests are green but we expected a failure.

Let's choose a more explicit way.

eugene-krivobokov (changes requested):

subprojects/gradle/slack/src/integTest/kotlin/com/avito/slack/SlackConditionalSenderIntegrationTest.kt
dsvoronin: any suggestions?

dsvoronin: tests will be ignored, not green.
We can forbid ignored tests at all in CI

eugene-krivobokov: > We can forbid ignored tests at all in CI

In this case, we need to distinguish explicitly `@Disabled` tests from implicitly skipped by assumptions.

I propose another solution, to pass an intention through parameters, e.g.:

- Default value in CI properties: `SLACK_INTEGRATION_TESTS` == enabled -> requires other parameters (url, token, ...).
- Default value in the project: `SLACK_INTEGRATION_TESTS` == disabled -> skip tests by assumptions
- Else (unset or unknown value) -> fail fast, require to pass the intention. 

We use a similar pattern in gradle plugins. 
It protects us from accidentally lost parameters. 

Also, I like to separate concerns when feature enabled and not configured properly.
If I don't have SLACK_TEST_CHANNEL it shouldn't mean that I don't want to run tests. It's harder to understand and triage.

subprojects/gradle/slack/src/integTest/kotlin/com/avito/slack/SlackConditionalSenderIntegrationTest.kt
```diff
@@ -6,16 +6,26 @@ import com.avito.slack.model.SlackSendMessageRequest
...
-    private val testChannel = SlackChannel(requireNotNull(System.getenv("SLACK_TEST_CHANNEL")))
-    private val testToken = requireNotNull(System.getenv("SLACK_TEST_TOKEN"))
-    private val slackClient: SlackClient =
-        SlackClient.Impl(testToken, requireNotNull(System.getenv("SLACK_TEST_WORKSPACE")))
+    private val testChannel by lazy {
+        System.getenv("SLACK_TEST_CHANNEL").assumeNotNullOrBlank().let { SlackChannel(it) }
+    }
+
+    private val testToken by lazy {
+        System.getenv("SLACK_TEST_TOKEN").assumeNotNullOrBlank()
```
sergioboi, line 20: i think we must mark it Depreceted and try to pass parameters via constructor

eugene-krivobokov: Do you know another way to pass a value to a unit-test?

sergioboi: Constructor. We can abstract env logic from this class becuase it is dependency details leak

eugene-krivobokov: Constructor of what exactly? 
I don't see a difference. We need to get value for tests.

sergioboi: Sorry i don' understand. Why do we need env for test?
I want hide token impl details.

sergioboi (changes requested):

subprojects/gradle/slack/src/integTest/kotlin/com/avito/slack/SlackConditionalSenderIntegrationTest.kt
dsvoronin: it's an integration test. we need slack hook token to execute it

sergioboi: Can we get slack data not from environment?

dsvoronin: not much of a choice: either env or gradle param

sergioboi (approved):

PR #211, opened 2020-02-26
MBS 7776 Clarify errors in configuration

eugene-krivobokov:


subprojects/gradle/android/src/main/kotlin/com/avito/android/AndroidSdk.kt
```diff
@@ -40,7 +40,7 @@ class AndroidSdk(
     private val buildToolsVersion: String
         get() {
             val value = System.getProperty("buildToolsVersion")
-            require(value.hasContent()) { "buildToolsVersion property should be set" }
+            require(value.hasContent()) { "buildToolsVersion env must be set" }
```
eugene-krivobokov, line 43: We need a strong obligation

dsvoronin (approved):
our own top level require functions will serve well in this case

PR #229, opened 2020-03-05
prepare to fix empty string property default behaviour

dsvoronin:
empty string value is a valid for getMandatoryStringProperty
and don't trigger default value usage in getOptionalStringProperty be default

it's about to change for more consistent behavior;
it will be done in two stages, first one in this PR,
calls in avito scripts will specify allowBlank/nullIfBlank explicitly
and then we'll change default behavior: empty string will be read as missing value

### DIFF
--- a/subprojects/gradle/kotlin-dsl-support/src/main/kotlin/com/avito/kotlin/dsl/ProjectExtensions.kt
+++ b/subprojects/gradle/kotlin-dsl-support/src/main/kotlin/com/avito/kotlin/dsl/ProjectExtensions.kt
@@ -25,9 +25,22 @@ fun Project.getOptionalStringProperty(name: String, default: String, defaultIfBl
 /**
  * @param allowBlank todo false by default
  */
-fun Project.getMandatoryStringProperty(name: String, allowBlank: Boolean = true): String =
-    getOptionalStringProperty(name, nullIfBlank = !allowBlank)
-        ?: throw RuntimeException("Parameter: $name is required (must be not empty)")
+fun Project.getMandatoryStringProperty(name: String, allowBlank: Boolean = true): String {
+    return if (hasProperty(name)) {
+        val string = property(name)?.toString()
+        if (string.isNullOrBlank()) {
+            if (allowBlank) {
+                ""
+            } else {
+                throw RuntimeException("Parameter: $name is blank but required")
+            }
+        } else {
+            string
+        }
+    } else {
+        throw RuntimeException("Parameter: $name is missing but required")
+    }
+}
 
 fun Project.getOptionalIntProperty(name: String): Int? =
     try {
--- a/subprojects/gradle/kotlin-dsl-support/src/main/kotlin/com/avito/kotlin/dsl/ProjectExtensions.kt
+++ b/subprojects/gradle/kotlin-dsl-support/src/main/kotlin/com/avito/kotlin/dsl/ProjectExtensions.kt
@@ -8,7 +8,7 @@ import kotlin.reflect.KProperty
 //TODO these methods should return Property<X>
 
 /**
- * @param nullIfBlank there may be cases when it's ok to pass empty string to override
+ * @param nullIfBlank we accept cases when user passes empty string to override
  * todo true by default, false to not break anything that rely on previous behavior
  */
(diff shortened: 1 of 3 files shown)


subprojects/gradle/kotlin-dsl-support/src/main/kotlin/com/avito/kotlin/dsl/ProjectExtensions.kt
```diff
@@ -5,56 +5,58 @@ import org.gradle.api.file.RegularFileProperty
...
-    getOptionalStringProperty(name)?.toInt()
-} catch (e: NumberFormatException) {
-    null
-}
+/**
+ * @param allowBlank todo false by default
+ */
+fun Project.getMandatoryStringProperty(name: String, allowBlank: Boolean = true): String =
+    getOptionalStringProperty(name, nullIfBlank = !allowBlank)
+        ?: throw RuntimeException("Parameter: $name is required (must be not empty)")
```
eugene-krivobokov: It's better to distinquish two different errors:

- "Parameter: $name is missing but required" <--- hasProperty == false
- "Parameter: $name is blank but required" <--- blank value

It'd easier to debug errors.

subprojects/gradle/kotlin-dsl-support/src/main/kotlin/com/avito/kotlin/dsl/ProjectExtensions.kt
```diff
@@ -5,56 +5,58 @@ import org.gradle.api.file.RegularFileProperty
...
+//TODO these methods should return Property<X>
 
-fun Project.getOptionalStringProperty(name: String, default: String): String {
-    return if (hasProperty(name)) {
-        property(name)?.toString() ?: default
+/**
+ * @param nullIfBlank there may be cases when it's ok to pass empty string to override
+ * todo true by default, false to not break anything that rely on previous behavior
+ */
+fun Project.getOptionalStringProperty(name: String, nullIfBlank: Boolean = false): String? =
```
eugene-krivobokov, line 14: How about another solution for blank values?

Now we have to add additional "nullIfBlank" logic and responsibility to all optional getters. But we can compose required behavior.

`project.getOptionalStringProperty("property", nullIfBlank = true)`
vs
`project.getOptionalStringProperty("property").orNull()` 
or `.nullIfBlank()`

In the second case, the client can choose any additional logic to handle the blank value.

eugene-krivobokov (changes requested):

subprojects/gradle/kotlin-dsl-support/src/main/kotlin/com/avito/kotlin/dsl/ProjectExtensions.kt
```diff
@@ -5,56 +5,58 @@ import org.gradle.api.file.RegularFileProperty
...
-    }
-}
+//TODO these methods should return Property<X>
 
-fun Project.getOptionalStringProperty(name: String, default: String): String {
-    return if (hasProperty(name)) {
-        property(name)?.toString() ?: default
+/**
+ * @param nullIfBlank there may be cases when it's ok to pass empty string to override
+ * todo true by default, false to not break anything that rely on previous behavior
```
sergioboi:
```suggestion
 * @param nullIfBlank we accept cases when user passes empty string to override
 * todo true by default, false to not break anything that rely on previous behavior
```

subprojects/gradle/kotlin-dsl-support/src/main/kotlin/com/avito/kotlin/dsl/ProjectExtensions.kt
dsvoronin: i really like this idea, but i see a problem with it: i need opposite default: 
`project.getMandatoryStringProperty("property").allowBlank()` - and can't express it with property syntax, because exception must be thrown in get function

eugene-krivobokov: I see. It may be possible when these functions will return a custom `Property<T>`

eugene-krivobokov (approved):

PR #231, opened 2020-03-06
Mbs 7535 Document network mocks

eugene-krivobokov:


docs/content/docs/test_framework/Mocking.md
```diff
@@ -5,11 +5,73 @@ type: docs
 
 # Mocking in tests
 
+To use a specific state you need to find a suitable `@Rule` or annotation.\
+Thus, test runner knows how to prepare it for the test ([more details]({{< relref "#writing-custom-mocks-for-tests">}})).
```
dsvoronin, line 9: i have a feeling that `#writing-custom-mocks-for-tests` should be on top of this page

dsvoronin (approved):

docs/content/docs/test_framework/Mocking.md
eugene-krivobokov: It's a tutorial and must cover popular use cases as simple as possible.

The explanation of implementation is verbose and needed only when I need to add a custom one.
Maybe it's better to extract it to another page.

docs/content/docs/test_framework/Mocking.md
```diff
@@ -5,11 +5,73 @@ type: docs
...
+
 ## Mocks (internal)
 
-### Mocking network
+### Networking
+
+#### Mocking a network on http layer
```
sergioboi, line 15: Does it work without additional logic in http client?
Mock webserver has url. Does url automatically used by applications?

eugene-krivobokov: 1. `AvitoInstrumentTestRunner : InHouseInstrumentationTestRunner()` 
  - Starts a MockWebServer
  - Pass the URL through a Bundle
2. `AvitoTestApp` reads this URL and uses to create Dagger modules

sergioboi: Why didn't we write that details? Without them we can confuse our users.

eugene-krivobokov: This scenario is [tutorial/how-to](https://avito-tech.github.io/avito-android/docs/contributing/docs/#use-appropriate-format-for-the-problem) for internal users. It should explain only how to use it.

I see two other scenarios:

- How to add a new custom rule? 
This explanation is already on this page. Agree, it may be better.
We use the same mechanics for all Rules. There is nothing specific in network.

- How to add ... to a new application?
I think it's better to add it when we explain how to use runner in another project. We may reuse previous scenario.

PR #242, opened 2020-03-08
MBS-7635 Fix IDE sync

eugene-krivobokov:


build.gradle.kts
```diff
@@ -33,8 +33,10 @@ val compileSdk = 29
 
 val publishToArtifactoryTask = tasks.register<Task>("publishToArtifactory") {
     group = "publication"
-    requireNotNull(artifactoryUrl) {
-        "Property artifactoryUrl is required for publishing"
+    doLast {
+        requireNotNull(artifactoryUrl) {
+            "Property artifactoryUrl is required for publishing"
+        }
     }
```
eugene-krivobokov, line 40: Again... It seems we have to use some custom logic to check gradle properties in a configuration a bit smarter.

Naive implementation in a configuration block breaks IDE sync.
(even when "Do not build Gradle task list during Gradle sync" is enabled)

```
    ...
	at org.gradle.plugins.ide.internal.tooling.GradleProjectBuilder.tasks(GradleProjectBuilder.java:104)
	at org.gradle.plugins.ide.internal.tooling.GradleProjectBuilder.buildHierarchy(GradleProjectBuilder.java:80)
	at org.gradle.plugins.ide.internal.tooling.GradleProjectBuilder.buildAll(GradleProjectBuilder.java:52)
	at org.gradle.plugins.ide.internal.tooling.IdeaModelBuilder.buildAll(IdeaModelBuilder.java:75)
	at org.gradle.plugins.ide.internal.tooling.IdeaModelBuilder.buildAll(IdeaModelBuilder.java:57)
```

But we want to check it before task execution.

sergioboi: Can you explain how does configuration block break sync?

eugene-krivobokov: As I understand the stacktrace from IDE, it creates greedily all tasks internally by tooling API.

dsvoronin (approved):

PR #245, opened 2020-03-10
MBS-1057: remove testType

dsvoronin:
deprecate separate types(used for filtering only now)
introduce `@Kind` to replace `@Type` and separate types

ignore `@Features`

Kind -> TestKind to avoid name clash, because both `@Kind` annotation and Kind enum will be used in test class


subprojects/android-test/test-annotations/src/main/kotlin/com/avito/android/test/annotations/TestType.kt
```diff
@@ -1,40 +1,46 @@
 package com.avito.android.test.annotations
 
+@Deprecated("2020.3.2; use @Kind")
```
sergioboi: Maybe we need to use `replaceWith` argument?

subprojects/android-test/test-annotations/build.gradle.kts
```diff
@@ -3,3 +3,7 @@ plugins {
     `maven-publish`
     id("com.jfrog.bintray")
 }
+
+dependencies {
+    implementation(project(":subprojects:common:report-viewer"))
```
sergioboi, line 8: It's not clear why do we need report-viewer in test-annotations module.
Maybe we have problems with module naming?

sergioboi (approved):

int02h (approved):

subprojects/android-test/test-annotations/build.gradle.kts
dsvoronin: yep, `report-viewer` -> `report` and annotations should be there

subprojects/android-test/test-annotations/src/main/kotlin/com/avito/android/test/annotations/TestType.kt
dsvoronin: i decided to revert `@Kind` annotation for now, and use old ones, so can't determine replaceWith value

PR #250, opened 2020-03-13
MBS-7908: ignore non-library dependencies in impact analysis

dsvoronin:


eugene-krivobokov (approved):

subprojects/gradle/impact-shared/src/main/java/com/avito/impact/configuration/ImplementationConfiguration.kt
```diff
@@ -58,6 +59,10 @@ class ImplementationConfiguration(module: InternalModule) : SimpleConfiguration(
                     .dependencies
                     .withType(DefaultProjectDependency::class.java)
             }
+            .filter {
+                val category = it.attributes.getAttribute(Category.CATEGORY_ATTRIBUTE)?.name
```
sergioboi, line 63: It is not clear line. Can you explain what is a `Category` and `CATEGORY_ATTRIBUTE`?

subprojects/gradle/impact-shared/src/main/java/com/avito/impact/configuration/ImplementationConfiguration.kt
```diff
@@ -58,6 +59,10 @@ class ImplementationConfiguration(module: InternalModule) : SimpleConfiguration(
...
+                category == null || category == Category.LIBRARY
```
sergioboi, line 64: What deps are with category null?

dsvoronin: most of them is null (i just used debugger to check, don't know much)

subprojects/gradle/impact-shared/src/main/java/com/avito/impact/configuration/ImplementationConfiguration.kt
dsvoronin: https://docs.gradle.org/current/javadoc/org/gradle/api/attributes/Category.html

PR #257, opened 2020-03-16
MBS-7688 Skip all espresso errors in retry

eugene-krivobokov:
- [x] fullCheck is ok (internal build 13811563)

### DIFF
--- a/subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/runner/ReportFriendlyFailureHandler.kt
+++ b/subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/runner/ReportFriendlyFailureHandler.kt
@@ -5,6 +5,7 @@ import android.view.View
 import androidx.test.espresso.AppNotIdleException
 import androidx.test.espresso.EspressoException
 import androidx.test.espresso.FailureHandler
+import androidx.test.espresso.NoMatchingRootException
 import androidx.test.espresso.PerformException
 import androidx.test.espresso.base.DefaultFailureHandler
 import junit.framework.AssertionFailedError
@@ -95,6 +96,12 @@ class ReportFriendlyFailureHandler(targetContext: Context) : FailureHandler {
                 }
                 e is AssertionFailedError -> {
                     throw AssertionFailedError(e.normalizedMessage())
+                }
+                e is NoMatchingRootException -> {
+                    throw UITestFrameworkException(
+                        message = e.normalizedMessage(),
+                        cause = e
+                    )
                 }
                 e is EspressoException -> {
                     throw e
--- a/subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/runner/ReportFriendlyFailureHandler.kt
+++ b/subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/runner/ReportFriendlyFailureHandler.kt
@@ -3,8 +3,8 @@ package com.avito.android.runner
 import android.content.Context
 import android.view.View
 import androidx.test.espresso.AppNotIdleException
+import androidx.test.espresso.EspressoException
 import androidx.test.espresso.FailureHandler
-import androidx.test.espresso.NoMatchingViewException
 import androidx.test.espresso.PerformException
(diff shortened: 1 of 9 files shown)


subprojects/android-test/ui-testing-core/src/main/kotlin/com/avito/android/test/UITestConfig.kt
```diff
@@ -45,8 +44,7 @@ object UITestConfig {
      * Exceptions to be waited for; any unregistered exceptions will be propagated
      */
     var waiterAllowedExceptions = setOf(
-        PerformException::class.java,
-        NoMatchingViewException::class.java,
+        EspressoException::class.java,
         AssertionError::class.java
     )
```
eugene-krivobokov, line 49: I've faced with another espresso exception - `AmbiguousViewMatcherException` in e2e tests.
I decided that it's unreliable to duplicate all possible espresso errors here. It's better to use a common marker type.

subprojects/android-test/ui-testing-core/src/main/kotlin/com/avito/android/test/Waiter.kt
```diff
@@ -11,7 +11,7 @@ import androidx.test.espresso.ViewInteraction
 fun <T> waitFor(
     frequencyMs: Long = UITestConfig.waiterFrequencyMs,
     timeoutMs: Long = UITestConfig.waiterTimeoutMs,
-    allowedExceptions: Set<Class<out Throwable>> = UITestConfig.waiterAllowedExceptions,
+    allowedExceptions: Set<Class<out Any>> = UITestConfig.waiterAllowedExceptions,
```
eugene-krivobokov, line 14: cause 

```
public interface EspressoException extends TestFrameworkException
public interface TestFrameworkException
```

subprojects/android-test/ui-testing-core/src/main/kotlin/com/avito/android/test/matcher/RecyclerViewMatcher.kt
```diff
@@ -14,7 +14,7 @@ class RecyclerViewMatcher {
             private var actualCount: Int? = null
 
             override fun performMatching(recyclerView: RecyclerView): Boolean {
-                actualCount = recyclerView.adapter!!.itemCount
+                actualCount = recyclerView.adapter?.itemCount ?: 0
```
eugene-krivobokov, line 17: Another example: missed NPE if a recycler view is not ready yet.

subprojects/android-test/ui-testing-core/src/main/kotlin/com/avito/android/test/Waiter.kt
sergioboi: Maybe we cant extend our exception from that interface. It will make the contract more clear

eugene-krivobokov: But we don't have the exception here, only marker interface (TestFrameworkException).

And I don't want to wrap all different errors in our single type. We had this approach partially before and it turned out as fragile. Even more, I don't see profits for the user to have our custom error from the test framework.

dsvoronin (approved):

sergioboi (approved):

subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/runner/ReportFriendlyFailureHandler.kt
```diff
@@ -96,7 +97,13 @@ class ReportFriendlyFailureHandler(targetContext: Context) : FailureHandler {
                 e is AssertionFailedError -> {
                     throw AssertionFailedError(e.normalizedMessage())
                 }
-                e is NoMatchingViewException -> {
+                e is NoMatchingRootException -> {
+                    throw UITestFrameworkException(
```
eugene-krivobokov, line 101: This is a mistake. We shouldn't wrap espresso error by UITestFrameworkException.
By default, retry skips Espresso errors but not UITestFrameworkException.

One possible solution is to check the cause of UITestFrameworkException.

PR #259, opened 2020-03-16
change env to gradle properties

sergioboi:
Save ~ 2min of build


_main.sh
```diff
@@ -87,15 +95,7 @@ function runInBuilder() {
     --workdir /app \
     --env GRADLE_USER_HOME=/gradle \
     --env LOCAL_USER_ID="$USER_ID" \
-    --env BINTRAY_USER="$BINTRAY_USER" \
-    --env BINTRAY_API_KEY="$BINTRAY_API_KEY" \
     --env BINTRAY_GPG_PASSPHRASE="$BINTRAY_GPG_PASSPHRASE" \
```
sergioboi, line 98: Is env used by bintray plugin? I decided to keep it because it doesn't influence our configuration

build.sh
```diff
@@ -7,7 +7,7 @@ TEMP_PROJECT_VERSION="temp-version"
 
 # `tasks` triggers full tasks graph resolving, checking for possible misconfigurations
 runInBuilder "./gradlew tasks build publishToMavenLocal ${GRADLE_ARGS} -PprojectVersion=${TEMP_PROJECT_VERSION};
-    ./gradlew :subprojects:android-test:test-app:instrumentationUi ${GRADLE_ARGS} -PinfraVersion=${TEMP_PROJECT_VERSION}"
+    ./gradlew :subprojects:android-test:test-app:instrumentationUi ${GRADLE_ARGS} -PinfraVersion=${TEMP_PROJECT_VERSION};
+    ./gradlew help"
```
sergioboi: run without GRADLE_ARGS it the same as run new docker without env

buildSrc/src/main/java/Dependencies.kt
```diff
@@ -55,6 +55,9 @@ object Dependencies {
     object gradle {
         val androidPlugin = "com.android.tools.build:gradle:3.5.3"
         val kotlinPlugin = "org.jetbrains.kotlin:kotlin-gradle-plugin:${Versions.kotlin}"
+        object avito {
+            val kotlinDslSupport = "com.avito.android:kotlin-dsl-support"
```
sergioboi, line 59: Minor change. Need to use in buildscript. I don't know how use infraVersion in buildSrc

build.sh
eugene-krivobokov: What about running it as the first command?

```
runInBuilder " 
1 mavenLocal
2 help
3 instrumentationUI
"
```

It shows problems in configuration. It's easier to catch them as early as possible, in the simplest run.

subprojects/gradle/slack/build.gradle.kts
```diff
@@ -21,3 +30,16 @@ dependencies {
...
+
+tasks.withType(Test::class.java).forEach { testTask ->
+    with(testTask) {
+        val testProperties = listOf(
+            "avito.slack.test.channel",
+            "avito.slack.test.token",
+            "avito.slack.test.workspace"
+        )
+        testProperties.forEach { key ->
+            systemProperty(key, project.getOptionalStringProperty(key) ?: "")
```
eugene-krivobokov, line 42: Good idea to have the same name of env and gradle property

eugene-krivobokov (approved):

PR #260, opened 2020-03-17
MBS-7944 Disable report by default in runner

eugene-krivobokov:


subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/runner/TestRunEnvironment.kt
```diff
@@ -113,8 +113,7 @@ fun provideEnvironment(
                 jobSlug = argumentsProvider.getMandatoryArgument("jobSlug"),
                 runId = argumentsProvider.getMandatoryArgument("runId")
             )
-            // todo move to instrumentation plugin
```
eugene-krivobokov, line 116: I didn't get it

dsvoronin (approved):

PR #261, opened 2020-03-17
MBS-7340 Add upload build to CI plugin

eugene-krivobokov:


subprojects/gradle/cd/src/main/kotlin/com/avito/ci/CdPlugin.kt
```diff
@@ -16,7 +16,7 @@ class CdPlugin : Plugin<Project> {
 
     override fun apply(project: Project) {
 
-        //todo создавать типы билдов динамически
+        //TODO: MBS-7948 create builds dynamically
```
eugene-krivobokov, line 19: I don't know how long it will take. I want to fix qapps builds firstly.

dsvoronin (approved):

PR #263, opened 2020-03-17
MBS-7865

sergioboi:
Move clear leaked deployments from python to kotlin
Roadmap:

- [x] Disable configuration on local builds
- [x] Add exclusive filter for kotlin-cli dependency
- [x] Delete leaked synthetic deployment
- [x] Delete local deployments
- [x] Add ability to delete deployment by name
- [x] Recognize synthetic and performance deployments
- [x] Delete connect-to-connect deployment
- [x] Add slack notification about leaked undeleted deployments
- [x] ~Reuse creation and delete k8s logic~
- [x] ~Add validation teamcityId when create deployment~
- [x] ~Create a task for buildscript modules~
- [x] After merging add teamcity trigger
- [x] ~Delete python code~
- [x] ~Delete deployments in bad status~


settings.gradle.kts
```diff
@@ -70,6 +72,8 @@ include(":subprojects:android-test:junit-utils")
 include(":subprojects:android-test:test-annotations")
 include(":subprojects:android-test:keep-for-testing")
 
+include(":subprojects:ci:clear-k8s-deployments")
```
eugene-krivobokov: Let's name modules as nouns, not verbs.

cleaner, collector, ...

subprojects/ci/clear-k8s-deployments/build.gradle.kts
```diff
@@ -0,0 +1,40 @@
...
+    api(Dependencies.kubernetesClient)
+    api(project(":subprojects:common:teamcity-common"))
+    api(Dependencies.kotlinXCli)
+}
+
+// todo add if ci
+//if(project.buildEnvironment is com.avito.utils.gradle.BuildEnvironment.CI) {
+if (project.getOptionalStringProperty("ci", "false").toBoolean()) {
+    tasks.register("run", JavaExec::class.java) {
+        main = "com.avito.ci.ClearK8SDeploymentsMain"
```
eugene-krivobokov: Why don't we wrap this logic into gradle plugin?

In CLI apps we have to solve again common issues.

sergioboi: How do you see the plugin solves these problems?

eugene-krivobokov: What I name as issues:
- Logging, monitoring
- Integration tests
- ... similar stuff

I know that it is easy in simple CLI libraries. But we use gradle plugins to deal with them in a common manner.
I'm not against CLI libraries. I just want to be sure that we can manage these issues easily and have appropriate utils for it.

subprojects/ci/clear-k8s-deployments/src/main/kotlin/com.avito.ci/ClearK8SDeploymentsByNamespaces.kt
```diff
@@ -0,0 +1,72 @@
...
+                        BuildState.QUEUED,
+                        BuildState.RUNNING -> false
+                    }
+                }
+                is DeploymentEnvironment.Local -> {
+                    val creation = Date(env.creationTimeInMillis)
+                    val now = Date(System.currentTimeMillis())
+                    val hourAgo = Date(
+                        now.toInstant()
+                            .minusMillis(TimeUnit.HOURS.toMillis(1))
```
eugene-krivobokov: Let's extract this value. At least to a named constant.

subprojects/ci/clear-k8s-deployments/build.gradle.kts
eugene-krivobokov: Example: 
in ClearDeployments we could reuse TeamcityCredentials, KubernetesCredentials and do no pass these parameters in another way.

eugene-krivobokov (approved):

subprojects/ci/clear-k8s-deployments/build.gradle.kts
sergioboi: I understand but a plugin is something for projects.
This is no plugin case. But I agree we need possibilities to reuse our logic

PR #264, opened 2020-03-18
Mbs 7932 synth monitoring

dsvoronin:


subprojects/android-test/test-annotations/src/main/kotlin/com/avito/android/test/annotations/TestType.kt
```diff
@@ -1,6 +1,5 @@
 package com.avito.android.test.annotations
 
-//todo move to report module
```
dsvoronin, line 3: depends on our decision about these annotations; so no direct actions yet

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/suite/filter/PackagePrefixFilter.kt
```diff
@@ -3,14 +3,9 @@ package com.avito.instrumentation.suite.filter
 import com.avito.instrumentation.suite.dex.TestInApk
 import com.avito.report.model.DeviceName
 
-//todo убрать nullable
```
dsvoronin, line 6: boy scouting

eugene-krivobokov (approved):

PR #267, opened 2020-03-18
Mbs 7955 performance fix

dsvoronin:


subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/rerun/BuildOnTargetCommitForTestTask.kt
```diff
@@ -16,7 +16,6 @@ import org.gradle.workers.WorkerExecutor
 import javax.inject.Inject
 
 @Suppress("UnstableApiUsage")
-@CacheableTask
```
sergioboi, line 19: :peka-face:

sergioboi (approved):

fat-fellow (approved):

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/rerun/BuildOnTargetCommitForTestTask.kt
dsvoronin: > because we don't fail this task, and even failed build without apks will be cached

motivation from commit message

PR #268, opened 2020-03-18
Remove deprecation from doesNotExist check

lukaville:
There are cases when it's not possible to check negative scenarios without `doesNotExist` check, therefore it would be better to remove the deprecation as there are no alternatives.


eugene-krivobokov:
> There are cases when it's not possible to check negative scenarios without `doesNotExist` check, therefore it would be better to remove the deprecation as there are no alternatives.

Agree, sometimes we have no choice. But we don't want to encourage using this check.
This is why it has been marked as deprecated.

Do we have another option to show this warning? Maybe custom lint check? We've discussed lint checks for many reasons.

sergioboi:
Deprecation with `Level.Warning` will be enough. And we can clarify our message in Deprecated

eugene-krivobokov:
> Deprecation with `Level.Warning` will be enough. And we can clarify our message in Deprecated

LGTM

lukaville:
OK, warning deprecation level is set by default, so I just translated the message

eugene-krivobokov (approved):

dsvoronin (approved):

PR #273, opened 2020-03-23
Mbs-7650 add instrumentationDynamic support for local runs

sergioboi:
- [x] Add parameter to dynamic invocation in Avito http://links.k.avito.ru/lr
- [x] Create a task for moving instrumentation-defaults-plugin to Avito
- [x] ~Added possibility to invoke `dynamic` without device targets?~

### DIFF
--- a/subprojects/android-test/test-app/build.gradle.kts
+++ b/subprojects/android-test/test-app/build.gradle.kts
@@ -46,7 +46,7 @@ android {
         }
     }
 
-    testBuildType = getOptionalStringProperty("avito.testBuildType", "staging")
+    testBuildType = getOptionalStringProperty("testBuildType", "staging")
 
     /**
      * Disable all buildTypes except testing
--- a/subprojects/android-test/test-app/build.gradle.kts
+++ b/subprojects/android-test/test-app/build.gradle.kts
@@ -46,7 +46,7 @@ android {
         }
     }
 
-    testBuildType = "staging"
+    testBuildType = getOptionalStringProperty("avito.testBuildType", "staging")
 
     /**
      * Disable all buildTypes except testing
--- a/subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/InstrumentationTestsPlugin.kt
+++ b/subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/InstrumentationTestsPlugin.kt
@@ -15,6 +15,9 @@ import com.avito.git.gitState
 import com.avito.git.isOnDefaultBranch
 import com.avito.instrumentation.configuration.ImpactAnalysisPolicy
 import com.avito.instrumentation.configuration.InstrumentationPluginConfiguration
+import com.avito.instrumentation.configuration.target.scheduling.SchedulingConfiguration
+import com.avito.instrumentation.configuration.target.scheduling.quota.QuotaConfiguration
+import com.avito.instrumentation.configuration.target.scheduling.reservation.TestsBasedDevicesReservationConfiguration
 import com.avito.instrumentation.executing.ExecutionParameters
 import com.avito.instrumentation.rerun.BuildOnTargetCommitForTestTask
 import com.avito.instrumentation.rerun.RunOnTargetBranchCondition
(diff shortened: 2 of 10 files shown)


subprojects/android-test/test-app/build.gradle.kts
```diff
@@ -46,7 +46,7 @@ android {
         }
     }
 
-    testBuildType = "staging"
+    testBuildType = getOptionalStringProperty("avito.testBuildType", "staging")
```
eugene-krivobokov: Why "avito"? This module and logic knows nothing about avito
What about simple "testBuildType"?

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/TestsFilter.kt
```diff
@@ -0,0 +1,44 @@
+package com.avito.instrumentation
+
+private val manualAnnotations = setOf(
+    "com.avito.android.test.annotations.ManualTest",
```
dsvoronin, line 4: that's a dependency leak
these annotations are avito TMS specific and should not be specified in generic instrumentation config

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/configuration/target/scheduling/quota/QuotaConfiguration.kt
```diff
@@ -17,7 +17,7 @@ open class QuotaConfiguration {
         require(minimumSuccessCount >= 0) { "minimumSuccessCount must be positive or 0" }
         require(minimumFailedCount >= 0) { "minimumFailedCount must be positive or 0" }
         require(minimumRequiredRunsCount > 0) {
-            "minimumRequiredRunsCount (minimumSuccessCount + minimumFailedCount) must be positive"
+            "minimumRequiredRunsCount (minimumSuccessCount + minimumFailedCount) must be greater then zero"
```
dsvoronin: than

dsvoronin (changes requested):

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/TestsFilter.kt
dsvoronin: you can probably ignore all the filtering besides prefixTestFilter in dynamic config

sergioboi: Do you mean I can make `annotatedWith` equals to null?

dsvoronin: yep, i think you can, dynamic means we could possibly run everything with optional filters, specified by client

sergioboi: It could break expectations from our users who ran from teamcity.
By default, we executed the only ui annotations. Is it critical?
Btw how do we filter performance tests?

dsvoronin: well, on the other hand you will introduce a step back in reusability of runner, is it really worth it?

sergioboi: I will create task to move instrumentation-defaults to Avito
Maybe better refactor current behavior during future task?

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/InstrumentationTestsPlugin.kt
```diff
@@ -259,7 +351,8 @@ class InstrumentationTestsPlugin : Plugin<Project> {
         args["sentryDsn"] = "http://stub-project@stub-host/0"
         args["fileStorageUrl"] = "http://stub"
         args["teamcityBuildId"] = project.envArgs.build.id.toString()
-        args["avito.report.enabled"] = project.getBooleanProperty("avito.report.enabled", default = false).toString()
+        args["avito.report.enabled"] =
+            project.getBooleanProperty("avito.report.enabled", default = false).toString()
```
eugene-krivobokov, line 355: Is it autoformatting or manual?

sergioboi: autoformatting

eugene-krivobokov (approved):

dsvoronin (approved):

PR #275, opened 2020-03-24
Mbs 7661 show touches in video

int02h:
Unfortunately, the common approach with `adb shell settings put system show_touches 1` is not working neither for Espresso-click nor for InProcess-clicks. The emulator only shows "true" clicks made by a real user.

So I've implemented a simple solution to show clicks and long clicks using foreground drawable of a `View`. This approach work only on API level 23 and above.


subprojects/android-test/ui-testing-core/src/main/kotlin/com/avito/android/test/espresso/action/click/ClickAction.kt
```diff
@@ -61,16 +62,19 @@ class ClickAction(
                 coordinates: FloatArray,
                 precision: FloatArray
             ) {
+                val clickVisualization = ClickVisualization(coordinates[0], coordinates[1])
```
eugene-krivobokov: This feature has to be optional for non-functional tests (performance, screenshots)

eugene-krivobokov (changes requested):

subprojects/android-test/ui-testing-core/src/main/kotlin/com/avito/android/test/espresso/action/click/ClickAction.kt
int02h: Fixed in 85d51aa32dabd529c5c6c9ec2bde91d27ac58c22

subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/runner/InHouseInstrumentationTestRunner.kt
```diff
@@ -225,6 +226,25 @@ abstract class InHouseInstrumentationTestRunner :
         return super.onException(obj, e)
     }
 
+    @CallSuper
+    open fun initUITestConfig() {
```
int02h, line 230: It allows `ScreenshotTestRunner` in main Avito project to disable clicks visualization

subprojects/android-test/ui-testing-core/src/main/kotlin/com/avito/android/test/UITestConfig.kt
```diff
@@ -40,6 +40,13 @@ object UITestConfig {
 
     var clicksType: ClickType = defaultClicksType
 
+    const val defaultVisualizeClicks = true
```
eugene-krivobokov: Why do we expose this value?

eugene-krivobokov (approved):

subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/runner/InHouseInstrumentationTestRunner.kt
int02h: Have a look at `subprojects/android-test/test-report/src/main/kotlin/com/avito/android/test/report/PerformanceTestRule.kt`. This rule is able to restore default value easily and without saving original value on its side.

subprojects/android-test/ui-testing-core/src/main/kotlin/com/avito/android/test/UITestConfig.kt
int02h: Have a look at `subprojects/android-test/test-report/src/main/kotlin/com/avito/android/test/report/PerformanceTestRule.kt`. This rule is able to restore default value easily and without saving original value on its side.

eugene-krivobokov: It can clear a custom value. If someone is already disabled it, we will enable it.

I think we should save and restore the current value, not the default one

subprojects/android-test/test-report/src/main/kotlin/com/avito/android/test/report/PerformanceTestRule.kt
```diff
@@ -17,10 +18,16 @@ class PerformanceTestRule : TestRule {
...
 
-    override fun apply(base: Statement, description: Description): Statement = object : Statement() {
-        override fun evaluate() {
-            UiDevice.getInstance(InstrumentationRegistry.getInstrumentation())
-            base.evaluate()
+    override fun apply(base: Statement, description: Description): Statement =
+        object : Statement() {
+            override fun evaluate() {
+                UITestConfig.visualizeClicks = false
+                UiDevice.getInstance(InstrumentationRegistry.getInstrumentation())
```
sboishtyan: Why do we need UIDevice?

subprojects/android-test/ui-testing-core/src/main/kotlin/com/avito/android/test/espresso/action/click/ClickVisualization.kt
```diff
@@ -0,0 +1,71 @@
...
+import android.graphics.Color
+import android.graphics.ColorFilter
+import android.graphics.Paint
+import android.graphics.PixelFormat
+import android.graphics.drawable.Drawable
+import android.os.Build
+import android.view.View
+import androidx.annotation.ColorInt
+
+class ClickVisualization(
```
sergioboi: Internal

sergioboi (changes requested):

subprojects/android-test/ui-testing-core/src/main/kotlin/com/avito/android/test/UITestConfig.kt
int02h: You're right. I will fix it. Thank you for spotting the problem.

subprojects/android-test/ui-testing-core/src/main/kotlin/com/avito/android/test/espresso/action/click/ClickVisualization.kt
int02h: Yep. I'll fix it.

subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/runner/InHouseInstrumentationTestRunner.kt
dsvoronin: Is is possible with `UITestConfig.visualizeClicks = false` only, right?
if so, i think its too much for this PR to change inhouse runner interface

sergioboi (approved):

subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/runner/InHouseInstrumentationTestRunner.kt
int02h: I think that derived classes should be able to setup UITestConfig.

int02h: This should be clear and obvious for derived classed where they can setup UITestConfig. If someone sets its values in config before super class initializes it then custom values of the derived class will be overwritten.

dsvoronin: it's ok for me, just have an issue with the weird scope of the PR


We should reconsider this UITestConfig singleton at all, Kaspresso made it right with its config imo

PR #277, opened 2020-03-25
MBS-7848 emulator debug in k8s

int02h:
Add the possibility to debug test in an emulator in k8s.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,6 @@
 test_build_type?=debug
 log_level?=-q
 filter_report_id?=
-ca_cert_file=~/.kube/avito_ca.crt
 
 help:
 
@@ -15,7 +14,7 @@ test_app_instrumentation:
 	./gradlew subprojects\:android-test\:test-app\:instrumentationUi -PinfraVersion=local -Pci=true -PtestBuildType=$(test_build_type) $(log_level)
 
 test_app_instrumentation_android_debug:
-	./gradlew :subprojects:android-test:test-app:instrumentationUiDebug -PinfraVersion=local -Pci=true -PkubernetesContext=beta -PkubernetesCaCertFile=$(ca_cert_file) -PtestBuildType=$(test_build_type) $(log_level)
+	./gradlew :subprojects:android-test:test-app:instrumentationUiDebug -PinfraVersion=local -Pci=true -PkubernetesContext=beta -PtestBuildType=$(test_build_type) $(log_level)
 
 dynamic_properties:
 	$(eval keepFailedTestsFromReport?=)
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,7 @@
 test_build_type?=debug
 log_level?=-q
 filter_report_id?=
+ca_cert_file=~/.kube/avito_ca.crt
 
 help:
 
@@ -12,6 +13,9 @@ test_app_instrumentation_debug:
 
 test_app_instrumentation:
 	./gradlew subprojects\:android-test\:test-app\:instrumentationUi -PinfraVersion=local -Pci=true -PtestBuildType=$(test_build_type) $(log_level)
+
+test_app_debug:
+	./gradlew :subprojects:android-test:test-app:instrumentationUiDebug -PinfraVersion=local -Pci=true -PkubernetesContext=beta -PkubernetesCaCertFile=$(ca_cert_file) -PtestBuildType=$(test_build_type) $(log_level)
 
 dynamic_properties:
 	$(eval keepFailedTestsFromReport?=)
--- a/Makefile
+++ b/Makefile
@@ -8,13 +8,13 @@ help:
 publish_to_maven_local:
 	./gradlew publishToMavenLocal -PprojectVersion=local $(log_level)
 
-test_app_instrumentation_debug:
+test_app_instrumentation_gradle_debug:
 	./gradlew subprojects\:android-test\:test-app\:instrumentationUi -Dorg.gradle.jvmargs='-Xdebug -Xrunjdwp:transport=dt_socket,server=y,address=5005,suspend=y' --no-daemon -PinfraVersion=local -Pci=true -PtestBuildType=$(test_build_type) $(log_level)
 
 test_app_instrumentation:
 	./gradlew subprojects\:android-test\:test-app\:instrumentationUi -PinfraVersion=local -Pci=true -PtestBuildType=$(test_build_type) $(log_level)
 
-test_app_debug:
+test_app_instrumentation_android_debug:
 	./gradlew :subprojects:android-test:test-app:instrumentationUiDebug -PinfraVersion=local -Pci=true -PkubernetesContext=beta -PkubernetesCaCertFile=$(ca_cert_file) -PtestBuildType=$(test_build_type) $(log_level)
 
 dynamic_properties:
--- a/subprojects/android-test/test-app/build.gradle.kts
+++ b/subprojects/android-test/test-app/build.gradle.kts
@@ -139,26 +139,27 @@ extensions.getByType<GradleInstrumentationPluginConfiguration>().apply {
     configurationsContainer.register("ui") {
         tryToReRunOnTargetBranch = false
         reportSkippedTests = true
-        rerunFailedTests = true
+        rerunFailedTests = false
         reportFlakyTests = true
+        prefixFilter = "com.avito.android.ui.test.AppBarTest"
 
-        targetsContainer.register("api22") {
-            deviceName = "API22"
-
-            scheduling = SchedulingConfiguration().apply {
-                quota = QuotaConfiguration().apply {
-                    retryCount = 1
-                    minimumSuccessCount = 1
-                }
-
-                reservation = TestsBasedDevicesReservationConfiguration().apply {
-                    device = Emulator22
-                    maximum = 50
-                    minimum = 2
-                    testsPerEmulator = 3
-                }
-            }
-        }
+//        targetsContainer.register("api22") {
+//            deviceName = "API22"
+//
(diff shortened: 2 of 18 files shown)


Makefile
```diff
@@ -1,4 +1,5 @@
 test_build_type?=debug
+ca_cert_file?=avito_ca.crt
```
eugene-krivobokov: Do you propose to store it in the project?
What about `~/.kube`?

subprojects/android-test/test-app/build.gradle.kts
```diff
@@ -136,7 +136,7 @@ extensions.getByType<GradleInstrumentationPluginConfiguration>().apply {
     configurationsContainer.register("ui") {
         tryToReRunOnTargetBranch = false
         reportSkippedTests = true
-        rerunFailedTests = true
+        rerunFailedTests = false
```
eugene-krivobokov: Why is that? How can I check my fixes without rerun?

subprojects/android-test/test-app/build.gradle.kts
```diff
@@ -168,9 +168,37 @@ extensions.getByType<GradleInstrumentationPluginConfiguration>().apply {
...
+            }
+        }
+    }
+
+    configurationsContainer.register("uiDebug") {
+        tryToReRunOnTargetBranch = false
+        reportSkippedTests = false
+        rerunFailedTests = false
+        reportFlakyTests = false
+        prefixFilter = "com.avito.android.ui.test.AppBarTest"
```
eugene-krivobokov: Can we not use filtering?
It's easy to forget about this filter

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/InstrumentationTestsTask.kt
```diff
@@ -33,6 +33,12 @@ abstract class InstrumentationTestsTask @Inject constructor(
     private val workerExecutor: WorkerExecutor
 ) : DefaultTask() {
 
+    init {
+        outputs.upToDateWhen {
+            false
+        }
+    }
```
eugene-krivobokov: ?

eugene-krivobokov (changes requested):

subprojects/android-test/test-app/build.gradle.kts
int02h: Oops. Accidentally changed. I'll turn it back. Thank you for your attention.

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/InstrumentationTestsTask.kt
int02h: Accidentally left. I'll remove it. It was necessary on the early steps of this task.

subprojects/android-test/test-app/build.gradle.kts
int02h: Yes. Of course. I'll remove it. But running all tests is not a good idea. We need to design the mechanism if running specified tests. And this mechanism should be convenient for our users.

Makefile
int02h: Fixed

subprojects/android-test/test-app/build.gradle.kts
int02h: Fixed

subprojects/android-test/test-app/build.gradle.kts
int02h: Fixed

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/InstrumentationTestsTask.kt
int02h: Fixed

subprojects/android-test/test-app/build.gradle.kts
```diff
@@ -178,6 +178,33 @@ extensions.getByType<GradleInstrumentationPluginConfiguration>().apply {
...
+            deviceName = "API27"
+
+            scheduling = SchedulingConfiguration().apply {
+                quota = QuotaConfiguration().apply {
+                    retryCount = 1
+                    minimumSuccessCount = 1
+                }
+
+                reservation = TestsBasedDevicesReservationConfiguration().apply {
+                    device = Emulator27
```
eugene-krivobokov, line 200: We have the only one emulator for easier debugging.
It'd be great if we could pass a version through makefile.
Change build.gradle just to switch a version is inconvenient.

eugene-krivobokov (approved):

subprojects/android-test/test-app/build.gradle.kts
int02h: I totally agree. We need to develop a convenient way to debug tests:
1. Run a specific test or the whole class
1. Choose API level

But I think the proposal is out of the scope of this task.

eugene-krivobokov: Ok

Makefile
```diff
@@ -13,6 +14,9 @@ test_app_instrumentation_debug:
 test_app_instrumentation:
 	./gradlew subprojects\:android-test\:test-app\:instrumentationUi -PinfraVersion=local -Pci=true -PtestBuildType=$(test_build_type) $(log_level)
 
+test_app_debug:
```
sergioboi: Bad name.

subprojects/android-test/test-app/build.gradle.kts
```diff
@@ -178,6 +178,33 @@ extensions.getByType<GradleInstrumentationPluginConfiguration>().apply {
...
         }
     }
+
+    configurationsContainer.register("uiDebug") {
+        tryToReRunOnTargetBranch = false
+        reportSkippedTests = false
+        rerunFailedTests = false
+        reportFlakyTests = false
+        // uncomment after 2020.3.6 release (MBS-8050)
+        // debugMode = true
```
sergioboi: better is add debugMode in dynamicParams.
One more thing. debugMode is bad name.

subprojects/android-test/test-app/build.gradle.kts
```diff
@@ -178,6 +178,33 @@ extensions.getByType<GradleInstrumentationPluginConfiguration>().apply {
...
+        targetsContainer.register("api27") {
+            deviceName = "API27"
+
+            scheduling = SchedulingConfiguration().apply {
+                quota = QuotaConfiguration().apply {
+                    retryCount = 1
+                    minimumSuccessCount = 1
+                }
+
+                reservation = TestsBasedDevicesReservationConfiguration().apply {
```
sergioboi, line 199: All blockers for usage must be explicit. Not it's not a blocker to merge but blocker of DoD.
Create a task for describing debug restriction or make them clear for clients.

sergioboi (changes requested):

subprojects/android-test/test-app/build.gradle.kts
int02h: Can you please share your ideas about perfect naming for this parameter? Right now I see only criticism but without any suggestions.

Makefile
int02h: `test_app_instrumentation_debug` is already reserved. How about `test_app_instrumentation_tests_debug`?

subprojects/android-test/test-app/build.gradle.kts
int02h: And why debugMode should be in dynamicParams? `uiDebug` configuration was create exactly for test debugging. So `debugMode` should be always `true`.

subprojects/android-test/test-app/build.gradle.kts
int02h: About what blockers are you talking about?

Makefile
sergioboi: let rename `test_app_instrumentation_debug` to `test_app_instrumentation_gradle_debug`
New we can name `test_app_instrumentation_android_debug`

subprojects/android-test/test-app/build.gradle.kts
sergioboi: I don't like that `debugMode` said me nothing about what I will debug.
Could we rename it to `enableDeviceDebug`

subprojects/android-test/test-app/build.gradle.kts
sergioboi: Discussed in private chat.

subprojects/android-test/test-app/build.gradle.kts
sergioboi: About dynamic.
I don't like that We have a specific flag that works only with concrete target specification. I think we need abstract instrumentationConfiguration. It will help us to define a specific type-safe configuration. And We could extract your build script code to plugin and client will set only a few of parameters

Makefile
```diff
@@ -1,6 +1,7 @@
 test_build_type?=debug
 log_level?=-q
 filter_report_id?=
+ca_cert_file=~/.kube/avito_ca.crt
```
dsvoronin: ??

subprojects/android-test/test-app/build.gradle.kts
int02h: I've renamed `debugMode` to `enableDeviceDebug`.

About the convenient way of running tests in debug mode. We discussed it internally that this configuration is intended for our internal usage. For our customers we will design a proper and handy way to run test with a debugger.

sergioboi (approved):

PR #278, opened 2020-03-26
Mbs 7652 Fix in reserving new local devices

eugene-krivobokov:


subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/reservation/client/local/LocalReservationClient.kt
```diff
@@ -128,7 +126,7 @@ class LocalReservationClient(
         GlobalScope.launch {
             var emulators = localEmulators(reservation)
 
-            while (!result.isClosedForSend && emulators.isNotEmpty()) {
```
eugene-krivobokov, line 131: Contrary to k8s, we don't know what to expect from `adb devices`

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/reservation/client/local/LocalReservationClient.kt
```diff
@@ -46,7 +46,7 @@ class LocalReservationClient(
 
         val devicesChannel: Channel<WorkerDevice> = reservations
             .iterateInParallel { _, reservation ->
-                val deploymentName = generateDeploymentName()
+                val deploymentName = "local-stub"
```
sergioboi: Why do we name it deploymentName?

subprojects/gradle/runner/service/src/main/kotlin/com/avito/runner/service/worker/device/Serial.kt
```diff
@@ -0,0 +1,32 @@
+package com.avito.runner.service.worker.device
+
+import com.google.common.net.InetAddresses
+
+sealed class Serial(open val value: String) {
```
sergioboi: Can we use abstract?

subprojects/gradle/runner/service/src/main/kotlin/com/avito/runner/service/worker/device/Serial.kt
```diff
@@ -0,0 +1,32 @@
...
+
+    companion object {
+        fun from(value: String): Serial {
+            return if (isRemote(value)) Remote(
+                value
+            ) else Local(value)
+        }
+
+        @Suppress("UnstableApiUsage")
+        private fun isRemote(serial: String): Boolean {
```
sergioboi, line 29: Propose to add serial validation. We need to fail if serial will be unexpected

sergioboi (changes requested):

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/reservation/client/local/LocalReservationClient.kt
eugene-krivobokov: It's a fake. I'll delete it after refactoring the `ReservationClient` interface. 
It doesn't fit local runs

subprojects/gradle/runner/service/src/main/kotlin/com/avito/runner/service/worker/device/Serial.kt
eugene-krivobokov: Do you know these cases? 
I don't see what value might be broken or illegal.

subprojects/gradle/runner/service/src/main/kotlin/com/avito/runner/service/worker/device/Serial.kt
eugene-krivobokov: fixed

subprojects/gradle/runner/service/src/main/kotlin/com/avito/runner/service/worker/device/Serial.kt
sergioboi: It's not clear for me what serials could pass there.
I want to have a comment or test where we describe possible serials explicitly.

eugene-krivobokov: This is a serial number from adb.

https://developer.android.com/studio/command-line/adb#devicestatus

```
adb devices
List of devices attached
emulator-5556 device product:sdk_google_phone_x86_64 model:Android_SDK_built_for_x86_64 device:generic_x86_64
emulator-5554 device product:sdk_google_phone_x86 model:Android_SDK_built_for_x86 device:generic_x86
0a388e93      device usb:1-1 product:razor model:Nexus_7 device:flo
10.0.0.5:5555 device
```
We have no control and contract for this value. It's just a string or ip address.

Test: https://github.com/avito-tech/avito-android/pull/278/files#diff-a2df6709784cffb936391793c6296297

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/reservation/client/local/LocalReservationClient.kt
```diff
@@ -46,14 +46,14 @@ class LocalReservationClient(
 
         val devicesChannel: Channel<WorkerDevice> = reservations
             .iterateInParallel { _, reservation ->
-                val deploymentName = generateDeploymentName()
-                reservationDeployments.send(deploymentName)
+                val fakeDeploymentName = "local-stub"
```
sergioboi, line 49: "Deployment" part triggers me. What do you think about renaming `reservationDeployments` to something else?
I can't propose smthg but we can add todo or whatever.

sergioboi (approved):

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/reservation/client/local/LocalReservationClient.kt
eugene-krivobokov: Renaming is not enough.
I'll try to hide the deployment state and don't expose it to clients at all.
I want to try this approach: https://youtu.be/a3agLJQ6vt8?t=1750
In this case: Downloader == ReservationClient, References == reservations

PR #280, opened 2020-03-27
MBS-8048 fix test running on a clean emulator

int02h:
**Problem**
Tests do not run from Android Studio on a clean emulator. Orchestrator isn't being installed. AS shows error `Test running failed: No test results`.

**Steps to reproduce**
1. Remove orchestrator from the emulator: `adb uninstall androidx.test.orchestrator`;
1. Run any test from `test-app` using Android Studio;
1. Error `Test running failed: No test results` occurred.

### DIFF
--- a/subprojects/android-test/test-app/build.gradle.kts
+++ b/subprojects/android-test/test-app/build.gradle.kts
@@ -90,6 +90,9 @@ dependencies {
     androidTestImplementation(project(":subprojects:common:okhttp"))
     androidTestImplementation(project(":subprojects:common:time"))
 
+    androidTestImplementation(Dependencies.androidTest.runner)
+    androidTestUtil(Dependencies.androidTest.orchestrator)
+
     androidTestImplementation(Dependencies.test.junit)
     androidTestImplementation(Dependencies.okhttp)
     androidTestImplementation(Dependencies.okhttpLogging)
--- a/buildSrc/src/main/kotlin/Dependencies.kt
+++ b/buildSrc/src/main/kotlin/Dependencies.kt
@@ -68,6 +68,7 @@ object Dependencies {
 
     object androidTest {
         val runner = "androidx.test:runner:${Versions.androidXTest}"
+        val orchestrator = "androidx.test:orchestrator:${Versions.androidXTest}"
         val ddmlib = "com.android.tools.ddms:ddmlib:26.2.0"
         val uiAutomator = "androidx.test.uiautomator:uiautomator:2.2.0"
         val core = "androidx.test:core:${Versions.androidXTest}"


eugene-krivobokov (approved):

subprojects/android-test/test-app/build.gradle.kts
```diff
@@ -90,6 +90,9 @@ dependencies {
     androidTestImplementation(project(":subprojects:common:okhttp"))
     androidTestImplementation(project(":subprojects:common:time"))
 
+    androidTestImplementation(Dependencies.androidTest.runner)
+    androidTestUtil(Dependencies.androidTest.orchestrator)
```
sergioboi, line 94: What is an androidTestUtil?

subprojects/android-test/test-app/build.gradle.kts
```diff
@@ -90,6 +90,9 @@ dependencies {
     androidTestImplementation(project(":subprojects:common:okhttp"))
     androidTestImplementation(project(":subprojects:common:time"))
 
+    androidTestImplementation(Dependencies.androidTest.runner)
```
sergioboi, line 93: How did it happen?

int02h: Happened what? Tests work only because the orchestrator was already installed on the emulator. For instance, when running Avito test.

subprojects/android-test/test-app/build.gradle.kts
int02h: It's an [official way](https://developer.android.com/training/testing/junit-runner#ato-gradle) to use runner.

subprojects/android-test/test-app/build.gradle.kts
sergioboi: Understand. Do we need Implementation dependency or can replace it with Api?

eugene-krivobokov: `implementation` dependency exposes less, thus it's preferable.
Also, it is in the application. No one can consume this module as dependency

int02h: And according to [doc](https://developer.android.com/training/testing/junit-runner#ato-gradle) we should use `implementation` not `api`.

sergioboi: Oh, wait. Implementation exposes more.

sergioboi: https://docs.gradle.org/current/userguide/java_library_plugin.html#sec:java_library_separation
```The plugin exposes two configurations that can be used to declare dependencies: api and implementation. The api configuration should be used to declare dependencies which are exported by the library API, whereas the implementation configuration should be used to declare dependencies which are internal to the component```
I think we need to use it like api

int02h: Why? It's a test application. Why should it export its dependencies like a library? Orchestrator is an internal dependency, so according to a piece of documentation you provided we should use `implementation`.

sergioboi: It is controversial. I was led by the principle that fewer dependencies are preferable. If we need something let add it explicitly

sergioboi: So, we discussed it in internal chat and realize that `implementation` and `api` influence only to consumers of a library.

PR #282, opened 2020-03-27
MBS-8042 remove unused output files

sergioboi:


subprojects/gradle/runner/shared/src/main/kotlin/com/avito/runner/CommandLineExecutor.kt
```diff
@@ -28,13 +28,14 @@ interface CommandLineExecutor {
             output: File?
         ): Observable<ProcessNotification> = Observable.create(
             { emitter ->
-                val outputFile: File = when {
-                    output == null || output.isDirectory -> {
-                        prepareOutputFile(output, true)
+                val outputFile: File? = when {
+                    output == null -> null
```
sergioboi, line 32: I supposed if the output file is null that we don't need to create ours because it won't be used

subprojects/gradle/runner/shared/src/main/kotlin/com/avito/runner/CommandLineExecutor.kt
```diff
@@ -80,15 +81,14 @@ interface CommandLineExecutor {
             }, Emitter.BackpressureMode.ERROR
         )
 
-        private fun prepareOutputFile(parent: File?, keepOnExit: Boolean): File = Random()
```
sergioboi, line 83: keepOnExit always was a true

Makefile
```diff
@@ -13,9 +13,10 @@ test_app_instrumentation:
 
 dynamic_properties:
 	$(eval testFilter?=empty)
+	$(eval skipTestsWithPrefix?=)
```
eugene-krivobokov: I propose to use the same names.
We have multiple options in filters and new entities make harder to understand the logic. I have to map values mentally.

eugene-krivobokov (approved):

Makefile
sergioboi: Agree. It's because of i cant rename everywhere

int02h (approved):

PR #284, opened 2020-03-30
Update error message

sergioboi:


subprojects/ci/k8s-deployments-cleaner/src/main/kotlin/com.avito.ci/ClearK8SDeploymentsByNamespaces.kt
```diff
@@ -37,7 +37,7 @@ class ClearK8SDeploymentsByNamespaces(
                         deployments.checkDeploymentLeak(deployment)
                     }
             } catch (e: Throwable) {
-                throw RuntimeException("Error when checked leak in namespace=$namespace", e)
+                throw RuntimeException("Error when checked leak in namespace=$namespace. If reason wasn't clear read previous discussion http://links.k.avito.ru/Az", e)
```
eugene-krivobokov, line 40: Can we extract any results from this conversation and add them to the error?

eugene-krivobokov (approved):

subprojects/ci/k8s-deployments-cleaner/src/main/kotlin/com.avito.ci/ClearK8SDeploymentsByNamespaces.kt
sergioboi: Everything has been extracted already. Look inside of `checkDeploymentLeak`

PR #285, opened 2020-03-30
MBS-7652 Consider reservation quota for local emulators

eugene-krivobokov:
In this fix, I intentionally have not changed behavior and API.

### DIFF
--- a/subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/reservation/client/kubernetes/KubernetesReservationClient.kt
+++ b/subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/reservation/client/kubernetes/KubernetesReservationClient.kt
@@ -9,7 +9,6 @@ import com.avito.instrumentation.reservation.request.Device
 import com.avito.instrumentation.reservation.request.Reservation
 import com.avito.instrumentation.util.forEachAsync
 import com.avito.instrumentation.util.iterateInParallel
-import com.avito.instrumentation.util.merge
 import com.avito.instrumentation.util.waitForCondition
 import com.avito.utils.logging.CILogger
 import com.fkorotkov.kubernetes.apps.metadata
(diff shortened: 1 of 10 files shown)


subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/reservation/client/kubernetes/KubernetesReservationClient.kt
```diff
@@ -65,46 +64,44 @@ class KubernetesReservationClient(
...
+                    deploymentName = deploymentName,
+                    count = reservation.count
+                )
+                is Device.Phone -> getDeviceDeployment(
+                    count = reservation.count,
+                    phone = reservation.device,
                     deploymentName = deploymentName
                 )
-            }
-            .merge()
```
eugene-krivobokov, line 105: `merge` function did all work in unconfined coroutines (globalscope == default dispatcher)

```kotlin
suspend fun <T> Collection<Channel<T>>.merge(): Channel<T> = Channel<T>().apply {
    forEach { channel ->
        GlobalScope.launch {
            for (item in channel) {
                send(item)
            }
        }
    }
}
```
We've closed resulting channel in the end, but cancellation was not propagated properly to the original coroutines.
Therefore, podsChannel was always open.

```kotlin
listenPodsFromDeployment(
  GlobalScope.launch {
    ....
    while (!podsChannel.isClosedForSend && pods.isNotEmpty()) {
```
It wasn't leaked only due to empty pods. After deleting deployment the `podsFromDeployment` fails and returns empty list. It breaks the loop.

This is why I've used only one channel to communicate.

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/reservation/client/local/LocalReservationClient.kt
```diff
@@ -89,19 +85,54 @@ internal class LocalReservationClient(
...
+    private fun findRemainingDevices(
+        reservation: Reservation.Data,
+        devices: Set<WorkerDevice>,
+        reserved: Set<Serial>
+    ): List<WorkerDevice> {
+        val remaining = reservation.count - reserved.size
+        return devices.filterNot { device ->
+            reserved.contains(device.id)
+        }.take(remaining)
+    }
```
eugene-krivobokov, line 132: I've fixed reservation size for local emulators.
In comparison to k8s, reservation state is local and stored in this client.

sergioboi (approved):

PR #293, opened 2020-04-01
MBS-8088 impact analysis through modulePaths and android gradle plugin 3.6.2

### LINKED ISSUE #292: UI Impact analysis through module pointers
## Problem

Android gradle plugin 3.6 makes namespacedRClass on by default, so transitive resources doesn't merge up through modules graph.
It breaks our approach to map R.id constants with it's modules as id's are no longer statically resolved.

## Proposal

We could add new field for `com.avito.android.screen.Screen`:

```kotlin
interface Screen {
    val rootId: Int
+   val module: String
}
```

`rootId` is still needed for isOpened method

`module` is a module path, like `:lib:foo`, it should point to the closest to app module containing this screen logic

Q: Will at be harder to support?
A: A bit, modules are more or less stable, but we lost isOpened guarantees (isOpened check kept rootId actual)

Q: What if someone renames module?
A: if `module` field point on nonexistent module we could fail build on analysis with descriptive error

Q: What if screen moved to another module?
A: Incorrect module pointers will lead to incorrect ui impact analysis results. Runtime check could be added to match computed rootId with module based on info from module's intermediates

## Future

We are going to left a small(and decreasing) amount of e2e tests in app module, and move component tests to library modules, so this impact analysis approach will be less effective, as libraries inherits standard "per module"
impact analysis.

Why not remove all ui impact analysis code at all?
1. We want to keep our PR times at current level before component tests goes to modules
1. Some of it's logic could be reused, for example to find new and changed tests on PR

dsvoronin:
fixes #292

- [x] check for wrong rootId <-> module link
- [x] check for nonexistent modules
- [x] check api compatibility


subprojects/android-test/ui-testing-core/src/main/kotlin/com/avito/android/screen/Screen.kt
```diff
@@ -5,32 +5,22 @@ package com.avito.android.screen
...
      * R.id of root view in hierarchy of a screen
      * Used to determine if screen is presented to user (opened)
+     */
+    val rootId: Int
+
+    /**
+     * Should be a full path to a module containing it's [rootId]
      *
-     * Also used to link "Screen" PageObject and "Screen" (android.view.View) in app code
+     * Used to link "Screen" PageObject and "Screen" (android.view.View) in app code
```
eugene-krivobokov: "used to" has the opposite meaning: https://dictionary.cambridge.org/grammar/british-grammar/used-to

We can say "Links ..." or "Connects ..."

eugene-krivobokov: My bad. We can say "It is used to ...".

subprojects/android-test/impact-analysis/src/main/kotlin/com/avito/android/impact/ImpactAnalysisAwareScreenValidator.kt
```diff
@@ -0,0 +1,60 @@
...
+         * Issue in screen configuration, most likely will be fixed with corrected modulePath or rootId
+         */
+        data class ConfigurationError(val reason: String) : Result()
+    }
+
+    /**
+     * checks impact analysis aware screen for (module <-> rootId) consistency
+     * more info: https://github.com/avito-tech/avito-android/issues/292
+     *
+     * should be called as early as possible on first screen access in tests
```
eugene-krivobokov, line 31: > should be called as early as possible on first screen access in tests

Can we validate screens in isolation and regardless of tests running?

subprojects/android-test/impact-analysis/src/test/kotlin/com/avito/android/impact/ImpactAnalysisAwareScreenValidatorTest.kt
```diff
@@ -0,0 +1,92 @@
...
+        }
+    }
+
+    @Test
+    fun `validateScreen - error - if no such module found in project`() {
+        val result = ImpactAnalysisAwareScreenValidator(
+            moduleToIds = mapOf(":one" to listOf("21356778")),
+            idsSymbolList = mapOf()
+        ).validateScreen(
+            screenClassName = "LeScreen",
```
eugene-krivobokov, line 31: `Le` - is it a french article? :fr:

dsvoronin: _oui_

subprojects/gradle/instrumentation-test-impact-analysis/src/main/kotlin/com/avito/instrumentation/impact/CopySymbolsToAssetsConfigurator.kt
```diff
@@ -0,0 +1,57 @@
...
+                    )
+                )
+            }
+
+        targetTask.configure { copy ->
+            copy.from(runtimeSymbolList)
+            librarySymbolsFiles.forEach { (modulePath, symbolFile) ->
+                copy.from(symbolFile) { spec ->
+                    spec.rename { "${modulePath.toFileName()}_$it" }
+                }
```
eugene-krivobokov, line 53: Can we copy file in place? Why do we need to delegate and postpone copying?

subprojects/android-test/impact-analysis/src/main/kotlin/com/avito/android/impact/ImpactAnalysisAwareScreenValidator.kt
```diff
@@ -0,0 +1,60 @@
...
+
+    sealed class Result {
+
+        object OK : Result()
+
+        /**
+         * Identifies that problem in the impact analysis mechanism itself,
+         * and you cannot fix it via configuration of a project (e.g. changing rootId/modulePath parameters)
+         */
+        data class ImpactAnalysisError(val reason: String) : Result()
```
sergioboi, line 19: It will be nice if we add info to reason that "error is internal is not your fault"

PR #298, opened 2020-04-04
MBS-8088 remove testFixtures usage, fixes #295

### LINKED ISSUE #295: Should we abandon testFixtures usage?
There is a problem that really hurts our productivity:

https://github.com/avito-tech/avito-android/blob/1016aeca15b71f7c7190b3d959bcbae6731c4ec9/gradle.properties#L7-L11

Should we drop it for more straightforward approach, at least temporary?

First thing comes to mind is just place fixtures in runtimeClasspath and package it this regular code?
It looks dirty, but we can use some additional instruments to point out it's "testing only" purpose.
Like maybe https://kotlinlang.org/docs/reference/opt-in-requirements.html custom annotation, or maybe you know better ways.

dsvoronin:


dsvoronin:
thought about tech radar page, but decided testFixtures is too small feature for that format, link to the issue will be enough

^ debatable ^

gradle.properties
```diff
@@ -4,12 +4,6 @@ org.gradle.parallel=true
 org.gradle.caching=true
 # mandatory from agp 3.6
 android.useAndroidX=true
-# we use testFixtures, and it has problems when project includes android modules
-# temporary set this to false in `~/.gradle/gradle.properties` to work with modules that uses testFixtures as a workaround
-# https://issuetracker.google.com/issues/139438142
```
eugene-krivobokov, line 9: Should we keep knowledge about this decision somewhere?

eugene-krivobokov: I see, https://github.com/avito-tech/avito-android/pull/298#issuecomment-609077272
Tech radar is good for me.
We can write it later in case of new questions.

eugene-krivobokov (approved):

PR #312, opened 2020-04-12
MBS 8128 Can pass APK to runner manually

eugene-krivobokov:


docs/content/docs/test/Runner.md
```diff
@@ -15,7 +15,43 @@ It can do the following:
...
+extensions.getByType<GradleInstrumentationPluginConfiguration>().apply {
+    // all options goes here
+}
+
+```
+
+### Test APK
+
+Plugin builds APKs on his own by default.\
+If for any reason you have to build APK externally, you can pass files manually:
```
eugene-krivobokov, line 41: I see probable reasons to pass APK:
- It has been built externally on another host due to performance
- It has been transformed due to security reasons and we can't get a final artifact from AGP. Avito's signer service as an example.

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/InstrumentationTestsTask.kt
```diff
@@ -65,9 +65,6 @@ abstract class InstrumentationTestsTask @Inject constructor(
     @Input
     val buildUrl = objects.property<String>()
 
-    @Input
-    val testedVariantName = objects.property<String>()
```
eugene-krivobokov, line 69: Unused

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/InstrumentationTestsPlugin.kt
```diff
@@ -254,9 +262,9 @@ class InstrumentationTestsPlugin : Plugin<Project> {
                                 }
 
                                 val isFullTestSuite = gitState.map {
-                                        it.isOnDefaultBranch
-                                                && instrumentationConfiguration.impactAnalysisPolicy is ImpactAnalysisPolicy.Off
-                                    }
+                                    it.isOnDefaultBranch
+                                        && instrumentationConfiguration.impactAnalysisPolicy is ImpactAnalysisPolicy.Off
```
matzuk, line 266: hm, where is the impact analysis module in the general scheme of Avito runner?

matzuk:
@eugene-krivobokov yep, it's a thing that we have discussed in an issue (83)

docs/content/docs/test/Runner.md
dsvoronin: i think this comment could be a part of documentation

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/InstrumentationTestsPlugin.kt
dsvoronin: it is a part of the tests filtration; it is optional now, but we can possibly not use it as a dependency altogether

dsvoronin (approved):

docs/content/docs/test/Runner.md
eugene-krivobokov: It is a documentation

PR #318, opened 2020-04-13
MBS-8088 impact analysis screenToModule relation via psi

### LINKED ISSUE #292: UI Impact analysis through module pointers

dsvoronin:
fixes #292, but with new implementation, just analyzing psi of Screen files to get package modules


subprojects/gradle/impact-shared/src/main/java/com/avito/impact/util/AndroidManifestPackageParser.kt
```diff
@@ -0,0 +1,13 @@
+package com.avito.impact.util
+
+import java.io.File
+
+object AndroidManifestPackageParser {
+
+    private val manifestPackagePattern = Regex(" package=\"([a-zA-Z0-9._]+)\"")
```
eugene-krivobokov:
```suggestion
    private val manifestPackagePattern = Regex(" package=\"([^"]+)\"")
```

I see no reasons to know about package valid value in this case.

subprojects/gradle/instrumentation-test-impact-analysis/src/main/kotlin/com/avito/instrumentation/impact/TestBytecodeAnalyzeAction.kt
```diff
@@ -303,22 +315,28 @@ abstract class TestBytecodeAnalyzeAction : WorkAction<TestBytecodeAnalyzeAction.
...
-        val duplicateCheckResult = DuplicateIdChecker.Impl(config.unknownRootId.get()).check(screenToIds)
-        if (duplicateCheckResult.hasDuplicates) {
-            ciLogger.info("There are duplicate id's: ${duplicateCheckResult.ids}")
-        }
+    private fun getScreenToModule(projectModules: Collection<InternalModule>): Collection<ScreenToModulePath> {
+
+        val sourceSets = projectModules.flatMap { it.androidTestConfiguration.sourceSets() }
 
-        return screenToIds.toMap()
+        ciLogger.info("projectModules=$projectModules")
```
eugene-krivobokov: debug ?

subprojects/gradle/instrumentation-test-impact-analysis/src/main/kotlin/com/avito/instrumentation/impact/TestBytecodeAnalyzeAction.kt
```diff
@@ -303,22 +315,28 @@ abstract class TestBytecodeAnalyzeAction : WorkAction<TestBytecodeAnalyzeAction.
...
+
+        val androidModules = projectModules.filter {
+            it.project.pluginManager.hasPlugin("com.android.library") || it.project.pluginManager.hasPlugin("com.android.application")
```
eugene-krivobokov:
```suggestion
            it.project.isAndroid()
```

We have an extension already.

subprojects/gradle/instrumentation-test-impact-analysis/src/main/kotlin/com/avito/instrumentation/impact/metadata/ModulePath.kt
```diff
@@ -0,0 +1,11 @@
...
+
+data class ModulePath(val path: String) {
+    init {
+        require(path.matches(validModulePath)) { "$path is not a valid gradle module path" }
+    }
+}
+
+fun ModulePath.toFileName() = path.drop(1).replace(":", "+")
+
+private val validModulePath = Regex("(:[a-zA-Z1-9\\-]+(_[a-zA-Z1-9\\-])?)*")
```
eugene-krivobokov:
```suggestion
private val validModulePath = Regex("(:[\w\d\\-]+(_[\w\d\\-])?)*")
```
It's good to use character classes for readability.

subprojects/gradle/instrumentation-test-impact-analysis/src/main/kotlin/com/avito/instrumentation/impact/metadata/ModulePath.kt
```diff
@@ -0,0 +1,11 @@
+package com.avito.instrumentation.impact.metadata
+
+data class ModulePath(val path: String) {
+    init {
+        require(path.matches(validModulePath)) { "$path is not a valid gradle module path" }
+    }
+}
+
+fun ModulePath.toFileName() = path.drop(1).replace(":", "+")
```
eugene-krivobokov: Can paths clash after dropping a caught group?

eugene-krivobokov (changes requested):

subprojects/gradle/instrumentation-test-impact-analysis/src/main/kotlin/com/avito/instrumentation/impact/TestBytecodeAnalyzeAction.kt
dsvoronin: yep, thanks

subprojects/gradle/impact-shared/src/main/java/com/avito/impact/util/AndroidManifestPackageParser.kt
dsvoronin: i agree

sergioboi: Could we use Xml Parser?

subprojects/gradle/instrumentation-test-impact-analysis/src/main/kotlin/com/avito/instrumentation/impact/metadata/ModulePath.kt
sergioboi: Could we use Gradle API? We may get project dir by a path. Parsing string, dropping ':' looks fragile.
Not a blocker

subprojects/gradle/instrumentation-test-impact-analysis/src/main/kotlin/com/avito/instrumentation/impact/metadata/ModulePath.kt
sergioboi: Yet another great part to delegate to Gradle

subprojects/gradle/instrumentation-test-impact-analysis/src/main/kotlin/com/avito/instrumentation/impact/metadata/MetadataParser.kt
```diff
@@ -0,0 +1,75 @@
...
+
+                                    val importLines = ktFile.importDirectives.map { it.importedFqName.toString() }
+                                    result[fullClassName] = getPackageName(expression.text, importLines)!!
+                                }
+                            }
+                        }
+                }
+        }
+
+        ciLogger.info("Analyzed screen classes in ${System.currentTimeMillis() - startTime}ms")
```
sergioboi: debug

subprojects/gradle/instrumentation-test-impact-analysis/src/main/kotlin/com/avito/instrumentation/impact/metadata/MetadataParser.kt
```diff
@@ -0,0 +1,75 @@
...
+                            val fullClassName = if (packageName.isBlank()) {
+                                className
+                            } else {
+                                "$packageName.$className"
+                            }
+
+                            it.body?.properties?.forEach { property ->
+                                if (property.name == fieldName) {
+
+                                    val expression = property.children.filterIsInstance<KtExpression>().single()
```
sergioboi: Could you create a task for assigning via functions and getters?

sergioboi (approved):

subprojects/gradle/instrumentation-test-impact-analysis/src/main/kotlin/com/avito/instrumentation/impact/metadata/ModulePath.kt
eugene-krivobokov: As far as I see, it's a custom format similar to the gradle path.

eugene-krivobokov (approved):

subprojects/gradle/impact-shared/src/main/java/com/avito/impact/util/AndroidManifestPackageParser.kt
dsvoronin: > Could we use Xml Parser?

yes, but we don't need deep file structure knowledge. i can optimize it reading line by line

subprojects/gradle/instrumentation-test-impact-analysis/src/main/kotlin/com/avito/instrumentation/impact/metadata/ModulePath.kt
dsvoronin: i used https://github.com/gradle/gradle/blob/master/subprojects/base-services/src/main/java/org/gradle/util/Path.java , seems like what we need here

subprojects/gradle/instrumentation-test-impact-analysis/src/main/kotlin/com/avito/instrumentation/impact/metadata/MetadataParser.kt
dsvoronin: thanks, changed.

subprojects/gradle/instrumentation-test-impact-analysis/src/main/kotlin/com/avito/instrumentation/impact/metadata/MetadataParser.kt
dsvoronin: supported it in new commit (kinda, maybe missed some corner cases, will see)

subprojects/gradle/instrumentation-test-impact-analysis/src/main/kotlin/com/avito/instrumentation/impact/metadata/ModulePath.kt
dsvoronin: used gradle one

subprojects/gradle/instrumentation-test-impact-analysis/src/main/kotlin/com/avito/instrumentation/impact/TestBytecodeAnalyzeAction.kt
dsvoronin: thanks, fixed

subprojects/gradle/instrumentation-test-impact-analysis/src/main/kotlin/com/avito/instrumentation/impact/metadata/ModulePath.kt
eugene-krivobokov: It doesn't have validation. But without the user's input it's not needed.

PR #325, opened 2020-04-15
MBS-8053 Change filter output format

sergioboi:


dsvoronin (approved):

dsvoronin:
This pr lacks of explanation, i'm not entirely sure if i understand the intention correctly

PR #324, opened 2020-04-15
MBS-8032 Configure build checks by extension

eugene-krivobokov:
I've moved build check plugin configuration from Gradle properties to the extension.
It gives type-safe and simpler DSL for clients.
If a client wants to change behavior by command-line flags it can be easily done in build.gradle without changes in the plugin. 

What was before:
```
avito.build.paramCheck.enabled=true
avito.build.failOnSdkMismatch=true
avito.build.androidJar.revision=4
```

What will be with the new config:
```groovy
buildChecks {

    androidSdk {
        it.compileSdkVersion = versions.compileSdk
        it.revision = 4
    }
    javaVersion {
        it.version = JavaVersion.VERSION_1_8
    }
}
```

PR #326, opened 2020-04-15
MBS-8032 Check unique R classes

eugene-krivobokov:
Merge after #324

### DIFF
--- a/subprojects/android-test/test-app/gradle.properties
+++ b/subprojects/android-test/test-app/gradle.properties
@@ -1,4 +1,2 @@
-android.enableR8=true
-
 test-app.versionCode=1
 test-app.versionName=1.0
--- a/subprojects/gradle/buildchecks/src/main/kotlin/com/avito/android/plugin/build_param_check/BuildParamCheckPlugin.kt
+++ b/subprojects/gradle/buildchecks/src/main/kotlin/com/avito/android/plugin/build_param_check/BuildParamCheckPlugin.kt
@@ -109,6 +109,18 @@ open class BuildParamCheckPlugin : Plugin<Project> {
                 dependsOn(task)
             }
         }
+        if (checks.hasInstance<Check.UniqueRClasses>()) {
+            check(project.pluginManager.hasPlugin("com.avito.android.impact")) {
+                "build check for unique R classes requires 'com.avito.android.impact' plugin"
+            }
+            val task = project.tasks.register<UniqueRClassesTask>("checkUniqueAndroidPackages") {
+                group = "verification"
+                description = "Verify unique R classes"
+            }
+            checkBuildEnvironment {
+                dependsOn(task)
+            }
+        }
         if (checks.hasInstance<Check.MacOSLocalhost>() && isMac()) {
             val task = project.tasks.register<MacOSLocalhostResolvingTask>("checkMacOSLocalhostResolving") {
                 group = "verification"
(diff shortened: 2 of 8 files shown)


subprojects/android-test/test-app/gradle.properties
```diff
@@ -1,4 +1,2 @@
-android.enableR8=true
```
eugene-krivobokov, line 1: Useless since AGP 3.6

```
WARNING: The option 'android.enableR8' is deprecated and should not be used anymore.
It will be removed in a future version of the Android Gradle plugin, and will no longer allow you to disable R8.
```

subprojects/gradle/buildchecks/src/main/kotlin/com/avito/android/plugin/build_param_check/BuildParamCheckPlugin.kt
```diff
@@ -112,6 +112,18 @@ open class BuildParamCheckPlugin : Plugin<Project> {
...
+            check(project.pluginManager.hasPlugin("com.avito.android.impact")) {
+                "build check for unique R classes requires 'com.avito.android.impact' plugin"
+            }
+            val task = project.tasks.register<UniqueRClassesTask>("checkUniqueAndroidPackages") {
+                group = "verification"
+                description = "Verify unique R classes"
+            }
+            checkBuildEnvironment {
+                dependsOn(task)
+            }
```
eugene-krivobokov, line 122: One additional consideration.
I don't rely on `android.namespacedRClass` setting because this issue is relevant to both states.

dsvoronin (approved):

subprojects/gradle/buildchecks/src/main/kotlin/com/avito/android/plugin/build_param_check/BuildParamCheckPlugin.kt
```diff
@@ -16,70 +17,118 @@ import org.gradle.api.Plugin
...
+        if (checks.hasInstance<Check.UniqueRClasses>()) {
+            check(project.pluginManager.hasPlugin("com.avito.android.impact")) {
+                "build check for unique R classes requires 'com.avito.android.impact' plugin"
+            }
+            val task = project.tasks.register<UniqueRClassesTask>("checkUniqueAndroidPackages") {
+                group = "verification"
+                description = "Verify unique R classes"
+            }
+            checkBuildEnvironment {
+                dependsOn(task)
```
sergioboi, line 121: Did you measure of time it spends? Not only a new one, the whole configuration phase with and without checks

eugene-krivobokov: We only register a new task here.

The plugin does some checks at the configuration phase. Gradle build scan shows up to 0.5s for this plugin. Definitely, we can improve it, but not in this PR.

eugene-krivobokov: `UniqueRClassesTask` itself lasts about 5s

sergioboi: Do checks run in execution phase?

eugene-krivobokov: Yes, they are Gradle tasks

```
checkBuildEnvironment { <--- task provider, "{" is a function for ".configure"
                dependsOn(task) <--- task dependency
```

sergioboi: Could we make them cacheble? Or is it overwork?

eugene-krivobokov: It would be good, but It depends.

For instance, checking localhost in macOs doesn't have inputs in the project. It's hard to say when to cache.

Another example: unique R classes. The input of the task is the whole project structure and manifests.

The first step is to migrate to the worker API. I think It will be enough.

PR #334, opened 2020-04-17
MBS-8155 support tests from android libraries in test runner

dsvoronin:
Test covers configuration phase, but no full integration yet


subprojects/gradle/performance/src/main/kotlin/com/avito/performance/PerformancePlugin.kt
```diff
@@ -87,6 +92,8 @@ open class PerformancePlugin : Plugin<Project> {
 
                     project.tasks.instrumentationTask(performanceConfig.name) { instrumentationTask ->
 
+                        logger.info("Based on instrumentation task: $instrumentationTask")
```
eugene-krivobokov, line 95: debug?

eugene-krivobokov (approved):

PR #335, opened 2020-04-19
MBS-7958 Add test-instrumentation-runner without avito specific depen…

sergioboi:
…dencies


subprojects/android-test/test-instrumentation-runner/src/main/kotlin/com/avito/android/runner/InstrumentationTestRunner.kt
```diff
@@ -0,0 +1,32 @@
...
+                return DefaultTestInstrumentationContext(
+                    errorsReporter = LogErrorsReporter()
+                )
+            }
+        }
+    }
+
+    override fun onCreate(arguments: Bundle) {
+        val context = createFactory().create(arguments)
+        if (context != null) {
```
eugene-krivobokov, line 22: Should we fail if null?

subprojects/android-test/test-instrumentation-runner/src/main/kotlin/com/avito/android/runner/Context.kt
```diff
@@ -0,0 +1,5 @@
+package com.avito.android.runner
+
+interface Context {
```
eugene-krivobokov, line 3: What is the responsibility of this abstraction?
What will it know besides errors reporter?

subprojects/android-test/test-instrumentation-runner/src/main/kotlin/com/avito/android/runner/InstrumentationTestRunner.kt
sergioboi: It returns null when we in orchestrator execution

subprojects/android-test/test-instrumentation-runner/src/main/kotlin/com/avito/android/runner/Context.kt
sergioboi: It is a service locator for the future. I consider to use it for holding all needed dependencies

eugene-krivobokov (approved):

PR #347, opened 2020-04-23
MBS-8098 Proposal extract DeployStep.kt

sergioboi:
DeployStep.kt accessible only from `release {}`

### DIFF
--- a/subprojects/gradle/cd/src/main/kotlin/com/avito/ci/ReleaseBuildStepsExtension.kt
+++ b/subprojects/gradle/cd/src/main/kotlin/com/avito/ci/ReleaseBuildStepsExtension.kt
@@ -1,0 +1,15 @@
+package com.avito.ci
+
+import com.avito.ci.steps.release.DeployStep
+import groovy.lang.Closure
+import org.gradle.api.model.ObjectFactory
+
+
+class ReleaseBuildStepsExtension(
+    objects: ObjectFactory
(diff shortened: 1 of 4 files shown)


subprojects/gradle/cd/src/main/kotlin/com/avito/ci/ReleaseBuildStepsExtension.kt
```diff
@@ -0,0 +1,15 @@
+package com.avito.ci
+
+import com.avito.ci.steps.release.DeployStep
+import groovy.lang.Closure
+import org.gradle.api.model.ObjectFactory
+
+
+class ReleaseBuildStepsExtension(
+    objects: ObjectFactory
+) : BuildStepListExtension("release", objects) {
```
eugene-krivobokov, line 10: > deploy step is accessible only from release {}

Why do we need this restriction?

I'd like to remove built-in builds from the Cd plugin (see MBS-7948).

sergioboi: Let discuss MBS-7948 and then return here.

sergioboi: We decided to leave the abilities of flexible steps configuration.
Because we have a lot of scenarios where our consumers need this flexibility.

PR #350, opened 2020-04-24
Mbs 7824 fast check no flaky

int02h:
Add new filter and related configuration that runs UI tests excepting E2E and `@Flaky` tests.

### DIFF
--- a/docs/content/docs/Infrastructure.md
+++ b/docs/content/docs/Infrastructure.md
@@ -175,4 +175,4 @@ You will see branches from both repositories:
 - By default, build uses develop from github against develop from avito
 - If you pick a branch from avito, it will run against develop on github
 - If you pick a branch from github, it will run against develop on avito
-- (UNTESTED) To build both projects of special branch, they should have the same name
+- To build both projects of special branch, they should have the same name
--- a/docs/content/docs/test/Run.md
+++ b/docs/content/docs/test/Run.md
@@ -14,7 +14,7 @@ type: docs
 При создании Pull Request'а ваша ветка ребейзится на целевую ветку и автоматически запускается
 [`fastCheck`](http://links.k.avito.ru/PV), объявленный в `avito/build.gradle`:
 - Юнит тесты
-- Инструментальные тесты:
+- Инструментальные тесты (кроме помеченных аннотацией [`@Flaky`]({{< ref "/docs/test/FlakyTests.md" >}})):
   - Перформанс
   - Компонентные
   - [Скриншотные]({{< ref "/docs/test/ScreenshotTesting.md" >}})
--- a/subprojects/gradle/instrumentation-tests-default-config/src/main/kotlin/com/avito/instrumentation/InstrumentationDefaultConfigPlugin.kt
+++ b/subprojects/gradle/instrumentation-tests-default-config/src/main/kotlin/com/avito/instrumentation/InstrumentationDefaultConfigPlugin.kt
@@ -57,6 +57,11 @@ class InstrumentationDefaultConfigPlugin : Plugin<Project> {
                     it.fromSource.includeByAnnotations(TestsFilter.uiNoE2E.annotatedWith)
                     it.fromRunHistory.excludePreviousStatuses(setOf(RunStatus.Success, RunStatus.Manual))
                 }
+                filters.register("uiNoE2eNoFlaky") {
+                    it.fromSource.includeByAnnotations(TestsFilter.uiNoE2E.annotatedWith)
+                    it.fromSource.excludeByAnnotations(TestsFilter.flaky.annotatedWith)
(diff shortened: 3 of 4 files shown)


docs/content/docs/Infrastructure.md
```diff
@@ -175,4 +175,4 @@ You will see branches from both repositories:
 - By default, build uses develop from github against develop from avito
 - If you pick a branch from avito, it will run against develop on github
 - If you pick a branch from github, it will run against develop on avito
-- (UNTESTED) To build both projects of special branch, they should have the same name
```
int02h, line 178: I've tested it. Works as expected.

eugene-krivobokov: The same

eugene-krivobokov: I see only one inconvenience If I want to check backward compatibility also.

There are at least two options:
- Remove branch from avito repository
- Make a new branch from feature-branch in github

Can you please add this case to the documentation?

docs/content/docs/test/Run.md
```diff
@@ -14,7 +14,7 @@ type: docs
 При создании Pull Request'а ваша ветка ребейзится на целевую ветку и автоматически запускается
 [`fastCheck`](http://links.k.avito.ru/PV), объявленный в `avito/build.gradle`:
 - Юнит тесты
-- Инструментальные тесты:
+- Инструментальные тесты (кроме помеченных аннотацией [`@Flaky`]({{< ref "/docs/test/FlakyTests.md" >}})):
```
eugene-krivobokov, line 17: I'm thinking about renaming `Flaky` to `Unstable`. 
Just a thought after today's case with Google Maps.

int02h: `Unstable` was one of the first naming variants. But I decided that Flaky is a commonly-known term.

docs/content/docs/Infrastructure.md
int02h: I don't think that this information is somehow related to this document. The doc only describes how to run infra against Avito. It doesn't describe the process itself or how to perform custom checks.

subprojects/gradle/instrumentation-tests-default-config/src/main/kotlin/com/avito/instrumentation/InstrumentationDefaultConfigPlugin.kt
```diff
@@ -57,6 +57,11 @@ class InstrumentationDefaultConfigPlugin : Plugin<Project> {
                     it.fromSource.includeByAnnotations(TestsFilter.uiNoE2E.annotatedWith)
                     it.fromRunHistory.excludePreviousStatuses(setOf(RunStatus.Success, RunStatus.Manual))
                 }
+                filters.register("uiNoE2eNoFlaky") {
+                    it.fromSource.includeByAnnotations(TestsFilter.uiNoE2E.annotatedWith)
+                    it.fromSource.excludeByAnnotations(TestsFilter.flaky.annotatedWith)
+                    it.fromRunHistory.excludePreviousStatuses(setOf(RunStatus.Success, RunStatus.Manual))
+                }
```
eugene-krivobokov, line 64: What about reusing filters?
We've talked about something like `filter.initWith(anotherFilter)` to copy an existing filter.
It'd better show the intention to add only one annotation without duplication.

eugene-krivobokov (changes requested):

subprojects/gradle/instrumentation-tests-default-config/src/main/kotlin/com/avito/instrumentation/InstrumentationDefaultConfigPlugin.kt
eugene-krivobokov: not a blocker

eugene-krivobokov (approved):

docs/content/docs/Infrastructure.md
eugene-krivobokov: > The doc only describes how to run infra against Avito

Not only. This doc describes publishing and testing changes: 
**Publishing a new release > CI integration tests against avito**

subprojects/gradle/instrumentation-tests-default-config/src/main/kotlin/com/avito/instrumentation/InstrumentationDefaultConfigPlugin.kt
int02h: I've tried to make something like `initWith(anotherFilter)` but current Filter implementation always rewrite its internal Sets like the following:
```kotlin
fun include(values: Set<T>) {
        included = values
}
```

So if I even init some filter with another I won't be able to include or exclude something because I'll overwrite values copied from another filter.

@sboishtyan fyi

PR #351, opened 2020-04-24
MBS-8283 Fix recursion in `isScreenOpened`

sergioboi:


subprojects/android-test/ui-testing-core/src/main/kotlin/com/avito/android/screen/ScreenChecks.kt
```diff
@@ -36,6 +36,7 @@ class SimpleScreenChecks(override val screen: Screen) : ScreenChecks
 
 open class StrictScreenChecks(override val screen: Screen) : ScreenChecks {
 
+    // Delete?
```
sergioboi: @eugene-krivobokov Could you explain the neediness of that function?

subprojects/android-test/ui-testing-core/src/main/kotlin/com/avito/android/test/InteractionContext.kt
```diff
@@ -30,11 +30,13 @@ class SimpleInteractionContext(
     private val precondition: () -> Unit = {}
 ) : InteractionContext {
 
+    private var inPrecondition = false
```
sergioboi, line 33: @dsvoronin I supposed that all checks happen on UI thread.
Or do we need concurrency safety?

subprojects/android-test/ui-testing-core/src/main/kotlin/com/avito/android/screen/ScreenChecks.kt
eugene-krivobokov: It is for backward compatibility with `ScreenChecks` interface.
Screens that haven't supported strict checks yet, use legacy `SimpleChecks`.

```kotlin
interface Screen {

  val checks: ScreenChecks
        get() = SimpleScreenChecks(this) // TODO: replace by StrictScreenChecks MBS-7204
```

We need this to [support strict checks on screens one by one](https://avito-tech.github.io/avito-android/docs/test_framework/testframework/#screenchecks). After this, we will use only one function.

eugene-krivobokov (approved):

subprojects/android-test/test-app/src/androidTest/kotlin/com/avito/android/ui/test/PageObjectTest.kt
```diff
@@ -39,12 +39,20 @@ class PageObjectTest {
     }
 
     @Test
-    fun element__notFound__byDefaultMatcherInTheWrongParent() {
+    fun parent__assertionError__byDefaultMatcherInTheWrongParent() {
         rule.launchActivity(PageObjectActivity.intent(R.layout.page_object_2)) // twin
 
-        exception.expect(NoMatchingViewException::class.java)
+        exception.expect(junit.framework.AssertionFailedError::class.java)
```
sergioboi, line 45: It was hard to realize how it worked before. And why now it fails with `junit.framework.AssertionFailedError::class.java`. @eugene-krivobokov  Could you check the correctness of that fix?

PR #359, opened 2020-04-29
MBS-7920 Change InstrumentationTask's creation

sergioboi:
**Summary**
- Create all clients InstrumentationTestTask
- Move validation of Kubernetes credentials presence to configuration phase if clients add InstrumentationConfiguration in the build script
- Create InstrumentationTestTask dependent on k8s in plugin only if user set Kubernetes credentials
- If clients add an InstrumentationConfiguration dependent on k8s they must put that logic to the `if clause` in the build script or that logic will fail without Kubernetes credentials


subprojects/gradle/kubernetes/src/main/kotlin/com/avito/utils/gradle/Kubernetes.kt
```diff
@@ -59,3 +65,25 @@ private fun oauthTokenProvider(config: File): OAuthTokenProvider {
 
     return OAuthTokenProvider { kubeConfig.accessToken }
 }
+
+val Project.kubernetesCredentials: KubernetesCredentials
+    get() {
+        val context = getOptionalStringProperty("kubernetesContext", nullIfBlank = true)
```
eugene-krivobokov, line 71: We shouldn't branch logic by one specific parameter from authentication parameters.

Service account = `kubernetesToken` + `kubernetesCaCertData` + `kubernetesUrl`
Open id auth = `kubernetesContext` + `kubernetesCaCertFile`

If I don't have `kubernetesContext`, it doesn't imply that I want to use service account.

eugene-krivobokov: My proposal is to declare auth credential by extension explicitly. 
A user has to tell what to use and pass required parameters.
Not in this PR

eugene-krivobokov (approved):

subprojects/gradle/kubernetes/src/main/kotlin/com/avito/utils/gradle/Kubernetes.kt
sergioboi: Discussed by voice. We decided to implement this logic like a part of emulator target refactoring

PR #365, opened 2020-04-30
MBS-8348 Consider deployment to Qapps in CD contract

eugene-krivobokov:


subprojects/gradle/cd/src/main/kotlin/com/avito/ci/steps/UploadToQapps.kt
```diff
@@ -19,13 +22,19 @@ class UploadToQapps(context: String, artifactsConfiguration: ArtifactsConfigurat
...
+
         return artifactsMap.map { (_, output) ->
             when (output) {
                 is Output.ApkOutput ->
                     project.tasks.qappsTaskProvider(output.variantName)
                         .apply {
                             configure {
                                 it.getApk().set(File(output.path))
+                                it.releaseChain.set(uploadCondition.isReleaseChain())
+                                it.onlyIf {
```
sergioboi, line 35: We can simply don't create a task and remove `onlyIf clause`

subprojects/gradle/cd/src/main/kotlin/com/avito/ci/steps/UploadToQapps.kt
```diff
@@ -19,13 +22,19 @@ class UploadToQapps(context: String, artifactsConfiguration: ArtifactsConfigurat
...
+                                    uploadCondition.canUpload()
```
sergioboi, line 36: Do we send PR artifacts to qapps?

sergioboi (approved):

subprojects/gradle/cd/src/main/kotlin/com/avito/ci/steps/UploadToQapps.kt
eugene-krivobokov: I prefer `onlyIf` in similar cases due to easier diagnostics in case of problems.
Let's imagine a typical bug: I don't see artifacts in Artifactory. The first thing to check is the task. I'll see that this task is skipped due to `onlyIf` clause. I know the reason and there to find it.
When I don't see the task, I have to check more code that creates it.

subprojects/gradle/cd/src/main/kotlin/com/avito/ci/steps/UploadToQapps.kt
eugene-krivobokov: No, at least for now. 
There is no conflict in the requirements. Cd build config can only disable uploading.

subprojects/gradle/cd/src/main/kotlin/com/avito/ci/steps/UploadToQapps.kt
sergioboi: Interesting point, I agree. Let write tests on that scenario

subprojects/gradle/cd/src/main/kotlin/com/avito/ci/steps/UploadToQapps.kt
```diff
@@ -34,3 +43,25 @@ class UploadToQapps(context: String, artifactsConfiguration: ArtifactsConfigurat
...
+internal class UploadToQappsCondition(
+    val config: CdBuildConfig?
+) {
+
+    fun isReleaseChain(): Boolean {
+        return findDeployment()?.isRelease ?: false
+    }
+
+    fun canUpload(): Boolean {
+        if (config == null) return true
```
int02h, line 56: Are you sure that this method should return `true` when `config` is absent?

eugene-krivobokov: Yes, it means that this build is not a part of a CD release.
In this case, uploading to QApss works as described in build script.

subprojects/gradle/cd/src/main/kotlin/com/avito/ci/steps/UploadToQapps.kt
eugene-krivobokov: Added `UploadToQappsTest`

subprojects/gradle/upload-cd-build-result/src/main/kotlin/com/avito/cd/DeploymentDeserializer.kt
```diff
@@ -12,7 +12,10 @@ internal object DeploymentDeserializer : JsonDeserializer<CdBuildConfig.Deployme
             "google-play" -> {
                 context.deserialize<CdBuildConfig.Deployment.GooglePlay>(json, CdBuildConfig.Deployment.GooglePlay::class.java)
             }
-            else -> throw IllegalArgumentException("Value of field \"type\":$type incorrect")
+            "qapps" -> {
+                context.deserialize<CdBuildConfig.Deployment.Qapps>(json, CdBuildConfig.Deployment.Qapps::class.java)
+            }
+            else -> throw IllegalArgumentException("Unsupported deployment \"type\": $type")
```
eugene-krivobokov: We have to allow for unknown deployments to deploy backend in advance.
I'll do it it another PR.

int02h (approved):

subprojects/gradle/upload-cd-build-result/src/main/kotlin/com/avito/cd/DeploymentDeserializer.kt
sergioboi: Why will backend send an unknown deployment?

eugene-krivobokov: We need to deploy infrastructure and backend independently because we have two dependent changes.
Each independent system must preserve backward compatibility for one version, at least.

sergioboi: Understand, It needs because we can't wait for two versions

eugene-krivobokov: https://github.com/avito-tech/avito-android/pull/393

eugene-krivobokov: We've discussed this issue with @sboishtyan. 
Decided to use `schemaVersion` to maintain strict contract.
See CdBuildConfigValidator

sergioboi (approved):

PR #368, opened 2020-05-01
MBS-8371 Fix build checks without buildToolsVersion system property

eugene-krivobokov:


subprojects/gradle/android/src/main/kotlin/com/avito/android/AndroidSdk.kt
```diff
@@ -56,18 +58,16 @@ class AndroidSdk(
             return ExistingDirectory.Impl(dir)
         }
 
-    private val apkSigner = ApkSigner.Impl(buildToolsPath, processRunner)
```
eugene-krivobokov, line 59: The problem: build check plugin fails on a new project. 
It is caused by an implicit reading of `buildToolsVersion` system properly.
We could ask to add the system property. This is possible only in `settings.gradle` before plugins applying. But this is inconvenient and unexpected behavior. We don't need this version in build checks at all.

It seems that a better solution would be to get rid of system properties and environment variables, especially in common utilities.
They are hard for testing. Gradle test kit can override only [environment variables](https://docs.gradle.org/current/javadoc/org/gradle/testkit/runner/GradleRunner.html#withEnvironment-java.util.Map-), but not properties. In junit tests we have only bad options with dirty reflection or races in a real state. Wrapper for `System` class won't solve the problem with implicit dependency on properties.

sergioboi: For unit tests, we can configure system properties in the test task.

subprojects/gradle/android/src/main/kotlin/com/avito/android/AndroidSdk.kt
```diff
@@ -56,18 +58,16 @@ class AndroidSdk(
...
 
     private val keytool = KeyTool(processRunner)
 
     val androidHome: ExistingDirectory
-        get() = ExistingDirectory.Impl(File(System.getenv("ANDROID_HOME")
-            ?: androidHomeFromLocalProperties(project.rootProject.file("local.properties")) {
-                logger.error(it)
-            }
-            ?: error("Can't resolve ANDROID_HOME")))
+        get() = ExistingDirectory.Impl(project.androidHome)
```
sergioboi, line 70: So I didn't get what did you do? What is the `project.androidHome`?

sergioboi (approved):

subprojects/gradle/android/src/main/kotlin/com/avito/android/AndroidSdk.kt
eugene-krivobokov: Technically, yes, we can. But it is a shared state for other tests. It's like a singleton on Gradle level.

subprojects/gradle/android/src/main/kotlin/com/avito/android/AndroidSdk.kt
eugene-krivobokov: It's the same function. It was a duplicate.

The difference is in another part. I've made the property lazy to avoid this logic in a configuration phase.

sergioboi: I can't see a difference on laziness

eugene-krivobokov: Laziness are only in `apkSigner` and `aapt` properties.
Sorry, I thought that the first question was more general.

PR #370, opened 2020-05-02
MBS-7857 Extract a build trace plugin

eugene-krivobokov:
I've extracted a build trace into a separate plugin to make it reusable.
No changes in its configuration or behavior. I'll do it independently.


settings.gradle.kts
```diff
@@ -132,6 +134,7 @@ pluginManagement {
                 gradlePluginPortal()
             }
             filter {
+                includeGroup("com.gradle")
```
eugene-krivobokov, line 137: To use a gradle build scan

subprojects/gradle/build-trace/src/main/kotlin/com/avito/android/build_trace/BuildTracePlugin.kt
```diff
@@ -0,0 +1,43 @@
...
+        if (isBuildTraceEnabled(project)) {
+            GradleCollector.initialize(
+                project, listOf(
+                    buildTraceConsumer(project)
+                )
+            )
+        }
+    }
+
+    private fun buildTraceConsumer(project: Project): MetricsConsumer = BuildTraceConsumer(
```
sergioboi, line 29: TODO add possibility to add your own consumer.

eugene-krivobokov: Maybe, If someone will ask for this flexibility.

sergioboi:
I tried to get into changes but couldn't
Let me ask some questions:
- Now we have 3 modules: build-metrics, build-trace, gradle-profile. And how it was before? What is a meaning of each module? Why do we need all three separate modules?

eugene-krivobokov:
> I tried to get into changes but couldn't
> Let me ask some questions:
> 
> * Now we have 3 modules: build-metrics, build-trace, gradle-profile. And how it was before? What is a meaning of each module? Why do we need all three separate modules?

It was only one plugin - `build-metrics`. It had multiple and unrelated responsibilities. For example, it sends avito-specific build metrics to statsd.

I've extracted build trace functionality into a separate plugin - `build-trace`.
`gradle-profile` is an internal module to reuse utility code for Gradle.

sergioboi (approved):

PR #374, opened 2020-05-07
MBS-8391 Fix failures on local instrumentation tasks run

sergioboi:
InstrumentationTestImpactAnalysisPlugin.kt need on local runs
GitState can't be empty, it is needed on local runs
EnvArgs can't be Stub, it is needed on local runs

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 test_build_type?=debug
 infra?=
-ci?=true
+ci?=false
 log_level?=-q
 
 localFilter?=
@@ -29,6 +29,10 @@ params +=-PtestBuildType=$(test_build_type)
 params +=-Pci=$(ci)
 
 params +=$(log_level)
+
+ifeq ($(gradle_debug),true)
+params +=-Dorg.gradle.jvmargs='-Xdebug -Xrunjdwp:transport=dt_socket,server=y,address=5005,suspend=y' --no-daemon
+endif
 
 filter_report_id?=
 
--- a/subprojects/gradle/build-environment/src/main/kotlin/com/avito/utils/gradle/EnvArgs.kt
+++ b/subprojects/gradle/build-environment/src/main/kotlin/com/avito/utils/gradle/EnvArgs.kt
@@ -14,7 +14,7 @@ import java.util.concurrent.TimeUnit
 
 // Used in build.gradle's of avito
 val Project.envArgs by ProjectProperty.lazy<EnvArgs>(scope = ROOT_PROJECT) { project ->
-    if (project.buildEnvironment is BuildEnvironment.CI) EnvArgs.Impl(project) else EnvArgs.Stub
+    EnvArgs.Impl(project)
 }
 
 interface EnvArgs {
@@ -35,7 +35,6 @@ interface EnvArgs {
             override val type = "local-${userName}"
 
             internal enum class Id(val id: Int) {
-                FOR_STUDIO_RUN(-1),
                 FOR_LOCAL_KUBERNETES_RUN(TimeUnit.MILLISECONDS.toSeconds(System.currentTimeMillis()).toInt())
             }
 
@@ -65,6 +64,7 @@ interface EnvArgs {
             project.getBooleanProperty("avito.tests.disableRerunOnTargetBranch", default = false)
 
         init {
+            // todo make local by default?
             build = when (project.getOptionalStringProperty("avito.build", "teamcity")) {
                 "teamcity" -> {
                     val teamcityBuildId = project.getMandatoryIntProperty("teamcityBuildId")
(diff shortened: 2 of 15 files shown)


subprojects/gradle/build-environment/src/main/kotlin/com/avito/utils/gradle/EnvArgs.kt
```diff
@@ -14,7 +14,7 @@ import java.util.concurrent.TimeUnit
 
 // Used in build.gradle's of avito
 val Project.envArgs by ProjectProperty.lazy<EnvArgs>(scope = ROOT_PROJECT) { project ->
-    if (project.buildEnvironment is BuildEnvironment.CI) EnvArgs.Impl(project) else EnvArgs.Stub
+    EnvArgs.Impl(project)
```
sergioboi, line 17: It can't be Stub. We need Env for instrumentation

subprojects/gradle/build-environment/src/main/kotlin/com/avito/utils/gradle/EnvArgs.kt
```diff
@@ -65,6 +64,7 @@ interface EnvArgs {
             project.getBooleanProperty("avito.tests.disableRerunOnTargetBranch", default = false)
 
         init {
+            // todo make local by default?
```
sergioboi, line 67: Consider to make avito.build=local by default

subprojects/gradle/git/src/main/kotlin/com/avito/git/ProjectExtensions.kt
```diff
@@ -5,25 +5,19 @@ package com.avito.git
...
 fun Project.gitState(logger: (String) -> Unit = {}): Provider<GitState> = lazyProperty("GIT_STATE_PROVIDER") { project ->
-    if (project.buildEnvironment is BuildEnvironment.CI) {
-        project.providers.lazy {
-            val strategy = project.getOptionalStringProperty("avito.git.state", default = "env")
-            when (strategy) {
-                "local" -> GitLocalStateImpl.from(project)
-                "env" -> GitStateFromEnvironment.from(project, logger)
-                else -> throw RuntimeException("Unknown git state strategy: $strategy")
-            }
+    project.providers.lazy {
```
sergioboi, line 13: When gitState is notDefined we will fail in execution phase.
It is needed by instrumentation task
One more point is gitState is utility, it musn't make decisions

subprojects/gradle/instrumentation-test-impact-analysis/src/main/kotlin/com/avito/instrumentation/impact/InstrumentationTestImpactAnalysisPlugin.kt
```diff
@@ -26,9 +24,6 @@ class InstrumentationTestImpactAnalysisPlugin : Plugin<Project> {
         val extension = project.extensions
             .create<InstrumentationTestImpactAnalysisExtension>("instrumentationTestImpactAnalysis", project)
 
-        //todo should work in any environment
-        if (project.buildEnvironment !is BuildEnvironment.CI) return
```
sergioboi, line 30: We were failing cause of that in IDE sync phase and it configuration phase without ci=true, cause instrumentation implicitly depends on that plugin

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/configuration/InstrumentationPluginExt.kt
```diff
@@ -59,7 +59,8 @@ internal fun Project.withInstrumentationExtensionData(action: (InstrumentationPl
                         "reportViewerUrl" to extension.reportViewerUrl,
                         "sentryDsn" to extension.sentryDsn,
                         "slackToken" to extension.slackToken,
-                        "fileStorageUrl" to extension.fileStorageUrl
+                        "fileStorageUrl" to extension.fileStorageUrl,
+                        "teamcityBuildId" to envArgs.build.id.toString()
```
sergioboi, line 63: Before that we will fail after test execution cause file for runner isn't generated

subprojects/gradle/module-types/src/test/kotlin/com/avito/android/ModuleTypesPluginTest.kt
```diff
@@ -48,6 +48,7 @@ class ModuleTypesPluginTest {
         val result = gradlew(
             projectDir, "assemble",
             "-Pavito.moduleTypeValidationEnabled=true",
+            "-PgitBranch=xxx", // todo need for impact plugin
```
sergioboi, line 51: All additions of gitBranch property happened because ImpactPlugin depends on gitState.
Before that pr gitState can be notDefined and depens on ci=true.
So if we want ImpactPlugin doesn't force git state we can wrap that logic with ci=true.

Makefile
```diff
@@ -30,6 +30,10 @@ params +=-Pci=$(ci)
 
 params +=$(log_level)
 
+ifeq ($(gradle_debug),true)
+params +=-Dorg.gradle.jvmargs='-Xdebug -Xrunjdwp:transport=dt_socket,server=y,address=5005,suspend=y' --no-daemon
```
eugene-krivobokov, line 34:
```suggestion
params +=-Dorg.gradle.debug=true --no-daemon
```
Is it the same?

eugene-krivobokov:
How much it affects a configuration time of avito project?

eugene-krivobokov (approved):

PR #377, opened 2020-05-09
Mbs 8273 Remove obsolete properties support from build check

eugene-krivobokov:


docs/content/docs/projects/BuildChecks.md
```diff
@@ -48,18 +48,49 @@ A problem occurred configuring root project
 
 After adding missing properties you will get something like this:
 
+{{< tabs "build checks example" >}}
```
eugene-krivobokov, line 51: Configs in Kotlin and Groovy are the same. But it's good to have both variants for the sake of consistent documentation.

sergioboi (approved):

PR #379, opened 2020-05-12
MBS-6978 copy artifacts even if build failed

int02h:


subprojects/gradle/cd/src/main/kotlin/com/avito/ci/steps/VerifyArtifactsStep.kt
```diff
@@ -51,13 +51,12 @@ open class VerifyArtifactsStep(
             entries.set(project.files(artifactsConfig.outputs.values.map { it.path }))
 
             project.gradle.buildFinished { buildResult ->
-                val isCopyScheduled = buildResult.gradle?.isTaskScheduled(this) ?: false
```
int02h, line 54: It seems that when a build is already finished there are no scheduled tasks, so this flag is always `false`.

subprojects/gradle/cd/src/main/kotlin/com/avito/ci/steps/VerifyArtifactsStep.kt
```diff
@@ -51,13 +51,12 @@ open class VerifyArtifactsStep(
...
-                // TODO: consider skipped state (--dry-run)
-                val copyFinished = this.didWork
-                if (isCopyScheduled && !copyFinished) {
-                    project.ciLogger.info("Artifacts copying was scheduled but didn't run. Start it now")
-                    if (!this.didWork) {
```
int02h, line 59: I am sure that double-check is not necessary here

subprojects/gradle/cd/src/main/kotlin/com/avito/ci/steps/VerifyArtifactsStep.kt
```diff
@@ -51,13 +51,12 @@ open class VerifyArtifactsStep(
...
-                        this.doAction()
+                if (buildResult.failure != null && buildResult.action == "Build") {
+                    if (!didWork) {
+                        // Copy artifacts that managed to be generated.
+                        // Do not verify them because it is last resort to save anything.
+                        project.ciLogger.info("Build failed. Trying to copy artifacts that managed to be generated.")
```
sergioboi: debug

subprojects/gradle/cd/src/main/kotlin/com/avito/ci/steps/VerifyArtifactsStep.kt
```diff
@@ -51,13 +51,12 @@ open class VerifyArtifactsStep(
...
             project.gradle.buildFinished { buildResult ->
-                val isCopyScheduled = buildResult.gradle?.isTaskScheduled(this) ?: false
-                // TODO: consider skipped state (--dry-run)
-                val copyFinished = this.didWork
-                if (isCopyScheduled && !copyFinished) {
-                    project.ciLogger.info("Artifacts copying was scheduled but didn't run. Start it now")
-                    if (!this.didWork) {
-                        this.doAction()
+                if (buildResult.failure != null && buildResult.action == "Build") {
+                    if (!didWork) {
```
sergioboi: What if we failed on the copy task?

subprojects/gradle/cd/src/main/kotlin/com/avito/ci/steps/VerifyArtifactsStep.kt
int02h: I've left the same log-level as it was before

subprojects/gradle/cd/src/main/kotlin/com/avito/ci/steps/VerifyArtifactsStep.kt
int02h: Build is already failed but you'll see a message from copy task too

subprojects/gradle/cd/src/main/kotlin/com/avito/ci/steps/VerifyArtifactsStep.kt
```diff
@@ -51,13 +51,12 @@ open class VerifyArtifactsStep(
...
 
             project.gradle.buildFinished { buildResult ->
-                val isCopyScheduled = buildResult.gradle?.isTaskScheduled(this) ?: false
-                // TODO: consider skipped state (--dry-run)
-                val copyFinished = this.didWork
-                if (isCopyScheduled && !copyFinished) {
-                    project.ciLogger.info("Artifacts copying was scheduled but didn't run. Start it now")
-                    if (!this.didWork) {
-                        this.doAction()
+                if (buildResult.failure != null && buildResult.action == "Build") {
```
eugene-krivobokov: It's good to hide this low-level knowledge about "action == "Build".

eugene-krivobokov (approved):

subprojects/gradle/cd/src/main/kotlin/com/avito/ci/steps/VerifyArtifactsStep.kt
int02h: Hide where?

eugene-krivobokov: For example, in an extension utility method.

We missed this one or two times already. `buildFinished` callback can be called multiple times. To distinguish a "real" build we've found only this workaround.

For me, the problem is similar to [a main process](https://github.com/int02h/primaree/blob/master/library/src/main/java/com/dpforge/primaree/ApplicationExtensions.kt) - workaround for lifecycle.

int02h: Done here - 3da553495ac0194d7b1ae822e0d163d6d4af731a

subprojects/gradle/utils/src/main/kotlin/com/avito/utils/Gradles.kt
```diff
@@ -0,0 +1,13 @@
+package com.avito.utils
+
+import org.gradle.api.invocation.Gradle
+
+fun Gradle.onBuildFailed(block: () -> Unit) {
+    buildFinished { buildResult ->
+        if (buildResult.failure != null && buildResult.action == ACTION_BUILD) {
```
sergioboi, line 7: Tell me please, what actions may I see except ACTION_BUILD?

eugene-krivobokov: `Configure`

https://docs.gradle.org/current/javadoc/org/gradle/BuildResult.html#getAction--

PR #385, opened 2020-05-13
MBS-8303 auto assign codeowners for pull requests

dsvoronin:


CODEOWNERS
```diff
@@ -0,0 +1 @@
+* @dsvoronin @eugene-krivobokov @sboishtyan @int02h
```
eugene-krivobokov, line 1: @int02h , is it ok for you?

We can mark only performance modules

eugene-krivobokov (approved):

CODEOWNERS
sergioboi: Did you mean @warabei14 ?

eugene-krivobokov: Yes

PR #391, opened 2020-05-15
MBS-8112 make ci steps plugin kotlin dsl compatible

dsvoronin:
Have some problems with "recommended approach": using `org.gradle.api.Action<T>` interface

Looks like this is my problem: https://docs.gradle.org/current/userguide/custom_gradle_types.html#dsl_support_and_extensibility

> When Gradle creates an instance of a custom type, it decorates the instance to mix-in DSL and extensibility support.
Each decorated instance implements ExtensionAware, and so can have extension objects attached to it.
Note that plugins and the elements of containers created using Project.container() are currently not decorated, due to backwards compatibility issues.

I had to create copy with `Action<T>` of each method with `Closure` and so far this configures OK inside `build.gradle.kts`
```
plugins {
    id("com.avito.android.cd")
}

builds {
    register("someChain") {
        uiTests {
            configurations("regress")
        }
    }
}
```

### DIFF
--- a/docs/content/docs/projects/CiSteps.md
+++ b/docs/content/docs/projects/CiSteps.md
@@ -30,18 +30,35 @@ The plugin can be applied to a root project or any module.
 
 First of all, name your chain:
 
-```groovy
+{{< tabs "chain" >}}
+{{< tab "Kotlin" >}}
+```kotlin
 builds {
-    myChain {
-       ...
+    register("myChain") {
+ 
+       //optional description for generated task
+       taskDescription.set("This chain does something useful")
     }
 }
 ```
+{{< /tab >}}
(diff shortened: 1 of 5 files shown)


eugene-krivobokov:
Would you please add Kotlin examples to the documentation?
[an example](https://avito-tech.github.io/avito-android/docs/test/runner/#filter-tests-by-annotations)

eugene-krivobokov (approved):

dsvoronin:
> Would you please add Kotlin examples to the documentation?
> [an example](https://avito-tech.github.io/avito-android/docs/test/runner/#filter-tests-by-annotations)

done, thanks for pointing out

PR #392, opened 2020-05-17
MBS-8477 Make build properties plugin more reusable

eugene-krivobokov:


subprojects/gradle/build-properties/src/main/kotlin/com/avito/android/info/BuildPropertiesExtension.kt
```diff
@@ -0,0 +1,11 @@
+package com.avito.android.info
+
+open class BuildPropertiesExtension {
+
+    internal val properties = mutableMapOf<String, String>()
+
+    fun buildProperty(name: String, value: String) {
```
eugene-krivobokov, line 7: This will be an input for a `WriteProperties` task. It supports deferred values (see DeferredUtil).
We can do it later without breaking backward compatibility for clients.

dsvoronin (approved):

subprojects/gradle/build-properties/src/main/kotlin/com/avito/android/info/BuildPropertiesPlugin.kt
```diff
@@ -11,20 +9,33 @@ import org.gradle.kotlin.dsl.register
...
     override fun apply(project: Project) {
-        logger = project.ciLogger
+        registerPropertiesTask(project)
+        registerLegacyPropertiesTask(project)
+    }
+
+    private fun registerPropertiesTask(project: Project) {
+        val extension = project.extensions.create<BuildPropertiesExtension>("buildProperties")
+
+        val task = project.tasks.register<WriteProperties>("generateBuildProperties") {
```
sergioboi, line 20: How will we depend on it? Should we declare contracts `val` with task names?

eugene-krivobokov: `project.addPreBuildTasks(task)` - that's all we need

PR #393, opened 2020-05-18
MBS-8348 Ignore unknown CD config deployments

eugene-krivobokov:
We have two components: CD service ("Nupokati") and Android infrastructure. CD service runs the release build.
These components deployed independently. CD service works with multiple versions of Android infrastructure at the same time.

We need to support a new deployment type: #365

To achieve backward compatibility I've chosen to support it on the config level. 
Config has a schema version, but it will be inconvenient in CD service to know about client version.


subprojects/gradle/upload-cd-build-result/src/main/kotlin/com/avito/cd/ProjectExtentions.kt
```diff
@@ -23,7 +24,7 @@ private class CdBuildConfigFactory : (Project) -> Provider<CdBuildConfig> {
                 Providers.of(configFilePath).map { path ->
                     val configFile = project.rootProject.file(path)
                     val config = deserializeToCdBuildConfig(configFile)
-                    validate(config)
+                    CdBuildConfigValidator(config, project.ciLogger).validate()
```
sergioboi, line 27: We should configure Validator to fail a build if we have unexpected deployment but we don't wait for it

sergioboi (changes requested):

PR #397, opened 2020-05-19
MBS-8094 refactor performance plugin extension to use properties

dsvoronin:


eugene-krivobokov (approved):

subprojects/gradle/performance/src/main/kotlin/com/avito/performance/PerformanceExtension.kt
```diff
@@ -1,14 +1,28 @@
...
-    /* Результаты перформанс-тестов, резултаты сравнения с предыдущим прогоном на девелопе */
-    lateinit var output: String
+open class PerformanceExtension @Inject constructor(objects: ObjectFactory) {
 
-    /* Название файла с результатами теста */
-    lateinit var performanceTestResultName: String
+    /**
+     * Name of a directory for performance test results
+     */
+    val output = objects.property<String>()
```
sergioboi, line 12: Why don't we use a Directory, FileCollenction type?

dsvoronin: it would be an API change; i tried to avoid it for now

PR #399, opened 2020-05-20
MBS-7579 implement check for overlapping views

int02h:


subprojects/android-test/test-app/src/androidTest/kotlin/com/avito/android/ui/test/OverlapTest.kt
```diff
@@ -0,0 +1,49 @@
...
+    @get:Rule
+    val rule = screenRule<OverlapActivity>()
+
+    @Test
+    fun isOverlapped_button_under_snackbar() {
+        rule.launchActivity(null)
+
+        Screen.overlapScreen.snackButton.click()
+
+        Screen.overlapScreen.snackButton.checks.isOverlapped()
```
sergioboi, line 20: I think we need only IsDisplayedWithoutOverlapped
This scenario we should check by occurred exception

subprojects/android-test/ui-testing-core/src/main/kotlin/com/avito/android/test/checks/Checks.kt
```diff
@@ -36,13 +36,23 @@ interface Checks : LabelChecks {
...
      * user.
      *
      * Note: isNotDisplayed inverts the logic by isDisplayed().
      */
     fun isNotDisplayed()
 
+    /**
+     * Accepts a [View] that overlapped with another view by at least one pixel
+     */
+    fun isOverlapped()
```
sergioboi, line 49: The same as in the test

subprojects/android-test/ui-testing-core/src/main/kotlin/com/avito/android/test/matcher/OverlapMatcher.kt
```diff
@@ -0,0 +1,52 @@
...
+
+    override fun describeTo(description: Description) {
+        description.appendText("overlapped by another views on the same screen")
+    }
+
+    override fun matchesSafely(view: View): Boolean {
+        val viewRect = Rect()
+        view.getGlobalVisibleRect(viewRect)
+
+        return view.parentViewGroup?.iterateViewsBelow(view) { viewBelow ->
```
sergioboi: I consider we need to use `rootView`

subprojects/android-test/ui-testing-core/src/main/kotlin/com/avito/android/test/matcher/OverlapMatcher.kt
```diff
@@ -0,0 +1,52 @@
...
+            val child = getChildAt(i)
+            if (block(child)) {
+                return true
+            }
+        }
+
+        return parentViewGroup?.iterateViewsBelow(this, block) ?: false
+    }
+
+    private fun View.isIntersectWith(otherRect: Rect): Boolean {
```
eugene-krivobokov:
```suggestion
    private fun View.isIntersectedWith(otherRect: Rect): Boolean {
```

subprojects/android-test/ui-testing-core/src/main/kotlin/com/avito/android/test/matcher/OverlapMatcher.kt
int02h: Why do we need to use `rootView`? This method should iterate views from current parent up to root view. Check the docs above the class.

subprojects/android-test/test-app/src/androidTest/kotlin/com/avito/android/ui/test/OverlapTest.kt
int02h: Could you please explain why? In my opinion, it's a separate check like `isDisplay` or `isVisible`. In task MBS-8512 (internal) they will be combined into one method.

subprojects/android-test/ui-testing-core/src/main/kotlin/com/avito/android/test/matcher/OverlapMatcher.kt
int02h: Fixed

subprojects/android-test/ui-testing-core/src/main/kotlin/com/avito/android/test/matcher/OverlapMatcher.kt
```diff
@@ -0,0 +1,52 @@
...
+    private fun ViewGroup.iterateViewsBelow(view: View, block: (View) -> Boolean): Boolean {
+        val index = indexOfChild(view)
+        for (i in (index + 1) until childCount) {
+            val child = getChildAt(i)
+            if (block(child)) {
+                return true
+            }
+        }
+
+        return parentViewGroup?.iterateViewsBelow(this, block) ?: false
```
sergioboi: I saw an another strategy without recursion. I consider writing code like in that assertion.androidx.test.espresso.assertion.LayoutAssertions#noOverlaps(org.hamcrest.Matcher<android.view.View>)
- You must check on VISIBILITY
I've made a utility function to get all above visible views `com.avito.android.test.util.ViewKt#getAboveVisibleViews`
PR: where I tried to find above clickable view https://github.com/avito-tech/avito-android/pull/146
matcher with simpler hierarchy traverse code https://github.com/avito-tech/avito-android/blob/a240ff41d9498045769abd3f8f5a9229620ff1c1/subprojects/android-test/ui-testing-core/src/main/kotlin/com/avito/android/test/matcher/ViewCoordinatesNotOverlappedMatcher.kt

int02h: - `com.avito.android.test.util.ViewKt#getAboveVisibleViews` works as expected but return not that views that I need to check for overlapping
- Visibility fixed in 746d623a. Thanks for pointing an issue.

subprojects/android-test/test-app/src/androidTest/kotlin/com/avito/android/ui/test/OverlapTest.kt
sergioboi: Because users need to check not overlapped scenario. Why do they check `isOverlapped`?

subprojects/android-test/ui-testing-core/src/main/kotlin/com/avito/android/test/util/View.kt
```diff
@@ -4,13 +4,18 @@ import android.graphics.Rect
...
 import android.view.ViewGroup
 import android.view.ViewParent
-import androidx.test.espresso.util.TreeIterables
 
-internal fun View.getAboveVisibleViews(): List<View> {
-    return TreeIterables.breadthFirstViewTraversal(this.rootView)
-        .dropWhile { it != this }
-        .drop(1)
-        .filter { it.visibility == View.VISIBLE }
+internal fun View.forEachVisibleViewsWithHigherZOrder(block: (View) -> Boolean): Boolean {
```
sergioboi: Could you add comment with link to TreeIterables.breadthFirstViewTraversal(this.rootView)
And describe that it's a better version because of something

subprojects/android-test/ui-testing-core/src/main/kotlin/com/avito/android/test/matcher/OverlapMatcher.kt
```diff
@@ -0,0 +1,30 @@
...
+ */
+class OverlapMatcher : TypeSafeMatcher<View>() {
+
+    override fun describeTo(description: Description) {
+        description.appendText("overlapped by another views on the same screen")
+    }
+
+    override fun matchesSafely(view: View): Boolean {
+        val viewRect = view.getRect()
+        return view.forEachVisibleViewsWithHigherZOrder { higherZOrderView ->
```
sergioboi: I consider better get the list and iterate in the matcher. It will give more control for those who want to reuse your function

int02h: Fixed

subprojects/android-test/ui-testing-core/src/main/kotlin/com/avito/android/test/util/View.kt
int02h: Done

eugene-krivobokov (approved):

subprojects/android-test/test-app/src/androidTest/kotlin/com/avito/android/ui/test/OverlapTest.kt
int02h: I think that API should have two related methods: `is*` and `isNot*`

PR #401, opened 2020-05-20
MBS-8423 Update keeper

eugene-krivobokov:
https://github.com/slackhq/keeper/blob/master/CHANGELOG.md

- Gradle 6.4 support
- Uses Zipflinger for packaging, which should give a nice speed boost in creating intermediate jars

### DIFF
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -182,7 +182,7 @@ pluginManagement {
                     useModule("com.avito.android:${pluginId.removePrefix("com.avito.android.")}:$infraVersion")
 
                 pluginId == "com.slack.keeper" ->
-                    useModule("com.slack.keeper:keeper:0.2.0")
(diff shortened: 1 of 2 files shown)


settings.gradle.kts
```diff
@@ -182,7 +182,7 @@ pluginManagement {
                     useModule("com.avito.android:${pluginId.removePrefix("com.avito.android.")}:$infraVersion")
 
                 pluginId == "com.slack.keeper" ->
-                    useModule("com.slack.keeper:keeper:0.2.0")
+                    useModule("com.slack.keeper:keeper:0.4.3")
```
sergioboi, line 185: Should run fullCheck against that branch

dsvoronin: it only affects github project

dsvoronin (approved):

PR #405, opened 2020-05-20
MBS-8132 Remove direct ReportsApi usages

sergioboi:
add markAsSuccessful to Report.kt
add getTests() to Report.kt

### DIFF
--- a/subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/report/Report.kt
+++ b/subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/report/Report.kt
@@ -3,6 +3,7 @@ package com.avito.instrumentation.report
 import com.avito.report.ReportsApi
 import com.avito.report.model.AndroidTest
 import com.avito.report.model.CreateResult
+import com.avito.report.model.CrossDeviceSuite
 import com.avito.report.model.GetReportResult
 import com.avito.report.model.ReportCoordinates
 import com.avito.report.model.SimpleRunTest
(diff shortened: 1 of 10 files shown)


subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/report/Report.kt
```diff
@@ -27,6 +28,9 @@ interface Report {
 
     fun finish(isFullTestSuite: Boolean, reportViewerUrl: HttpUrl)
 
+    fun getTests(): Try<List<SimpleRunTest>>
```
sergioboi, line 32: It's the first small step.
1) Add all needed functions in Report
2) Replace all direct usages of ReportApi
3) Hide Report creation behind factories
4) Refactor Report if needed

dsvoronin: Looks like SimpleRunTest also from `:report-viewer`? do you have plans to abstract it?
If no maybe it's better to move all the stuff from `:instrumentation-test` into `:report-viewer`? 
It's strange that we're increasing complexity of a gradle plugin module

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/report/Report.kt
```diff
@@ -150,6 +158,18 @@ interface Report {
             }
         }
 
+        override fun getTests(): Try<List<SimpleRunTest>> {
+            return reportsApi.getTestsForRunId(reportCoordinates)
+        }
+
+        override fun markAsSuccessful(testRunId: String, author: String, comment: String): Try<Unit> {
+            return reportsApi.markAsSuccessful(testRunId, author, comment)
```
dsvoronin, line 166: why do you think it's a good idea to write these proxies? 
They have no real value, instead it's for possible Allure alternative implementation; i didn't saw such intent in this PR, so it's unclear.
If it's for allure, we definitely need to move this code away from this module

dsvoronin (approved):

PR #406, opened 2020-05-21
MBS-8094 removed debug log

dsvoronin:


sergioboi (approved):

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/rerun/RunOnTargetCommitCondition.kt
```diff
@@ -18,8 +18,6 @@ object RunOnTargetCommitCondition {
     ): Result {
         val useArtifactsFromTargetBranch = isArtifactsFromTargetBranchNeeded(instrumentationConfiguration)
 
-        println("DEBUG: ${instrumentationConfiguration.name} useArtifactsFromTargetBranch=$useArtifactsFromTargetBranch")
```
eugene-krivobokov, line 21: It'd good to have a linter for println

PR #407, opened 2020-05-21
MBS-8303 remove mockwebserver spam in tests

dsvoronin:


subprojects/common/test-okhttp/src/main/kotlin/com/avito/test/http/MockDispatcher.kt
```diff
@@ -16,15 +25,15 @@ class MockDispatcher(
 
         capturers.find { it.requestMatcher.invoke(request) }?.run {
             capture(request)
-            println("request captured: $request")
+            logger.info("request captured: $request")
```
sergioboi, line 28: debug?

dsvoronin: there is no debug in java.util.Logger that is used here

eugene-krivobokov: what about `com.avito.logger.Logger`? Is it for such kotlin libraries?

eugene-krivobokov (approved):

sergioboi (approved):

subprojects/common/test-okhttp/src/main/kotlin/com/avito/test/http/MockDispatcher.kt
dsvoronin: Mockwebserver uses java.util under the hood, and it seems overcomplicated to redirect it to our logger

PR #408, opened 2020-05-21
MBS-6514 Remove android-maven-publish

sergioboi:
publish via agp https://developer.android.com/studio/build/maven-publish-plugin

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -138,36 +138,25 @@ subprojects {
             }
         }
     }
-
-    plugins.withType<LibraryPlugin> {
-        tasks.register<Jar>(sourcesTaskName).configure {
-            archiveClassifier.set("sources")
-            from(this@withType.extension.sourceSets["main"].java.srcDirs)
+    plugins.withId("com.android.library") {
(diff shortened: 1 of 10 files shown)


dsvoronin:
Wow! Thanks for that.

How did you test it?

dsvoronin (approved):

build.gradle.kts
```diff
@@ -138,11 +138,19 @@ subprojects {
             }
         }
     }
-
-    plugins.withType<LibraryPlugin> {
-        tasks.register<Jar>(sourcesTaskName).configure {
-            archiveClassifier.set("sources")
-            from(this@withType.extension.sourceSets["main"].java.srcDirs)
+    plugins.withId("com.android.library") {
```
dsvoronin, line 141: Is type -> string weakening necessary here?

sergioboi: No, we can replace by Type

sergioboi:
> Wow! Thanks for that.
> 
> How did you test it?

I test it locally. I haven't tested a bintray scenario only. Because I didn't get how.

PR #409, opened 2020-05-22
Mbs 6859 use emulator 28

int02h:


subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/EmulatorSet.kt
```diff
@@ -5,12 +5,12 @@ import com.avito.instrumentation.reservation.request.Device
 object EmulatorSet {
     val fast = setOf(
         Device.Emulator.Emulator22,
-        Device.Emulator.Emulator27
+        Device.Emulator.Emulator28
```
sergioboi: Help me, why did we decide to replace `27 API`?

int02h: Because it's a set for the fast check. It should be rapid and limited to 40 minutes. So we can run our tests only on 2 emulators. It seems consequent to run tests on the minimum supported API and the target supported API.

ci/docker/android-emulator/Dockerfile
```diff
@@ -55,7 +56,7 @@ RUN mkdir $HOME/.android && \
         "platforms;android-28" \
         "emulator" \
         "tools" \
-        "system-images;android-${SDK_VERSION};google_apis;x86" && \
+        "system-images;android-${SDK_VERSION};google_apis;${EMULATOR_ARCH}" && \
```
eugene-krivobokov, line 59: Is it the same for all emulator versions?

eugene-krivobokov: I see, the logic in emulator.py

ci/docker/docker-in-docker-image/publish-docker-image/publish_docker_image/emulator.py
```diff
@@ -23,7 +23,10 @@ def publish_emulator(directory, registry, image, versions):
         not_prepared_image = build_image(
             image=full_image,
             directory=directory,
-            build_args={'SDK_VERSION': version}
+            build_args={
+                'SDK_VERSION': version,
+                'EMULATOR_ARCH': "x86" if int(version) < 28 else "x86_64"
```
eugene-krivobokov, line 28: Can we use x64 images for pre 28 API also?

ci/docker/android-emulator/prepare_snapshot.sh
```diff
@@ -56,6 +56,10 @@ adb shell "settings put secure show_ime_with_hard_keyboard 1"
 adb shell "settings put system screen_off_timeout 2147483647"
 adb shell "settings put secure long_press_timeout 1500"
 
+# https://developer.android.com/distribute/best-practices/develop/restrictions-non-sdk-interfaces#how_can_i_enable_access_to_non-sdk_interfaces
+adb shell "settings put global hidden_api_policy_pre_p_apps 1"
```
int02h, line 60: TODO: add task for Avito to disallow hiddem api in app

docs/content/docs/test/Runner.md
```diff
@@ -672,16 +672,16 @@ instrumentation {
...
+                    deviceName = "API28"
                     scheduling {
                         quota {
                             retryCount = 1
                             minimumSuccessCount = 1
                         }
 
                         testsCountBasedReservation {
-                            device = new LocalEmulator("27", 27, "Android_SDK_built_for_x86")
+                            device = new LocalEmulator("28", 28, "AOSP_on_IA_Emulator")
```
eugene-krivobokov: I've checked the image `37ac40d0fa` locally and see another state: `Model = Android_SDK_built_for_x86_64`

subprojects/android-test/test-app/build.gradle.kts
```diff
@@ -180,7 +180,7 @@ extensions.getByType<GradleInstrumentationPluginConfiguration>().apply {
                 }
 
                 reservation = TestsBasedDevicesReservationConfiguration().apply {
-                    device = LocalEmulator.device(27)
+                    device = LocalEmulator.device(28)
```
eugene-krivobokov: Please, support also model parameter here.

docs/content/docs/test/Runner.md
```diff
@@ -672,16 +672,16 @@ instrumentation {
...
+                    deviceName = "API28"
                     scheduling {
                         quota {
                             retryCount = 1
                             minimumSuccessCount = 1
                         }
 
                         testsCountBasedReservation {
-                            device = new LocalEmulator("27", 27, "Android_SDK_built_for_x86")
+                            device = new LocalEmulator("28", 28, "Android_SDK_built_for_x86_64")
```
eugene-krivobokov, line 684: I think we should reconsider how DeviceWorker selects available devices.
As a user, I want to request only an API level. I have different local devices with different device names.

sergioboi: I think we must add api like a
emulator27 {
}
emulator22 {
}
and
customEmulator {
imagePath =
version = 
}

int02h: I agree that we should reconsider DeviceWorker's mechanism but I believe it's out of the scope of this task.

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/EmulatorSet.kt
int02h: Not actual. I've removed `Emulator28` from these sets.

ci/docker/docker-in-docker-image/publish-docker-image/publish_docker_image/emulator.py
int02h: I don't understand what profit will we get?

eugene-krivobokov: The main hypothesis is performance: https://stackoverflow.com/a/47451413/1282707 
But we need benchmarks for emulators.

eugene-krivobokov (approved):

PR #410, opened 2020-05-22
MBS-8094 track performance compare task outcome in statsd

dsvoronin:


subprojects/gradle/performance/src/main/kotlin/com/avito/performance/PerformanceCompareAction.kt
```diff
@@ -78,6 +79,8 @@ open class PerformanceCompareAction(
     constructor(params: Params) : this(params, params.logger)
 
     override fun run() {
+        val metricPrefix = "ci.performance.outcome.${params.pullRequestId ?: "null"}"
```
eugene-krivobokov:
```suggestion
        val metricPrefix = "ci.performance.${params.pullRequestId ?: "null"}.compare.outcome"
```

It might be more convenient to aggregate more hierarchical metrics:
builds > build (id) > build metric > sub metric

Think of it as about nested metrics inside a common namespace.

eugene-krivobokov (approved):

PR #414, opened 2020-05-22
MBS-8303 remove stdout stacktrace spam

dsvoronin:


subprojects/gradle/utils/src/main/kotlin/com/avito/utils/logging/CILoggingDestination.kt
```diff
@@ -25,7 +25,6 @@ object StdoutDestination : CILoggingDestination() {
         println(formattedMessage.message)
         formattedMessage.cause?.let { throwable ->
             println(throwable.message)
-            throwable.printStackTrace()
```
sergioboi, line 28: Do we lose a stacktrace now?

dsvoronin: it's available in sentry
you can find it via error message

sergioboi: It's hard to find all stacktraces related to build
Could we generate a link to the Sentry with applied filters to see all reported error by current build?

dsvoronin: great idea! will do in separate PR

int02h (approved):

PR #415, opened 2020-05-22
MBS-8303 dismiss dialog only for >=28 api

dsvoronin:
to prevent extra No field mHiddenApiWarningShown

### DIFF
--- a/subprojects/android-test/test-instrumentation-runner/src/main/kotlin/com/avito/android/runner/SystemDialogsManager.kt
+++ b/subprojects/android-test/test-instrumentation-runner/src/main/kotlin/com/avito/android/runner/SystemDialogsManager.kt
@@ -1,6 +1,7 @@
 package com.avito.android.runner
 
 import android.annotation.SuppressLint
+import android.os.Build
 import android.util.Log
 import androidx.test.platform.app.InstrumentationRegistry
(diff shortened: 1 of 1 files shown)


eugene-krivobokov (approved):

subprojects/android-test/test-instrumentation-runner/src/main/kotlin/com/avito/android/runner/SystemDialogsManager.kt
```diff
@@ -27,11 +28,13 @@ class SystemDialogsManager(private val errorsReporter: ErrorsReporter) {
...
         }
 
-        try {
-            dismissHiddenApiDialog()
-        } catch (t: Throwable) {
-            Log.v(TAG, "Failed to close hidden api dialog: ${t.message}")
-            errorsReporter.reportError(t)
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
+            try {
+                dismissHiddenApiDialog()
```
sergioboi, line 33: Would it increase test flakiness because of dialogs?

eugene-krivobokov: This dialog is about restricted API and must be only on 28+ API

PR #418, opened 2020-05-25
MBS-8526 Wait device properly before executing tests

eugene-krivobokov:


subprojects/gradle/runner/service/src/main/kotlin/com/avito/runner/service/worker/DeviceWorker.kt
```diff
@@ -30,18 +31,26 @@ class DeviceWorker(
         Executors.newSingleThreadExecutor().asCoroutineDispatcher()
     )
 
-    private var state: State = device.state()
```
eugene-krivobokov, line 33: The problem was here. We use ADB implicitly to create a device state. If anything happens, the worker will fail.

```
java.lang.RuntimeException: Timeout to execute command: [shell, getprop, ro.build.version.sdk] on device 10.21.198.180:5555
...
 at com.avito.runner.service.worker.device.adb.AdbDevice.getApi(AdbDevice.kt)
 at com.avito.runner.service.worker.DeviceWorker.state(DeviceWorker.kt:172)
 at com.avito.runner.service.worker.DeviceWorker.<init>(DeviceWorker.kt:33)
```

In the `run` function it will wait with proper retries.

subprojects/gradle/runner/service/src/main/kotlin/com/avito/runner/service/worker/device/adb/AdbDevice.kt
```diff
@@ -38,9 +38,19 @@ data class AdbDevice(
...
+            block = { attempt ->
+                log("Attempt $attempt: getting ro.build.version.sdk")
+                loadProperty(
+                    key = "ro.build.version.sdk",
+                    cast = { it.toInt() }
+                )
+            },
+            attemptFailedHandler = { attempt, throwable ->
+                log("Attempt $attempt: reading ro.build.version.sdk failed. Reason: ${throwable.message}")
+            }
```
eugene-krivobokov, line 53: This is not a solution to the problem. Anyway, it's better to retry all unstable operations.

subprojects/gradle/runner/service/src/test/kotlin/com/avito/runner/service/DeviceWorkerTest.kt
```diff
@@ -76,11 +78,12 @@ class DeviceWorkerTest {
                         )
                     )
                 },
-                gettingDeviceStatusResults = intentions.map {
-                    MockActionResult.Success<Device.DeviceStatus>(
-                        Device.DeviceStatus.Alive
+                gettingDeviceStatusResults = listWithDefault(
+                    1 + intentions.size,
```
eugene-krivobokov, line 82: Unfortunately, tests know too much about inner logic.
This is a bad example of custom stubs.
I haven't changed it intentionally to reduce the scope of changes.

We will discuss it additionally in another PR.

subprojects/gradle/runner/service/src/test/kotlin/com/avito/runner/service/DeviceWorkerTest.kt
```diff
@@ -124,7 +127,46 @@ class DeviceWorkerTest {
         }
 
     @Test
-    fun `worker returns only one worker failed message when device is freeze`() =
+    fun `fail with no events - device is freeze before processing intentions`() =
```
eugene-krivobokov, line 130: New test for changed logic.

subprojects/gradle/runner/service/src/main/kotlin/com/avito/runner/service/worker/DeviceWorker.kt
```diff
@@ -30,18 +31,26 @@ class DeviceWorker(
...
+        } catch (t: Exception) {
+            device.notifyError("Error while checking device initial state", t)
+            // No intention is lost. DeviceWorkerMessage.WorkerFailed event is unnecessary.
+            // Can't use this device any more. ReservationClient will get a new one.
+            return@launch
+        }
+
         for (intention in intentionsRouter.observeIntentions(state)) {
             try {
                 checkDeviceAlive()
```
eugene-krivobokov, line 48: This is ok to check again. The worker will receive an intention later after the initial check.

subprojects/gradle/runner/service/src/main/kotlin/com/avito/runner/service/worker/DeviceWorker.kt
```diff
@@ -30,18 +31,26 @@ class DeviceWorker(
         Executors.newSingleThreadExecutor().asCoroutineDispatcher()
     )
 
-    private var state: State = device.state()
-
     fun run() = scope.launch {
 
+        var state: State = try {
```
sergioboi, line 36: Could we replace `var` with two `val`
`previousState` and `preparedState`

subprojects/gradle/runner/service/src/main/kotlin/com/avito/runner/service/worker/DeviceWorker.kt
```diff
@@ -30,18 +31,26 @@ class DeviceWorker(
...
+            checkDeviceAlive()
+            device.state()
+        } catch (t: Exception) {
+            device.notifyError("Error while checking device initial state", t)
+            // No intention is lost. DeviceWorkerMessage.WorkerFailed event is unnecessary.
+            // Can't use this device any more. ReservationClient will get a new one.
+            return@launch
+        }
```
sergioboi, line 44: As I see If we return there we will lose the deviceWorker

subprojects/gradle/runner/service/src/main/kotlin/com/avito/runner/service/worker/device/adb/AdbDevice.kt
```diff
@@ -38,9 +38,19 @@ data class AdbDevice(
 ) : Device {
 
     override val api: Int by lazy {
```
sergioboi, line 40: Could we take API from the configuration? And try to load API Property only to check that our API and API from the emulator are the same

subprojects/gradle/runner/service/src/main/kotlin/com/avito/runner/service/worker/DeviceWorker.kt
eugene-krivobokov: Yes, this is ~correct~ expected behavior. We can't retry infinitely on the same device.

Killing and recovering device is a responsibility of ReservationClient. We will try to do it by k8s (readiness/liveness probe) in another issue.

subprojects/gradle/runner/service/src/main/kotlin/com/avito/runner/service/worker/DeviceWorker.kt
eugene-krivobokov: Can't get rid of mutable state. It's mutated in the loop on each cycle.

subprojects/gradle/runner/service/src/main/kotlin/com/avito/runner/service/worker/DeviceWorker.kt
```diff
@@ -75,13 +84,13 @@ class DeviceWorker(
...
-    /**
-     * Готовим состояние девайса к прогну теста
-     */
-    private suspend fun prepareState(intentionState: State): Try<State> {
-        device.log("Check worker state. Current: ${state.digest}, desired: ${intentionState.digest}")
-        if (intentionState.digest == state.digest) {
-            clearStatePackages().get()
+    private suspend fun prepareDeviceState(
+        currentState: State,
+        intentionState: State
```
int02h, line 89: Maybe `targetState` will be a little bit more proper name

subprojects/gradle/runner/service/src/main/kotlin/com/avito/runner/service/worker/device/adb/AdbDevice.kt
```diff
@@ -410,11 +425,23 @@ data class AdbDevice(
...
+    }
+
+    override fun toString(): String {
+        return "Device $TAG"
+    }
+
+    private val TAG: String by lazy {
+        try {
+            "[$id $api]"
+        } catch (ignore: Exception) {
```
int02h: It looks crazy that we catch all exceptions only for retrieving tag value. When does exception can occur here?

int02h (approved):

subprojects/gradle/runner/service/src/main/kotlin/com/avito/runner/service/worker/device/adb/AdbDevice.kt
eugene-krivobokov: This is a hypothetical case when a device is frozen.
We don't want to lose an error message in this case.

We know how to avoid this problem. ReservationClient can send validated Device with already checked API level. Hope to do it in MBS-8531

subprojects/gradle/runner/service/src/main/kotlin/com/avito/runner/service/worker/device/adb/AdbDevice.kt
eugene-krivobokov: Added to MBS-8531

subprojects/gradle/runner/service/src/main/kotlin/com/avito/runner/service/worker/device/adb/AdbDevice.kt
eugene-krivobokov: Thanks for pointing attention. It was a recursion here.
Removed this logic to avoid such problems.

PR #419, opened 2020-05-25
MBS-8007 fix lazy provider

dsvoronin:
Added some more tech debt in this PR;
Tests shows what a mess is ProjectProperty.lazy returning Provider<T>, and why we need to get rid of it soon

### DIFF
--- a/subprojects/gradle/kotlin-dsl-support/src/test/kotlin/com/avito/kotlin/dsl/PropertyExtensionsKtTest.kt
+++ b/subprojects/gradle/kotlin-dsl-support/src/test/kotlin/com/avito/kotlin/dsl/PropertyExtensionsKtTest.kt
@@ -76,24 +76,38 @@ internal class PropertyExtensionsKtTest {
     }
 
     @Test
-    fun `single value - lazy project providers property`() {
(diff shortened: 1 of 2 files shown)


subprojects/gradle/kotlin-dsl-support/src/test/kotlin/com/avito/kotlin/dsl/PropertyExtensionsKtTest.kt
```diff
@@ -88,11 +146,41 @@ internal class PropertyExtensionsKtTest {
...
+
     private val Project.singleProperty: String
         by ProjectProperty.lazy(scope = PropertyScope.ROOT_PROJECT) { UUID.randomUUID().toString() }
+
     private val Project.intProperty: Int by ProjectProperty.lazy { project ->
         project.property("intProperty").toString().toInt()
     }
+
+    private val Project.transformedProviderProperty: Provider<String> by ProjectProperty.lazy { project ->
+        Providers.of(UUID.randomUUID().toString() + project.providerProperty.get())
```
eugene-krivobokov, line 161: Why do we use `providerProperty` here?

eugene-krivobokov (approved):

PR #428, opened 2020-06-02
MBS-8742 Add test-app-second module 

sergioboi:
It shall help to check writing to report-viewer from multiple modules


eugene-krivobokov (approved):

eugene-krivobokov:
> It shall help to check writing to report-viewer from multiple modules

Also, this is a good example of using a test runner in multiple applications in a monorepo.

PR #434, opened 2020-06-03
MBS-8574-instrumentation-verdict-file

sergioboi:


subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/suite/TestParser.kt
```diff
@@ -52,53 +70,39 @@ internal fun parseTest(testInApk: TestInApk, deviceName: DeviceName): TestStatic
     kind = determineKind(testInApk.annotations),
 
     flakiness = testInApk.annotations
-        .find { it.name == FLAKY_NAME }
-        ?.getStringValue(FLAKY_REASON_KEY)
-        ?.let { reason -> Flakiness.Flaky(reason)}
+        .find { it.name == Flaky::class.java.canonicalName }
+        ?.let { Flakiness.Flaky(it.getStringValue(FLAKY_REASON_KEY) ?: NO_REASON) } // todo we can't parse annotations default from dex
```
sergioboi, line 74: So the bug was because we didn't parse annotation default value and this logic skipped annotation without value
I will create a task for support annotation's default value parsing

subprojects/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,5 @@
+distributionBase=GRADLE_USER_HOME
```
sergioboi, line 1: Without wrapper IDEA runs tests with strange gradle version
I looked for the decision and found only that solution. Keeper made the same

subprojects/settings.gradle.kts
```diff
@@ -78,11 +78,11 @@ include(":common:statsd")
 include(":common:statsd-test-fixtures")
 include(":common:logger")
 include(":common:teamcity-common")
+include(":common:test-annotations")
```
sergioboi, line 81: It's a breaking change but I want to support it on the client-side

dsvoronin (approved):

PR #436, opened 2020-06-05
MBS-8001 extract emulator configs, to change it in clients

dsvoronin:


subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/reservation/request/Device.kt
```diff
@@ -29,6 +29,20 @@ sealed class Device : Serializable {
         }
     }
 
+    class CloudEmulator(
```
sergioboi, line 32: Let's make it data class

sergioboi (approved):

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/reservation/request/Device.kt
dsvoronin: do you see a specific reason for it?

sergioboi: No. This class looks like data class

PR #447, opened 2020-06-16
MBS-8599 update recyclerview lib; cosmetics

dsvoronin:


subprojects/android-test/ui-testing-core/src/main/kotlin/com/avito/android/test/element/field/actions/TypeText.kt
```diff
@@ -52,8 +52,23 @@ internal class TypeText(private val stringToBeTyped: String) : ViewAction {
             return
         }
 
-        tapForFocus(uiController = uiController, editText = view)
-        writeText(uiController = uiController, editText = view)
+        try {
+            tapForFocus(uiController = uiController, editText = view)
```
eugene-krivobokov: Is it ok if we already have a focus before action?

eugene-krivobokov:
I can't get the main point. What's the difference? recyclerview version?

subprojects/android-test/ui-testing-core/src/main/kotlin/com/avito/android/test/element/field/actions/TypeText.kt
sergioboi: So, the problem was because of the focus error?

dsvoronin:
So i reverted all changes except recycler view update in this PR; fix will be in separate one

eugene-krivobokov (approved):

PR #449, opened 2020-06-17
MBS-8599 fix scroll problem (closing soft keyboard)

dsvoronin:


subprojects/android-test/ui-testing-core/src/main/kotlin/com/avito/android/test/element/field/TextFieldAction.kt
```diff
@@ -41,9 +41,9 @@ internal class TextFieldAction(
                 TypeText(
                     stringToBeTyped = text
                 )
-            )
+            ),
+            closeSoftKeyboard()
```
eugene-krivobokov, line 45: Can we reproduce the problem in a test?

eugene-krivobokov (approved):

subprojects/android-test/ui-testing-core/src/main/kotlin/com/avito/android/test/element/field/TextFieldAction.kt
dsvoronin: will try

subprojects/android-test/ui-testing-core/src/main/kotlin/com/avito/android/test/element/field/TextFieldAction.kt
```diff
@@ -57,9 +57,9 @@ internal class TextFieldAction(
                     stringToBeTyped = text
                 )
             ),
-            pressImeActionButton()
+            pressImeActionButton(),
```
sergioboi, line 60: So what is changed? I didn't get the idea of that fix
Is the difference between closeSoftKeyboard or we must call pressImeActionButton differently?

dsvoronin: no, closeSoftKeyboard was called without waitFor

PR #450, opened 2020-06-17
MBS-8739 Add SnackbarProxyRule.kt

sergioboi:
it helps to check Snackbar


settings.gradle.kts
```diff
@@ -2,19 +2,24 @@
 
 rootProject.name = "avito-android-infra"
 
+val useCompositeBuild: String by settings
+
 include(":samples:test-app")
 include(":samples:test-app-kaspresso")
-include(":samples:test-app-second")
+if (useCompositeBuild.toBoolean()) {
+    include(":samples:test-app-without-backward-compatibility")
```
sergioboi, line 10: It's an important change for working flow. Without that, we have no ability to create module and add it to sample app because it will break configuration on the previous binary version where new module is absent.
I renamed module and consider to use it with breaking changes and new modules if need

subprojects/android-lib/snackbar-proxy/src/main/kotlin/com/avito/android/snackbar/proxy/SnackbarProxy.kt
```diff
@@ -0,0 +1,20 @@
+package com.avito.android.snackbar.proxy
+
+import androidx.annotation.VisibleForTesting
+import com.google.android.material.snackbar.Snackbar
+
+// VisibleForTesting doesn't work in Kotlin https://issuetracker.google.com/issues/140642032
+// Fixed in AS 4.0
+@VisibleForTesting
+object SnackbarProxyHolder {
+    var proxy: SnackbarProxy? = null
```
sergioboi, line 10: I consider don't create the platform realization like in the ToastProxy behaviour

subprojects/android-test/snackbar-rule/src/main/kotlin/com/avito/android/test/app/second/SnackbarProxyRule.kt
```diff
@@ -0,0 +1,62 @@
...
+        }
+    }
+
+    fun clear() {
+        proxy.snackbarTexts.clear()
+    }
+
+    @SuppressLint("VisibleForTests")
+    private class Proxy : SnackbarProxy {
+        val snackbarTexts = mutableListOf<String>()
```
sergioboi: I decide to make the model simple. And do not save the whole Snackbar.
If we think I should save Snackbar or create a model to replace Sting we could discuss it
But I consider that code simple and straight forward

sergioboi:
- [ ] Create documentation after review

subprojects/android-test/snackbar-rule/src/main/kotlin/com/avito/android/test/app/second/SnackbarProxyRule.kt
```diff
@@ -0,0 +1,62 @@
...
+import android.widget.TextView
+import com.avito.android.rule.SimpleRule
+import com.avito.android.snackbar.proxy.SnackbarProxy
+import com.avito.android.snackbar.proxy.SnackbarProxyHolder
+import com.avito.android.test.waitFor
+import com.google.android.material.snackbar.Snackbar
+import org.junit.Assert
+import com.google.android.material.R as google_R
+
+class SnackbarProxyRule : SimpleRule() {
```
dsvoronin: i vote for `SnackbarRule()` - proxy is an implementation detail

samples/test-app-without-backward-compatibility/src/androidTest/kotlin/com/avito/android/test/app/second/SnackbarProxyTest.kt
```diff
@@ -0,0 +1,100 @@
...
+
+        snackbarRule.isShownLast("snackbar number 3")
+    }
+
+    @Test
+    fun check_snackbar_is_shown_after_delay() {
+        rule.launchActivity(null)
+        Screen.snackbarProxyScreen.clickShowSnackbar()
+
+        Thread.sleep(1000L)
```
dsvoronin: i know that the intention is to show that assertion is reliable even with a delay, but sleep catches the eye, we strongly encourage users not to use Thread.sleep();

Maybe wrap it in function like `artificialDelay()` to show it clear ?

settings.gradle.kts
dsvoronin: But what are you trying to achieve? 
Use new breaking API right away, without new infra version?
Is snackbar change even breaking?

subprojects/android-test/snackbar-rule/src/main/kotlin/com/avito/android/test/app/second/SnackbarProxyRule.kt
```diff
@@ -0,0 +1,62 @@
...
+
+    private val proxy = Proxy()
+
+    override fun before() {
+        SnackbarProxyHolder.proxy = proxy
+    }
+
+    fun isShown(text: String) {
```
dsvoronin: Consider adding an interface `SnackbarChecks` and a field `checks` to follow framework style

dsvoronin: `isShownWithText` reads easier to my taste

dsvoronin: Consider adding function with `Matcher<String>` argument, and a `contains` one

subprojects/android-test/snackbar-rule/src/main/kotlin/com/avito/android/test/app/second/SnackbarProxyRule.kt
dsvoronin: Snackbar also has an action button to tap on and it's text to check
Should we also support it right in this PR to be sure we don't like basic functionality?

subprojects/android-test/snackbar-rule/src/main/kotlin/com/avito/android/test/app/second/SnackbarProxyRule.kt
```diff
@@ -0,0 +1,62 @@
...
+        waitFor {
+            Assert.assertTrue("Snackbar with text=\"$text\" wasn't shown", proxy.snackbarTexts.any { it == text })
+        }
+    }
+
+    fun isShownLast(text: String) {
```
dsvoronin: Are you sure we need this check? 
It doesn't look like a simple API for me now;

Real world usage is : 'was snackbar with this text(or containtig) shown'? and you probably don't need anything else; besides tap on action

Same goes for `isShownTimes`

subprojects/android-lib/snackbar-proxy/src/main/kotlin/com/avito/android/snackbar/proxy/SnackbarProxy.kt
dsvoronin: Why safe call in `showSnackbar` then? It make sense to fail there

dsvoronin (changes requested):

settings.gradle.kts
sergioboi: I'm trying to use new snackbar modules. Without that change, I will fail on configuration time with the previous binary version

subprojects/android-lib/snackbar-proxy/src/main/kotlin/com/avito/android/snackbar/proxy/SnackbarProxy.kt
sergioboi: When the proxy is null then nothing to do, when the proxy is not null I call a method on it

subprojects/android-test/snackbar-rule/src/main/kotlin/com/avito/android/test/app/second/SnackbarProxyRule.kt
sergioboi: I don't know if that check makes sense I could add the possibility for checking the action button text.

subprojects/android-test/snackbar-rule/src/main/kotlin/com/avito/android/test/app/second/SnackbarProxyRule.kt
sergioboi: NP

samples/test-app-without-backward-compatibility/src/androidTest/kotlin/com/avito/android/test/app/second/SnackbarProxyTest.kt
sergioboi: Agree and will add a comment that we simulate delay behavior for testing our framework

subprojects/android-test/snackbar-rule/src/main/kotlin/com/avito/android/test/app/second/SnackbarProxyRule.kt
sergioboi: I will ask our user. I thought it's common behaviour

subprojects/android-test/snackbar-rule/src/main/kotlin/com/avito/android/test/app/second/SnackbarProxyRule.kt
sergioboi: NP

sergioboi: Done

sergioboi: Done

subprojects/android-lib/snackbar-proxy/src/main/kotlin/com/avito/android/snackbar/proxy/SnackbarProxy.kt
```diff
@@ -0,0 +1,20 @@
+package com.avito.android.snackbar.proxy
+
+import androidx.annotation.VisibleForTesting
+import com.google.android.material.snackbar.Snackbar
+
+// VisibleForTesting doesn't work in Kotlin https://issuetracker.google.com/issues/140642032
+// Fixed in AS 4.0
```
eugene-krivobokov, line 7: It's a stable release and it works with AGP 3+. It's no use to comment old issues.

subprojects/android-test/snackbar-rule/build.gradle.kts
```diff
@@ -0,0 +1,12 @@
+plugins {
```
int02h, line 1: Why do you decide to put it into a separate module? IMHO it's an overkill to make two modules for every rule or check in our test framework.

subprojects/android-test/snackbar-rule/src/main/kotlin/com/avito/android/test/app/second/SnackbarChecks.kt
```diff
@@ -0,0 +1,63 @@
...
+import org.hamcrest.Matcher
+import org.hamcrest.MatcherAssert
+import org.hamcrest.TypeSafeMatcher
+import ru.avito.util.Is
+
+class SnackbarChecks internal constructor(
+    private val proxy: SnackbarRule.Proxy
+) {
+
+    fun isShownWithTextMatches(text: Matcher<String>) {
```
int02h: Maybe we should name it shorter? Like `isShownWith()`?

subprojects/android-test/snackbar-rule/src/main/kotlin/com/avito/android/test/app/second/SnackbarChecks.kt
```diff
@@ -0,0 +1,63 @@
...
+        isShownWithTextMatches(text, 1)
+    }
+
+    fun isShownWithExactlyText(text: String) {
```
int02h: And here `isShownWithText`. I think it would be enough.

subprojects/android-test/snackbar-rule/src/main/kotlin/com/avito/android/test/app/second/SnackbarChecks.kt
eugene-krivobokov: I'd prefer to see in API just overloaded methods when we have different matchers.

```kotlin
isShownWithText("text")
isShownWithText(containsString("text"))
```

subprojects/android-lib/snackbar-proxy/src/main/kotlin/com/avito/android/snackbar/proxy/SnackbarProxy.kt
sergioboi: I faced that VisibleForTesting didn't work correctly

subprojects/android-test/snackbar-rule/build.gradle.kts
sergioboi: I did it like for toast-rule. But we could overthink this strategy

samples/test-app-without-backward-compatibility/src/androidTest/kotlin/com/avito/android/test/app/second/SnackbarProxyTest.kt
```diff
@@ -0,0 +1,111 @@
...
+        rule.launchActivity(null)
+
+        Screen.snackbarProxyScreen.clickShowSnackbar()
+        Screen.snackbarProxyScreen.clickShowSnackbar()
+
+        snackbarChecks.isShownLastWithExactlyText("snackbar number 2")
+    }
+
+    @Test
+    fun check_snackbar_is_shown_N_times() {
```
eugene-krivobokov: Why do we need to support this check? Have we seen it in tests?

subprojects/android-lib/snackbar-proxy/src/main/kotlin/com/avito/android/snackbar/proxy/SnackbarProxy.kt
eugene-krivobokov: Do you use AS 3+?

subprojects/android-test/snackbar-rule/build.gradle.kts
eugene-krivobokov: At least, It's good to separate classes that you can use in tests without our test library. 
All further separations are questionable for me.

samples/test-app-without-backward-compatibility/src/androidTest/kotlin/com/avito/android/test/app/second/SnackbarProxyTest.kt
sergioboi: Asked users. As for me, it looks preferable

sergioboi: Scenarios
- Check zero times when you don't want a snackbar
- Check two times because you have an old one snackbar

settings.gradle.kts
dsvoronin: ok, i understand, but this will be outdated right after release

subprojects/android-lib/snackbar-proxy/src/main/kotlin/com/avito/android/snackbar/proxy/SnackbarProxy.kt
dsvoronin: i can't see how production code will use `showSnackbar` method, but i'm afraid that it won't show snackbar when it's needed.
Even if it's not the case, this code style creates that type of anxiety

subprojects/android-test/snackbar-rule/build.gradle.kts
dsvoronin: Same, but looking from user's perspective i would be glad to have a choice not to apply features i don't need

samples/test-app-without-backward-compatibility/src/androidTest/kotlin/com/avito/android/test/app/second/SnackbarProxyTest.kt
eugene-krivobokov: Ok, zero times looks like a valid case.

eugene-krivobokov (approved):

dsvoronin (approved):

PR #451, opened 2020-06-18
MBS-8739 Add Snackbar.md

sergioboi:
SnackbarRule documentation


docs/content/docs/test/Snackbar.md
```diff
@@ -0,0 +1,72 @@
+# Motivation
+
+**Why do we need a custom component for snackbar testing?** \
+Snackbar nature is unpredictable: it appears and disappears in asynchronious way. \
```
dsvoronin: i see nothing unpredictable in snackbar, it just hides with delay

sergioboi: Unpredictable means that we can't predict when it will appear or disappear
Because we have no API to watch that behaviour

dsvoronin: that's not what you wrote )

docs/content/docs/test/Snackbar.md
```diff
@@ -0,0 +1,72 @@
+# Motivation
```
eugene-krivobokov:
```suggestion
# Snackbar
```
It looks ok to have a short motivational description right after a header.

eugene-krivobokov (approved):

PR #455, opened 2020-06-19
Don't fail MockApiRule if test hasn't run due to another failed TestRule

eugene-krivobokov:


subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/api/AbstractMockApiRule.kt
```diff
@@ -7,13 +7,12 @@ abstract class AbstractMockApiRule<T : RequestRegistry> : SimpleRule() {
...
-                reset()
-                registry = null
-            }
+        val registry = registry ?: return // can be null if another rule failed after us
+        try {
+            registry.registeredMocks.values.forEach(ApiRequest::verify)
+        } finally {
+            registry.reset()
+            this.registry = null
         }
```
eugene-krivobokov, line 16: Example of the problem.

```kotlin
val apiRule = DummyMockApiRule()

    @get:Rule
    val rule = RuleChain.outerRule(apiRule)
        .around(FailedRule())

    @Test
    fun test() {
        apiRule.stub {
            // this code was not invoked
        }
    }

    class FailedRule : SimpleRule() {

        override fun before() {
            throw RuntimeException("Original failure")
        }
    }

    class DummyMockApiRule : AbstractMockApiRule<RequestRegistry>() {

        override fun createRegistry(): RequestRegistry = object : RequestRegistry() {
            override fun resetApi() {
            }
        }
    }
```

```kotlin
abstract class SimpleRule : TestRule {

    override fun apply(base: Statement, description: Description): Statement =
        object : Statement() {
            override fun evaluate() {
                before()
                try {
                    base.evaluate()
                } finally {
                    after()
                }
            }
        }
```

Execution order:
- ApiRule.apply -> before() - do nothing
- ApiRule.apply -> base.evaluate() - delegates to Statement
- FailedRule.apply -> before() - fails! The test won't run.
- ApiRule.apply -> after() in finally block. It fails and the original error is missed.

Conclusion: 
TestRule must be ready that test hasn't run for any reason.

dsvoronin (approved):

subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/api/AbstractMockApiRule.kt
sergioboi: @eugene-krivobokov If ApiRile before was called why it does nothing?
If ApiRule clears a resource in `after` does it mean that it initializes the resource in the `before`?

eugene-krivobokov: Yes, this might be a better option - initialize in before and make no assumptions about test execution.

PR #474, opened 2020-06-29
MBS-8782 Enable impact analysis in CI steps by default

eugene-krivobokov:
It's a stable optimization and we don't want to miss it accidentally.


subprojects/gradle/cd/src/main/kotlin/com/avito/ci/BuildStepListExtension.kt
```diff
@@ -29,7 +29,7 @@ open class BuildStepListExtension(internal val name: String, objects: ObjectFact
     internal val steps: ListProperty<BuildStep> = objects.listProperty(BuildStep::class.java).empty()
 
     //todo property
```
eugene-krivobokov, line 31: It also requires to migrate all other public build steps variables to properties for the sake of consistency.
I think it's out of scope for now and not a blocker.

dsvoronin (approved):

int02h (approved):

PR #475, opened 2020-06-29
MBS-8753 Update verdict message

sergioboi:
prettify verdict details

### DIFF
--- a/subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/TestRunResult.kt
+++ b/subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/TestRunResult.kt
@@ -19,10 +19,26 @@ data class TestRunResult(
             override val message: String
         ) : Verdict()
 
-        data class Failed(
+        data class Failure(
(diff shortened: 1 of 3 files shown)


subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/TestRunResult.kt
```diff
@@ -95,8 +142,23 @@ data class TestRunResult(
...
+                    message,
+                    arrayOf(this, throwable)
+                )
+            } else {
+                this
+            }
+        }
+        else -> throwable
+    }
 }
```
eugene-krivobokov, line 162: I'd prefer to avoid triple nesting for just null checks.

```kotlin
private fun Throwable?.composeWith(throwable: Throwable?): Throwable? {
    if (this == null) return throwable
    if (throwable == null) return this

    val message = "${message}. \n ${throwable.message}"
    return CompositeException(
        message,
        arrayOf(this, throwable)
    )
}
```

eugene-krivobokov (approved):

PR #476, opened 2020-06-29
unify mockwebserver test api

dsvoronin:
needs check against avito


subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/mock/Mock.kt
```diff
@@ -1,7 +1,12 @@
 package com.avito.android.mock
 
+import com.avito.test.http.RequestData
 import okhttp3.mockwebserver.MockResponse
 
+@Deprecated(
+    "use same class from common module",
```
eugene-krivobokov:
```suggestion
    "Since 2020.10; Use the same class from common module",
```

eugene-krivobokov:
Can you please describe shortly important changes in modules and behavior?
It's hard to get the idea of many small changes.

dsvoronin:
> Can you please describe shortly important changes in modules and behavior?
> It's hard to get the idea of many small changes.

Yep, the goal is to remove duplicated library, wrapper for MockWebServer for testing.
We had one in `:test-okhttp` module, and almost the same in `:inhouse-test-runner`,

First one was used in opensource unit(integration) tests, and second one in internal ui tests
Over time changes in first and second lib created API incompatibility

In this PR i moved all the code to `:test-okhttp`, so i have to change all usages in unit tests (it's easier and `inhouse-test-runner` version is more mature)

Next step will be changing usages in internal project to `:test-okhttp` ones (imports only),
and deleting deprecated lib in `:in-house-test-runner`

eugene-krivobokov (approved):

PR #483, opened 2020-06-30
MBS-7702 specify commons-io explicitly to use fixed version of Tailer

int02h:
The problem with mojibake in the log was in Tailer from an outdated version of `commons-io`. That outdated Tailer wasn't considering `Charset` when reading strings from a file. Android's LogCat has UTF-8 encoding but Java has UTF-16 encoding. So Cyrillic symbols were shown incorrectly. 

Futhermore outdated version of `commons-io` was given transitively through `project(":gradle:android")`. I've updated `commons-io` to the latest version where Tailer uses `Charset.defaultCharset()` to decode string.

### DIFF
--- a/subprojects/gradle/instrumentation-tests/build.gradle.kts
+++ b/subprojects/gradle/instrumentation-tests/build.gradle.kts
@@ -38,6 +38,9 @@ dependencies {
     implementation(Dependencies.retrofit)
     implementation(Dependencies.kotson)
     implementation(Dependencies.funktionaleTry)
+    implementation(Dependencies.commonsIo) {
+        because("LogcatBuffer.Impl.tailer needs to consider Charset (https://issues.apache.org/jira/browse/IO-354)")
+    }
 
     testImplementation(project(":gradle:test-project"))
     testImplementation(project(":gradle:logging-test-fixtures"))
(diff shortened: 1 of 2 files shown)


int02h:
I am not sure whether to provide Charset explicitly when creating Tailer or rely on `Charset.defaultCharset()`

sergioboi (approved):

dsvoronin (approved):

subprojects/gradle/instrumentation-tests/build.gradle.kts
```diff
@@ -38,6 +38,9 @@ dependencies {
     implementation(Dependencies.retrofit)
     implementation(Dependencies.kotson)
     implementation(Dependencies.funktionaleTry)
+    implementation(Dependencies.commonsIo) {
+        because("LogcatBuffer.Impl.tailer needs to consider Charset (https://issues.apache.org/jira/browse/IO-354)")
```
dsvoronin, line 42: love it!
i mean ... this explanation of deps usage

eugene-krivobokov (approved):

eugene-krivobokov:
> I am not sure whether to provide Charset explicitly when creating Tailer or rely on Charset.defaultCharset()

 I think `defaultCharset` will do. If I'm not mistaken, It's a Charset from Gradle/JVM process in our Docker image.

PR #484, opened 2020-07-02
MBS-8791 Fix PerformanceProvider.isPresent=true

sergioboi:
we must use return Providers.notDefined in Profiders.flatMap to make Provider.isPresent=false


subprojects/gradle/performance/src/main/kotlin/com/avito/performance/PerformancePlugin.kt
```diff
@@ -169,7 +175,7 @@ open class PerformancePlugin : Plugin<Project> {
                                 }
                             }
 
-                        val performanceProvider = gitState.map {
+                        val performanceProvider: Provider<Task> = gitState.flatMap {
```
eugene-krivobokov, line 178: I didn't get the difference:

```
/**
     * Returns a new {@link Provider} whose value is the value of this provider transformed using the given function.
     *
     * <p>The new provider will be live, so that each time it is queried, it queries this provider and applies the transformation to the result.
     * Whenever this provider has no value, the new provider will also have no value and the transformation will not be called.
     *
     * <p>When this provider represents a task or the output of a task, the new provider will be considered an output of the task and will carry dependency information that Gradle can use to automatically attach task dependencies to tasks that use the new provider for input values.</p>
     *
     * @param transformer The transformer to apply to values. May return {@code null}, in which case the provider will have no value.
     * @since 4.3
     */
    <S> Provider<S> map(Transformer<? extends S, ? super T> transformer);

    /**
     * Returns a new {@link Provider} from the value of this provider transformed using the given function.
     *
     * <p>The new provider will be live, so that each time it is queried, it queries this provider and applies the transformation to the result.
     * Whenever this provider has no value, the new provider will also have no value and the transformation will not be called.
     *
     * <p>Any task details associated with this provider are ignored. The new provider will use whatever task details are associated with the return value of the function.</p>
     *
     * @param transformer The transformer to apply to values. May return {@code null}, in which case the provider will have no value.
     * @since 5.0
     */
    @Incubating
    <S> Provider<S> flatMap(Transformer<? extends Provider<? extends S>, ? super T> transformer);
```

sergioboi: Devil is the return type. With `map` return value was Provider<Any> where Any transforms to Provider<Task> so the result was Provider<Provider<Task>> and outer Provider has a value that was a `Provider.notDefined()`

Behavior similar to RxJava Observable `map` and `flatMap` difference. `flatMap` can generate a new Observable and `map` is only able to transform values.

sergioboi: I wrote tests for making code more clear and will commit them tomorrow

dsvoronin (approved):

eugene-krivobokov (approved):

PR #493, opened 2020-07-06
MBS-7793 de-obfuscate log that uploaded to RV

int02h:


buildscript/src/main/kotlin/Dependencies.kt
```diff
@@ -54,6 +54,7 @@ object Dependencies {
 
     //https://r8.googlesource.com/r8/
     val r8 = "com.android.tools:r8:1.6.89"
+    val proguardRetrace = "net.sf.proguard:proguard-retrace:6.2.2"
```
eugene-krivobokov, line 57: Does this version correlate with R8 version in a client project?

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/InstrumentationTestsTask.kt
```diff
@@ -159,7 +165,8 @@ abstract class InstrumentationTestsTask @Inject constructor(
                     registry = registry.get(),
                     reportConfig = reportConfig,
                     reportFactory = reportFactory,
-                    reportCoordinates = reportCoordinates
+                    reportCoordinates = reportCoordinates,
+                    proguardMappings = proguardMappings
```
eugene-krivobokov: Can we pass Provider here?

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/InstrumentationTestsTask.kt
```diff
@@ -123,6 +125,10 @@ abstract class InstrumentationTestsTask @Inject constructor(
             reportConfig
         )
 
+        val proguardMappings = project.tasks.withType<ProguardConfigurableTask>()
```
eugene-krivobokov, line 128: Should we consider only the "actual" build variant?

subprojects/common/retrace/src/test/kotlin/com/avito/retrace/ProguardRetracerTest.kt
```diff
@@ -0,0 +1,59 @@
...
+internal class ProguardRetracerTest {
+
+    @Test
+    fun retrace(@TempDir mappingDir: File) {
+        val mappingFile = writeMapping(mappingDir)
+        val result = ProguardRetracer.Impl(listOf(mappingFile)).retrace(STACKTRACE)
+
+        assertThat(result).doesNotContain("b.a.a.a.a.a(:5)")
+        assertThat(result).contains("com.avito.android.ui.Crasher.crash")
+    }
```
eugene-krivobokov, line 18: :+1:

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/InstrumentationTestsTask.kt
eugene-krivobokov: Should we filter only tasks from a build variant in `android.testBuildType`?

buildscript/src/main/kotlin/Dependencies.kt
int02h: I don't think so. As I understand the format of mapping files as old as ProGuard so there is no need to keep these versions in some kind of sync.

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/InstrumentationTestsTask.kt
int02h: To do lazy calculation? Or what `Provider` do you mean?

eugene-krivobokov:
```kotlin
ProguardConfigurableTask {

@get:OutputFile
abstract val mappingFile: RegularFileProperty <---
```

As I see in the documentation, `WorkParameters` supports properties - https://docs.gradle.org/current/javadoc/org/gradle/workers/WorkParameters.html

`.filter { it.mappingFile.isPresent }` - this is a logic. 
I don't see reasons to spread any logic between Task and WorkAction

eugene-krivobokov: Found more important reason in a discussion. 
Mappings must be input for the task.

subprojects/common/retrace/src/main/kotlin/com/avito/retrace/ProguardRetracer.kt
```diff
@@ -0,0 +1,36 @@
...
+interface ProguardRetracer {
+
+    fun retrace(content: String): String
+
+    class Impl(
+        private val mappings: List<File>
+    ) : ProguardRetracer {
+
+        override fun retrace(content: String): String {
+            var result = content
```
sergioboi, line 23: It's strange we write into `result` in `forEach` but then return one value
Should we somehow filter mappings to find firstone

subprojects/common/retrace/src/main/kotlin/com/avito/retrace/ProguardRetracer.kt
```diff
@@ -0,0 +1,36 @@
...
+            mappings.map { ReTrace(ReTrace.STACK_TRACE_EXPRESSION, false, it) }
```
sergioboi: We create a `ReTrace` for the same mapping file on each call of `retrace(content)`. It looks inefficient to read mapping file too many times

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/InstrumentationTestsTask.kt
```diff
@@ -123,6 +125,10 @@ abstract class InstrumentationTestsTask @Inject constructor(
...
+            .filter { it.mappingFile.isPresent }
+            .map { it.mappingFile.get().asFile }
```
sergioboi: I consider we need to avoid using `get` on configuration time.
You should use `flatMap { it.mappingFile }` and then at execution time you would extract provider value by `get()`

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/InstrumentationTestsTask.kt
```diff
@@ -123,6 +125,10 @@ abstract class InstrumentationTestsTask @Inject constructor(
             reportConfig
         )
 
+        val proguardMappings = project.tasks.withType<ProguardConfigurableTask>()
+            .filter { it.mappingFile.isPresent }
```
sergioboi: I want to request a change

sergioboi (changes requested):

subprojects/common/retrace/src/main/kotlin/com/avito/retrace/ProguardRetracer.kt
int02h: We have multiple mapping. In logs can be several stacktraces that mapped into different mapping.txt. So this code iterate over all available mappings and trying to apply them to log.

sergioboi: I understand that behavior. But It looks like we rewriting previous results in forEach and we will see only the last write to `result` variable

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/InstrumentationTestsTask.kt
int02h: It is not configuration time. It is inside `doWork` of the task. But now I'm refactoring and moving this code into a plugin.

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/InstrumentationTestsTask.kt
int02h: Fixed

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/InstrumentationTestsTask.kt
int02h: I've refactored this piece of code

subprojects/common/retrace/src/main/kotlin/com/avito/retrace/ProguardRetracer.kt
int02h: Fixed that

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/InstrumentationTestsTask.kt
```diff
@@ -54,6 +54,14 @@ abstract class InstrumentationTestsTask @Inject constructor(
     @InputFile
     val testApkOnTargetCommit: RegularFileProperty = objects.fileProperty()
 
+    @Optional
```
dsvoronin, line 57: https://github.com/gradle/gradle/issues/2016
be aware that `@Optional` + `@InputFile` is a dangerous combo

dsvoronin (approved):

subprojects/common/retrace/src/main/kotlin/com/avito/retrace/ProguardRetracer.kt
int02h: We sequentially apply mappings to logcat. I don't know another away except iterating over mappings and apply them to logcat one-by-one. And updating variable with logcat on every iteration.

Could you please provide piece of code that looks good to you?

subprojects/common/retrace/src/main/kotlin/com/avito/retrace/ProguardRetracer.kt
int02h: Fixed

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/InstrumentationTestsTask.kt
int02h: This code was rewritten

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/InstrumentationTestsTask.kt
int02h: I did it like other properties. For instance for APK.

int02h: But thank you for pointing the issue

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/configuration/GradleInstrumentationPluginConfiguration.kt
```diff
@@ -34,6 +34,9 @@ object InstrumentationPluginConfiguration {
 
         var unitToChannelMap: Map<String, String> = emptyMap()
 
+        var applicationProguardMapping: String? = null
```
sergioboi: Is it mapping or path to mapping? I consider the type of this `File?`

subprojects/gradle/kotlin-dsl-support/src/main/kotlin/com/avito/kotlin/dsl/TaskExtensions.kt
```diff
@@ -18,6 +18,11 @@ inline fun <reified T : Task> Task.dependencyOn(anotherTaskProvider: TaskProvide
     dependsOn(anotherTaskProvider)
 }
 
+inline fun <reified T : Task> Task.dependencyOn(anotherTask: T, configuration: (T) -> Unit) {
```
sergioboi, line 21: Could we get a TaskProvider rather than Task?

subprojects/common/retrace/src/main/kotlin/com/avito/retrace/ProguardRetracer.kt
sergioboi: `retraces.first { }` or `retraces.scan { }`

sergioboi (changes requested):

subprojects/common/retrace/src/main/kotlin/com/avito/retrace/ProguardRetracer.kt
int02h: Why `scan` is better than `forEach`?

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/configuration/GradleInstrumentationPluginConfiguration.kt
int02h: I did it like other properties around.

subprojects/gradle/kotlin-dsl-support/src/main/kotlin/com/avito/kotlin/dsl/TaskExtensions.kt
int02h: Why do I need TaskProvider if I already have a Task at hand?

subprojects/common/retrace/src/main/kotlin/com/avito/retrace/ProguardRetracer.kt
sergioboi: I consider that Retrace API doesn't help us.

subprojects/gradle/kotlin-dsl-support/src/main/kotlin/com/avito/kotlin/dsl/TaskExtensions.kt
sergioboi: Only If we would use `named` (

subprojects/common/retrace/src/main/kotlin/com/avito/retrace/ProguardRetracer.kt
int02h: Furthermore `scan` is an experimental API. Is this any significant reason to use it?

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/configuration/GradleInstrumentationPluginConfiguration.kt
int02h: Replace `String` with `File`

sergioboi (approved):

PR #497, opened 2020-07-09
Mbs 8795

sergioboi:


sergioboi:
### `Test showViews` 

ReportViewer ui shows `Flaky` icon when you switch `API` tab

**stable on API28**
Screenshot 2020-07-09 at 18 23 36
**flaky on API22**
Screenshot 2020-07-09 at 18 23 30

### `Test collapseViews` 

ReportViewer ui shows `Flaky` on both tabs

**flaky on API28**
Screenshot 2020-07-09 at 18 23 24
**flaky on API22**
Screenshot 2020-07-09 at 18 23 17

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/configuration/InstrumentationFilter.kt
```diff
@@ -24,6 +24,8 @@ abstract class InstrumentationFilter(val name: String) {
         private val annotations = Filter<String>()
         private val prefixes = Filter<String>()
 
+        var excludeFlaky = false
```
sergioboi, line 27: I had to separate flaky filtration from other annotations because AnnotationFilter is able to filter only by annotation name but in Flaky case, we have to parse annotation fields

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/suite/filter/TestsFilter.kt
```diff
@@ -42,7 +47,8 @@ interface TestsFilter {
         val name: String,
         val annotations: List<AnnotationData>,
         val deviceName: DeviceName,
-        val api: Int
+        val api: Int,
+        val flakiness: Flakiness
```
sergioboi, line 51: When we create that model we already parsed annotations. So, I decided to reuse the result of parsing

subprojects/gradle/instrumentation-tests/src/test/kotlin/com/avito/instrumentation/suite/filter/FilterFactoryFactory.kt
```diff
@@ -0,0 +1,37 @@
...
+            buildId = "stub"
+        )
+    ): FilterFactory {
+        return FilterFactory.create(
+            filterData = filter,
+            impactAnalysisResult = impactAnalysisFile,
+            reportConfig = reportConfig,
+            factory = object : Report.Factory {
+                override fun createReport(config: Report.Factory.Config): Report {
+                    TODO("Not yet implemented")
```
eugene-krivobokov, line 28: Don't we need it?

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/suite/filter/TestsFilter.kt
```diff
@@ -16,6 +17,10 @@ interface TestsFilter {
                 name, "test has SkipSdk with value sdk=$sdk"
             )
 
+            class HaveFlakyAnnotation(name: String, sdk: Int) : Excluded(
```
eugene-krivobokov:
```suggestion
            class HasFlakyAnnotation(name: String, sdk: Int) : Excluded(
```
"A _test_ has flaky annotation." - is it correct?

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/suite/filter/ExcludeBySkipOnSdkFilter.kt
```diff
@@ -1,7 +1,7 @@
 package com.avito.instrumentation.suite.filter
 
 // it makes [name] instance field. It needs for gson
-internal class ExcludeBySdkFilter : TestsFilter {
+internal class ExcludeBySkipOnSdkFilter : TestsFilter {
 
     override val name = "ExcludeBySdkFilter"
```
eugene-krivobokov: Should we rename "name" accordingly?

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/configuration/InstrumentationFilter.kt
eugene-krivobokov: Can we use a more helpful model for annotation in AnnotationFilter?
I suppose that we will have other annotations with parameters.
It would be good to have a common abstraction.

eugene-krivobokov (approved):

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/configuration/InstrumentationFilter.kt
sergioboi: I don't see scenarios. 
Users don't need knowledge about Annotations values.
There are two annotation types:
- Marker - you simple use name for filtering
- Complex with logic: Flaky, SkipOnSdk. Filters are written with parsing and custom logic. I consider staying logic separated.
One option I see is the merging of SkipOnSdk and Flaky. We can add a `skip` parameter to Flaky annotation

subprojects/gradle/instrumentation-tests/src/test/kotlin/com/avito/instrumentation/suite/filter/FilterFactoryFactory.kt
sergioboi: Yes it testing entity

eugene-krivobokov (approved):

PR #498, opened 2020-07-10
MBS-8682 Make ownership extension optional

eugene-krivobokov:


subprojects/gradle/code-ownership/src/main/kotlin/com/avito/android/CheckProjectDependenciesOwnershipTask.kt
```diff
@@ -29,42 +28,39 @@ abstract class CheckProjectDependenciesOwnershipTask : DefaultTask() {
...
                 }
             }
-        }
     }
 
-    private fun projectDependencies() = project.internalModule
-        .implementationConfiguration
-        .dependencies
-        .filterNot {
-            it.module.project.pluginManager.hasPlugin("java-platform")
```
eugene-krivobokov, line 66: Sometimes this check failed and we've got an error exactly for `platform` module:
```
at com.avito.android.CodeOwnershipExtensionKt.getOwnership(CodeOwnershipExtension.kt:36)
at com.avito.android.CheckProjectDependenciesOwnershipTask.checkPrivateDependenciesRule(CheckProjectDependenciesOwnershipTask.kt:33)
```
I haven't found reasons. 
The task is not cacheable, not UP-TO-DATE. We check the plugin in the execution phase.
All my attempts to test, debug, or log additional info were failed - bug become unreproducible.
It looks like different values due to some race but I couldn't collect proofs.

I've decided to avoid this check and make ownership extension optional. The default state seems appropriate and more robust in similar cases with other plugins.

subprojects/gradle/code-ownership/src/main/kotlin/com/avito/android/CodeOwnershipExtension.kt
```diff
@@ -33,4 +33,4 @@ open class CodeOwnershipExtension(
 }
 
 val ExtensionContainer.ownership: CodeOwnershipExtension
-    get() = findByType() ?: error("ownership should be available")
+    get() = findByType() ?: CodeOwnershipExtension()
```
sergioboi, line 36: I consider seeing Stub instance

eugene-krivobokov: `CodeOwnershipExtension` can be called a stub. It contains valid defaults by design.
Any customization is optional in this plugin.
Additional abstraction would require compatible contract. I don't see any value to split it.

sergioboi (approved):

PR #500, opened 2020-07-10
MBS-8896 Swallowed exception in SimpleRule

eugene-krivobokov:


samples/test-app/src/androidTest/kotlin/com/avito/android/ui/test/SimpleRuleTest.kt
```diff
@@ -0,0 +1,52 @@
...
+
+        thrown.expect(IllegalStateException::class.java)
+        thrown.expectMessage("Error from test")
+
+        throw IllegalStateException("Error from test")
+    }
+
+    private class RuleError : java.lang.RuntimeException()
+
+    class FailedRule : SimpleRule() {
```
eugene-krivobokov, line 39: Technically, I can test changes in unit tests. I wanted to cover them in integration inside AndroidJUnitRunner to be sure.

subprojects/android-test/junit-utils/src/main/java/com/avito/android/rule/SimpleRule.kt
```diff
@@ -10,15 +10,41 @@ abstract class SimpleRule : TestRule {
         object : Statement() {
             override fun evaluate() {
                 before()
-                try {
-                    base.evaluate()
-                } finally {
-                    after()
```
eugene-krivobokov, line 16: The problem:
If test fails (`base.evaluate()`) and test rule fails (`after()`), test's exception will be lost.

subprojects/android-test/junit-utils/src/main/java/com/avito/android/rule/SimpleRule.kt
```diff
@@ -10,15 +10,41 @@ abstract class SimpleRule : TestRule {
         object : Statement() {
             override fun evaluate() {
                 before()
```
sergioboi, line 12: What will happen If before throws an Exception?

eugene-krivobokov: This method in Rule executes prior to `@Before` methods and test itself.
If it fails, neither `@Before` nor test will be executed.
We will see an original and only one error.

subprojects/android-test/junit-utils/src/main/java/com/avito/android/rule/SimpleRule.kt
eugene-krivobokov: Additional note.
It's important to execute `after()` even in case of an error in tests. 
Custom Rules are counting on this behavior. They clean up a shared state.

dsvoronin (approved):

PR #504, opened 2020-07-14
Add target kotlin-dsl compatibility

sergioboi:


sergioboi:
### **Before**
Screenshot 2020-07-14 at 20 12 59
### **After**
Screenshot 2020-07-14 at 20 14 43

eugene-krivobokov (approved):

PR #508, opened 2020-07-22
MBS-8026 Remove from logs unrelated exceptions

sergioboi:
### UIAutomation not connected
Exception occurs when test execution ended during `Instrumentation.finish()`
Screenshot 2020-07-21 at 17 23 14
Let’s look closer to the `finish` function. As we see `Instrumentation` tried to `mUiAutimation.disconnect` and `mUiAtomation` must exist and be not destroyed.
Screenshot 2020-07-21 at 17 24 59
I’ve searched and found only one place where `mUiAutomation` is created.
It’s the `Instrumentation.getUiAutomation` function.
Screenshot 2020-07-21 at 17 27 01
So I decided to find all places where the `getUiAutomation` is executed.
I overrode the `getUiAutomation` and add logging there.
I used logging because I think attach and debug could affect the behavior more then logging.
Screenshot 2020-07-21 at 17 28 54
I knew that we used video capturing that uses `mUiAutomation` on devices with API greater than 23. So I observed two API versions: 22, 28
I ran a simple test 10 times on each of those APIs.
#### on 22 API
I didn’t see `getUiAutomation` logs. So anyone didn’t call `getUiAutomation` function but `UiAutomation not connected` still occurs
Sum up: I don’t know Who, When, and How created `mUiAutomation`. So I don’t have any idea how to influence on that behavior.
#### on 28 API
Logs contained `getUiAutomation`. And I could connect this with video capturing. `com.avito.android.test.report.video.VideoCapturerImpl#execute`
Screenshot 2020-07-21 at 17 33 42
More `getUiAutomation` logs. There we could see that system created `getUiAutomation` too

`UiAutomation not connected` still occurs
Sum up: `mUiAutomation` was created. But It’s not clear why exception occurs

Let’s look closer to `UiAutomatuion.disconnect()` function
Screenshot 2020-07-21 at 17 40 00￼
We got exception in the `throwIfNotConnectedLocked()`. 
Screenshot 2020-07-21 at 17 45 11
￼Screenshot 2020-07-21 at 17 45 49
On two above pictures we could see that `disconnect` made `mConnectionId = CONNECTION_ID_UNDEFINED` and exception occurs if `mConnectionId == CONNECTION_ID_UNDEFINED`
Sum up: If two threads call `disconnect` concurrently and It will lead to an exception. Because after second waiting thread will get the `mLock` it will fail at `throwIfNotConnectedLocked`

I found one place where `UiAutomation.disconnect()` is called. It’s the `Instrumentation.finish()`. So I decided to log the `finish` invocation
￼Screenshot 2020-07-21 at 17 51 50
I saw in logs only one `finish` for both APIs.
Sum up: We need to find other places where `UiAutomation.disconnect()` is called.

I searched through all usages of `instrumentation.getUiautomation` I was able to find but I didn’t see any `disconnect` calls

Summary: Exception occurs even on APIs where we don’t use the `UiAutomation` and I couldn’t find any related to `UiAutomation` logic. So I think we couldn’t control it and influence on that behavior.
The best decision is to catch that exception

### DIFF
--- a/subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/runner/InHouseInstrumentationTestRunner.kt
+++ b/subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/runner/InHouseInstrumentationTestRunner.kt
@@ -321,6 +321,21 @@ abstract class InHouseInstrumentationTestRunner :
         arguments.putString("newRunListenerMode", "true")
     }
 
+    override fun finish(resultCode: Int, results: Bundle?) {
+        try {
+            super.finish(resultCode, results)
+        } catch (e: IllegalStateException) {
+            // we made a research.
+            // It showed that IllegalStateException("UiAutomation not connected") occurs unrelated to our code.
+            // We use uiAutomation only for VideoCapture but without capturing this exception occurs with same frequency
+            if (e.message?.contains("UiAutomation not connected") != true) {
+                throw e
+            } else {
+                Log.d(tag, "Got UiAutomation not connected when finished")
+            }
+        }
+    }
+
     companion object {
         val instance: InHouseInstrumentationTestRunner by lazy {
             InstrumentationRegistry.getInstrumentation() as InHouseInstrumentationTestRunner


subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/runner/InHouseInstrumentationTestRunner.kt
```diff
@@ -321,6 +321,21 @@ abstract class InHouseInstrumentationTestRunner :
...
     }
 
+    override fun finish(resultCode: Int, results: Bundle?) {
+        try {
+            super.finish(resultCode, results)
+        } catch (e: IllegalStateException) {
+            // we made a research.
+            // It showed that IllegalStateException("UiAutomation not connected") occurs unrelated to our code.
+            // We use uiAutomation only for VideoCapture but without capturing this exception occurs with same frequency
+            if (e.message?.contains("UiAutomation not connected") != true) {
```
eugene-krivobokov, line 331: Let's check also the exception class to be sure

sergioboi: I think `catch` did it for us

sergioboi: @eugene-krivobokov Is this enough?

eugene-krivobokov (approved):

subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/runner/InHouseInstrumentationTestRunner.kt
eugene-krivobokov: sorry, missed that we are already in catch block with this error

PR #515, opened 2020-07-23
Mbs 9069 body access bug

dsvoronin:
Could not reproduce bug in test;
 tried access old dispatcher implementation from multiple threads concurrently, no luck; all is green

### DIFF
--- a/buildscript/src/main/kotlin/Dependencies.kt
+++ b/buildscript/src/main/kotlin/Dependencies.kt
@@ -24,7 +24,7 @@ object Dependencies {
     val retrofitConverterGson = "com.squareup.retrofit2:converter-gson:${Versions.retrofit}"
     val okhttp = "com.squareup.okhttp3:okhttp:${Versions.okhttp}"
     val okhttpLogging = "com.squareup.okhttp3:logging-interceptor:${Versions.okhttp}"
-    val okio = "com.squareup.okio:okio:2.1.0"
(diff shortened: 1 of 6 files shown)


sergioboi (approved):

buildscript/src/main/kotlin/Dependencies.kt
```diff
@@ -24,7 +24,7 @@ object Dependencies {
     val retrofitConverterGson = "com.squareup.retrofit2:converter-gson:${Versions.retrofit}"
     val okhttp = "com.squareup.okhttp3:okhttp:${Versions.okhttp}"
     val okhttpLogging = "com.squareup.okhttp3:logging-interceptor:${Versions.okhttp}"
-    val okio = "com.squareup.okio:okio:2.1.0"
+    val okio = "com.squareup.okio:okio:2.7.0"
```
eugene-krivobokov, line 27: Why do we use a custom version? It overrides a tested one from okhttp.

dsvoronin: Good question; 
my logic here that: 
1. we already using 2.6.0 in avito project
2. 2.x is api compatible and mostly consist of performance improvements and bugfixes

eugene-krivobokov: OkHttp 4.8.0 already uses Okio 2.7.0.
I'd preferer to override transitive dependencies only when we have reasons.

PR #528, opened 2020-08-04
MBS-9172 Disable dependency analysis plugin by default to mitigate GC limitations

eugene-krivobokov:
Sometimes we've got  and `GC overhead limit exceeded` error in PRs:

```
  * What went wrong:
  Execution failed for task ':gradle:kotlin-root:dependenciesReportMain'.
  > GC overhead limit exceeded
```
See examples in #527 builds.

Can't reproduce it locally and increased limits a bit to be safe.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,5 +1,5 @@
 kotlin.code.style=official
-org.gradle.jvmargs=-Xmx4g -XX:+UseGCOverheadLimit -XX:GCTimeLimit=10 -Dfile.encoding=UTF-8
+org.gradle.jvmargs=-Xmx5g -XX:+UseGCOverheadLimit -XX:GCTimeLimit=20 -Dfile.encoding=UTF-8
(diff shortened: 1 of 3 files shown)


gradle.properties
```diff
@@ -1,5 +1,5 @@
 kotlin.code.style=official
-org.gradle.jvmargs=-Xmx4g -XX:+UseGCOverheadLimit -XX:GCTimeLimit=10 -Dfile.encoding=UTF-8
+org.gradle.jvmargs=-Xmx5g -XX:+UseGCOverheadLimit -XX:GCTimeLimit=20 -Dfile.encoding=UTF-8
```
sergioboi: I propose next time do not apply analysis-plugin
because now we don't use it.
I want to see those resource usages necessary

eugene-krivobokov: Ok, agree for the next incident

int02h (approved):

PR #536, opened 2020-08-05
MBS-9071 Test BitmapMatcher common cases

eugene-krivobokov:


subprojects/android-test/ui-testing-core/src/main/kotlin/com/avito/android/test/matcher/BitmapMatcher.kt
```diff
@@ -17,17 +18,26 @@ class BitmapMatcher(private val bitmap: Bitmap) :
...
+                    "actual [${actualBitmap.height}x${actualBitmap.width}], compared [${bitmap.height}x${bitmap.width}]"
+            )
+        }
+        val actualPixels = actualBitmap.getPixelsSnapshot()
+        val shouldBePixels = bitmap.getPixelsSnapshot()
+
+        shouldBePixels.forEachIndexed { index, shouldBePixel ->
+            val actualPixel = actualPixels[index]
+            if (!hasSameColor(shouldBePixel, actualPixel)) {
+                throw AssertionFailedError("Bitmaps are different")
```
int02h: Maybe it is possible to calculate (x, y) position of the first different pixel so it will be much easier to a developer to find the difference between images in any graphical editor.

subprojects/android-test/ui-testing-core/src/main/kotlin/com/avito/android/test/matcher/BitmapMatcher.kt
```diff
@@ -42,10 +52,10 @@ class BitmapMatcher(private val bitmap: Bitmap) :
         return r < CONFIDENCE_INTERVAL && g < CONFIDENCE_INTERVAL && b < CONFIDENCE_INTERVAL
     }
 
-    private fun Bitmap.getTopLeftPixels(): List<Int> {
+    private fun Bitmap.getPixelsSnapshot(): List<Int> {
         val result = mutableListOf<Int>()
-        for (i in 5..30 step 5) {
-            for (j in 5..30 step 5) {
+        for (i in 1 until this.width step 5) {
```
int02h: Why do we do a step of 5 pixels? It is too expensive to fairly check all pixels of images? The most of images have 32x32 pixel size. It is only 1024 int value to check. Seems not too much.

subprojects/android-test/ui-testing-core/src/main/kotlin/com/avito/android/test/matcher/BitmapMatcher.kt
eugene-krivobokov: Agree, will add information about pixels

subprojects/android-test/ui-testing-core/src/main/kotlin/com/avito/android/test/matcher/BitmapMatcher.kt
eugene-krivobokov: Didn't want to change previous logic, cause I don't know the motivation for it.
Agree, will compare all the pixels.

subprojects/android-test/ui-testing-core/src/main/kotlin/com/avito/android/test/matcher/BitmapMatcher.kt
eugene-krivobokov: Fixed

subprojects/android-test/ui-testing-core/src/main/kotlin/com/avito/android/test/matcher/BitmapMatcher.kt
eugene-krivobokov: Fixed

buildscript/src/main/kotlin/Dependencies.kt
```diff
@@ -60,8 +60,7 @@ object Dependencies {
     // 1.6.x <-> AGP 3.6.x
     // 2.0.x <-> AGP 4.0.x
     // 2.1.x <-> AGP 4.1.x
-    // TODO: MBS-9089 2.0.94
-    val r8 = "com.android.tools:r8:1.6.89"
+    val r8 = "com.android.tools:r8:2.0.94"
```
eugene-krivobokov, line 63: Had to revert #530. The problem is relevant to local builds in IDEA 2020.2. In AS 4.x it works fine. 
From this point, see no reason to use an unsupported r8 version.

eugene-krivobokov:
:pray:

eugene-krivobokov:
ping

int02h (approved):

PR #541, opened 2020-08-06
MBS-9197 Clarify error messages for nested steps

eugene-krivobokov:


subprojects/android-test/test-report/src/test/kotlin/com/avito/android/test/report/NestedStepsTest.kt
```diff
@@ -0,0 +1,60 @@
+package com.avito.android.test.report
+
+import com.google.common.truth.Truth.assertThat
+import org.junit.jupiter.api.Assertions.assertThrows
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.extension.RegisterExtension
+
+class NestedStepsTest {
```
eugene-krivobokov, line 9: Renamed from "internal" to "nested".
It seems for me that "nested" describes relationships between steps more precisely. They are equal but only organized in this way.

int02h (approved):

PR #550, opened 2020-08-16
MBS-9161 update kotlin to 1.4

dsvoronin:
https://github.com/JetBrains/kotlin/releases/tag/v1.4.0


eugene-krivobokov (approved):

dsvoronin:
We can't merge it in current state, because some functions compiled in 1.4.0 are not compatible with latest gradle versions (6.6 comes with bundled 1.3.72). It causes runtime error `'void kotlin.jvm.internal.PropertyReference1Impl.<init>(java.lang.Class, java.lang.String, java.lang.String, int)'`

PR #554, opened 2020-08-17
MBS-9283 Delete unused Logger.exception

sergioboi:


sergioboi:
Avito deletion http://links.k.avito.ru/Mj

int02h (approved):

PR #560, opened 2020-08-24
MBS-9333 deprecate performance plugin

dsvoronin:
we decided to stop investing in blocking performance tests for now


subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/TestsFilter.kt
```diff
@@ -27,9 +27,11 @@ private val regressionNoE2EAnnotations = uiNoE2EAnnotations + manualAnnotations
 
 private val regressionAnnotations = uiAnnotations + manualAnnotations
 
+@Deprecated("remove after 2020.18")
```
sergioboi, line 30: What clients do need this?

sergioboi (approved):

PR #562, opened 2020-08-26
MBS-9315 impact analysis false negative fix

dsvoronin:
fix false negative case of parent gradle module, unsupported by impact analysis, but affecting child modules


dsvoronin:
There is another tricky case, not covered here:

```
-android module
---child module (not dependent on android one at all)
```

changes in android module's build.gradle won't trigger child module branch checks
it's completely legal in terms of gradle
i find this case kinda exotic and prefer not to worry about it until we'll find a reproducible real world sample with motivation behind it

sergioboi (approved):

PR #565, opened 2020-08-31
Fix androidTest configuration affecting by test changes

sergioboi:


subprojects/gradle/impact-shared/src/main/java/com/avito/impact/configuration/AndroidTestConfiguration.kt
```diff
@@ -9,14 +9,14 @@ class AndroidTestConfiguration(module: InternalModule) : SimpleConfiguration(mod
 
     override val isModified: Boolean by lazy {
         dependencies.any { it.isModified }
-            || module.testConfiguration.isModified
+            || module.implementationConfiguration.isModified
```
sergioboi, line 12: `androidTest` doesn't depend on `test` sources.
So If `test` sources are changed then `androidTest` will not be affected

eugene-krivobokov (approved):

PR #571, opened 2020-09-10
MBS-9422 leftovers for TMS plugin

dsvoronin:
- failfast at configuration time if no reportHost provided
- rewrite log messages to clearly understand state

### DIFF
--- a/subprojects/gradle/tms/src/main/kotlin/com/avito/plugin/MarkReportAsSourceAction.kt
+++ b/subprojects/gradle/tms/src/main/kotlin/com/avito/plugin/MarkReportAsSourceAction.kt
@@ -19,42 +19,46 @@ class MarkReportAsSourceAction(
     private val timeProvider: TimeProvider,
     private val logger: Logger
 ) {
+    private val tag = "TMS"
 
     fun mark(reportCoordinates: ReportCoordinates) {
         val testSuiteVersion = timeProvider.nowInMillis()
 
-        logger.debug("This is a new version [$testSuiteVersion] of full test suite for tms")
+        when (val result = reportsApi.getReport(reportCoordinates)) {
+            is GetReportResult.Found -> {
 
-        val reportId = tryGetId(reportCoordinates)
-
-        if (reportId != null) {
-            reportsApi.pushPreparedData(
-                reportId = reportId,
-                analyzerKey = "test_suite",
-                preparedData = jsonObject(
-                    "full" to true,
-                    "version" to testSuiteVersion
-                )
-            ).onFailure { error ->
-                logger.critical("Can't push prepared data: testSuite info", error)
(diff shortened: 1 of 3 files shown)


subprojects/gradle/tms/src/main/kotlin/com/avito/plugin/MarkReportAsSourceAction.kt
```diff
@@ -19,42 +19,46 @@ class MarkReportAsSourceAction(
...
+                    preparedData = jsonObject(
+                        "full" to true,
+                        "version" to testSuiteVersion
+                    )
+                ).fold(
+                    {
+                        reportsApi.setFinished(reportCoordinates).fold(
+                            {
+                                logger.info(
+                                    "[$tag] Test suite for tms version $testSuiteVersion, " +
```
dsvoronin, line 42: do we need tags like that? it's not straightforward to me how to filter build log for TMS messages only

eugene-krivobokov: CILogger already appends a prefix. Isn't that enough?

eugene-krivobokov (approved):

subprojects/gradle/tms/src/main/kotlin/com/avito/plugin/MarkReportAsSourceAction.kt
```diff
@@ -19,42 +19,46 @@ class MarkReportAsSourceAction(
...
-
-    private fun tryGetId(reportCoordinates: ReportCoordinates): String? {
-        return when (val result = reportsApi.getReport(reportCoordinates)) {
-            is GetReportResult.Found -> result.report.id
             GetReportResult.NotFound -> {
-                logger.warn("Can't find report for runId=${reportCoordinates.runId}")
-                null
+                logger.critical(
+                    "[$tag] Can't get reportId for coordinates: $reportCoordinates",
+                    IllegalStateException("stub")
```
sergioboi, line 57: Add message

subprojects/gradle/tms/src/main/kotlin/com/avito/plugin/MarkReportAsSourceAction.kt
dsvoronin: stdout destination doesn't print any tag as i see; we should handle it; but not in this PR

subprojects/gradle/tms/src/main/kotlin/com/avito/plugin/MarkReportAsSourceAction.kt
dsvoronin: i want optional Throwable arg here, or i need stub exceptions like here

PR #579, opened 2020-09-21
MBS-9441 impact analysis filters fix

dsvoronin:
added and modified tests was added to affected test suite run by mistake

- [x] tests

### DIFF
--- a/subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/suite/filter/FilterFactoryImpl.kt
+++ b/subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/suite/filter/FilterFactoryImpl.kt
@@ -1,11 +1,11 @@
 package com.avito.instrumentation.suite.filter
 
+import com.avito.instrumentation.configuration.ImpactAnalysisPolicy
 import com.avito.instrumentation.configuration.InstrumentationFilter
 import com.avito.instrumentation.configuration.InstrumentationFilter.FromRunHistory
 import com.avito.instrumentation.report.Report
 import com.avito.instrumentation.suite.filter.FilterFactory.Companion.JUNIT_IGNORE_ANNOTATION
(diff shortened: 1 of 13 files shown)


subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/suite/filter/FilterFactoryImpl.kt
```diff
@@ -169,12 +169,30 @@ private class FilterFactoryImpl(
...
-        if (impactAnalysisResult != null && impactAnalysisResult.exists()) {
-            val testNames = impactAnalysisResult.readLines()
+        return when (impactAnalysisResult.policy) {
+            is ImpactAnalysisPolicy.Off -> {
+                // do nothing
+            }
+            is ImpactAnalysisPolicy.On.RunAffectedTests -> {
+                addImpactTests(impactAnalysisResult.affectedTests)
+                removeImpactTests(impactAnalysisResult.addedTests)
+                removeImpactTests(impactAnalysisResult.modifiedTests)
```
eugene-krivobokov, line 179: Why did you choose that this removing is the responsibility of filters?
Let's imagine multiple clients of impact analysis' results. Whill they duplicate this logic?

dsvoronin: that is good catch; i didn't think of it that way
let me change it

eugene-krivobokov (approved):

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/suite/filter/FilterFactoryImpl.kt
dsvoronin: Decided to improve this part later

PR #582, opened 2020-09-30
MBS-9441 fix impact analysis filters

dsvoronin:


subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/InstrumentationTestsAction.kt
```diff
@@ -40,6 +40,7 @@ class InstrumentationTestsAction(
 
     override fun run() {
         logger.debug("Starting instrumentation tests action for configuration: ${params.instrumentationConfiguration}")
+        logger.info("impact analysis = ${params.impactAnalysisResult}")
```
sergioboi: debug
Maybe better write this to the file system

sergioboi (approved):

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/InstrumentationTestsAction.kt
dsvoronin: debug doing exactly that; done

PR #586, opened 2020-10-01
Add kotlin-android, kapt plugins to the TestProjectGenerator.kt

sergioboi:

PR #588, opened 2020-10-01
Add BuildVerdictPlugin.kt

sergioboi:
Capture
- kotlin compilation logs
- kapt logs
- unit tests logs


build.gradle.kts
```diff
@@ -27,6 +27,7 @@ plugins {
     id("com.android.application") apply false
     id("com.avito.android.instrumentation-tests") apply false
     id("com.avito.android.impact")
+    id("com.avito.android.build-verdict")
```
sergioboi: I don't get how to customize this block with `if(ci) id("build-verdict")

subprojects/gradle/build-verdict/src/main/kotlin/com/avito/android/build_verdict/TaskErrorOutputCaptureExecutionListener.kt
```diff
@@ -0,0 +1,46 @@
...
+import org.gradle.api.Task
+import org.gradle.api.execution.TaskExecutionListener
+import org.gradle.api.tasks.TaskState
+import org.gradle.api.tasks.testing.Test
+import org.gradle.api.tasks.testing.TestDescriptor
+import org.gradle.api.tasks.testing.TestResult
+import org.jetbrains.kotlin.gradle.internal.KaptTask
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
+
+internal class TaskErrorOutputCaptureExecutionListener(
```
sergioboi, line 16: logic for getting error outputs

build.gradle.kts
dsvoronin: you simply don't )

subprojects/gradle/build-verdict/src/test/kotlin/com/avito/android/build_verdict/BuildVerdictPluginTest.kt
```diff
@@ -0,0 +1,194 @@
...
+import java.io.File
+
+class BuildVerdictPluginTest {
+
+    @field:TempDir
+    lateinit var temp: File
+    private val gson = GsonBuilder().create()
+
+    @Test
+    fun `When kotlin compilation fails Then build-verdict contains kotlin compile info`() {
```
dsvoronin: name should be on lower-case

subprojects/gradle/build-verdict/src/test/kotlin/com/avito/android/build_verdict/BuildVerdictPluginTest.kt
```diff
@@ -0,0 +1,194 @@
...
+
+        assertBuildVerdict(
+            failedTask = "compileDebugKotlin",
+            errorOutput = "e: /private$temp/app/src/main/kotlin/Uncompiled.kt: (1, 1): Expecting a top level declaration\n" +
+                "e: /private$temp/app/src/main/kotlin/Uncompiled.kt: (1, 11): Expecting a top level declaration"
+        )
+    }
+
+    @Test
+    fun `When kapt stubs generating fails Then build-verdict contains kapt info`() {
```
dsvoronin: `when` `what` `if` blocks should be separated with ` - ` for consistency in other tests

subprojects/gradle/build-verdict/src/test/kotlin/com/avito/android/build_verdict/BuildVerdictPluginTest.kt
```diff
@@ -0,0 +1,194 @@
...
+
+        assertBuildVerdict(
+            failedTask = "testDebugUnitTest",
+            errorOutput = """FAILED tests:
+	AppTest.test assert true
+	AppTest.test runtime exception"""
+        )
+    }
+
+    fun `When lint fails Then build-verdict contains lint info`() {
```
dsvoronin: better mark these stubs with TODO, or even better delete it
i'm still finding your empty stubs(written year ago) in instrumentation plugin :)

https://github.com/avito-tech/avito-android/blob/develop/subprojects/gradle/instrumentation-tests/src/test/kotlin/com/avito/instrumentation/InstrumentationTestActionFinalizerTest.kt

subprojects/gradle/build-verdict/src/test/kotlin/com/avito/android/build_verdict/BuildVerdictPluginTest.kt
```diff
@@ -0,0 +1,194 @@
...
+    ) {
+        val buildVerdict = File(temp, "outputs/build-verdict/$buildVerdictFileName")
+
+        Truth.assertWithMessage("outputs/build-verdict/$buildVerdictFileName exist")
+            .that(buildVerdict.exists())
+            .isTrue()
+
+        val actualBuildVerdict = gson.fromJson(buildVerdict.readText(), BuildVerdict::class.java)
+
+        Truth.assertThat(actualBuildVerdict.failedTasks)
```
dsvoronin: static import `assertThat` for consistency

build.gradle.kts
```diff
@@ -27,6 +27,7 @@ plugins {
     id("com.android.application") apply false
     id("com.avito.android.instrumentation-tests") apply false
     id("com.avito.android.impact")
+    // if(ci) id("com.avito.android.build-verdict")
```
dsvoronin: you should not conditionally apply plugin; it should be inside plugin's logic

subprojects/gradle/build-verdict/build.gradle.kts
```diff
@@ -0,0 +1,28 @@
...
+    id("java-gradle-plugin")
+    `maven-publish`
+    id("com.jfrog.bintray")
+}
+
+dependencies {
+    implementation(gradleApi())
+    implementation(Dependencies.gradle.kotlinPlugin)
+    implementation(Dependencies.gradle.androidPlugin) {
+        because("without error occurs in tests")
```
dsvoronin: you can remove this line now? after #586 merged?

subprojects/gradle/build-verdict/src/main/kotlin/com/avito/android/build_verdict/BuildVerdict.kt
```diff
@@ -0,0 +1,18 @@
+package com.avito.android.build_verdict
+
+internal data class GsonableError(
```
dsvoronin: why `gson` in naming? i see nothing about concrete json serialization library in it's signature

i see no need in this class actually, you can pass classic `Throwable` and use function to serialize stacktrace inside serialization function

subprojects/gradle/build-verdict/src/main/kotlin/com/avito/android/build_verdict/BuildVerdictPlugin.kt
```diff
@@ -0,0 +1,36 @@
...
+
+internal data class TaskPath(val value: String)
+
+class BuildVerdictPlugin : Plugin<ProjectInternal> {
+
+    override fun apply(target: ProjectInternal) {
+        require(target.isRoot()) {
+            "build-verdict plugin must be applied to the root project"
+        }
+        val logs = mutableMapOf<TaskPath, StringBuilder>()
```
dsvoronin: i'd wrap it to some `LogCapturer`, with interface, to easily change implementation, if this data structure show itself uncapable

subprojects/gradle/build-verdict/src/main/kotlin/com/avito/android/build_verdict/BuildVerdictPlugin.kt
```diff
@@ -0,0 +1,36 @@
+package com.avito.android.build_verdict
+
+import com.avito.kotlin.dsl.isRoot
+import com.avito.utils.logging.ciLogger
+import com.google.gson.GsonBuilder
+import org.gradle.api.Plugin
+import org.gradle.api.internal.project.ProjectInternal
+import org.gradle.kotlin.dsl.create
+
+internal data class TaskPath(val value: String)
```
dsvoronin, line 12: i believe gradle has a built-in public class for that

yep: https://github.com/gradle/gradle/blob/master/subprojects/base-services/src/main/java/org/gradle/util/Path.java

subprojects/gradle/build-verdict/src/main/kotlin/com/avito/android/build_verdict/BuildVerdictPluginExtension.kt
```diff
@@ -0,0 +1,10 @@
+package com.avito.android.build_verdict
+
+import org.gradle.api.file.Directory
+import org.gradle.api.file.ProjectLayout
+
+abstract class BuildVerdictPluginExtension(
+    projectLayout: ProjectLayout
+) {
+    val buildVerdictOutputDir: Directory = projectLayout.projectDirectory.dir("outputs/build-verdict")
```
dsvoronin: extension, but with val? why?

subprojects/gradle/build-verdict/src/main/kotlin/com/avito/android/build_verdict/BuildFailureListener.kt
```diff
@@ -0,0 +1,73 @@
...
+    override fun execute(result: BuildResult) {
+        result.failure?.apply {
+            onFailure(this)
+        }
+    }
+
+    private fun onFailure(failure: Throwable) {
+        val failedTasks = graph.allTasks
+            .filter { it.state.failure != null }
+        val verdict = gson.toJson(
```
dsvoronin: i want to see `Writer` extracted as separate abstraction;
this way `private val gson: Gson` should not be in `BuildFailureListener` constructor (it seemed weird for me as start, and because of hardcoded writer here)

subprojects/gradle/build-verdict/src/main/kotlin/com/avito/android/build_verdict/DefaultTestListener.kt
```diff
@@ -0,0 +1,23 @@
+package com.avito.android.build_verdict
+
+import org.gradle.api.tasks.testing.TestDescriptor
+import org.gradle.api.tasks.testing.TestListener
+import org.gradle.api.tasks.testing.TestResult
+
+abstract class DefaultTestListener : TestListener {
```
dsvoronin, line 7: Do you need this now, or better in separate PR about tests?

subprojects/gradle/build-verdict/src/test/kotlin/com/avito/android/build_verdict/BuildVerdictPluginTest.kt
```diff
@@ -0,0 +1,194 @@
...
+import com.avito.test.gradle.gradlew
+import com.avito.test.gradle.kotlinClass
+import com.avito.test.gradle.module.AndroidAppModule
+import com.google.common.truth.Truth
+import com.google.gson.GsonBuilder
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.io.TempDir
+import java.io.File
+
+class BuildVerdictPluginTest {
```
dsvoronin, line 17: i vote for another positive scenario test here (build success -> verdict empty? or not created at all?)

build.gradle.kts
sergioboi: Let's discuss this. For me is a clear intention I like a user want to decide when I want to apply plugin

subprojects/gradle/build-verdict/build.gradle.kts
sergioboi: #586 fixes not all scenarios. For example, if you depend on `kotlin-android` it will work:
- when your module doesn't have any of kotlin, android plugins
- when your module has both kotlin and android plugins

subprojects/gradle/build-verdict/src/main/kotlin/com/avito/android/build_verdict/BuildVerdict.kt
sergioboi: Throwable can't be serialized to the JSON. I see two options to fix:
- Create a wrapper object
- Register adapter
I prefer a wrapper because it is closer to the domain and I want to decide what data I will extract from Throwable

subprojects/gradle/build-verdict/src/main/kotlin/com/avito/android/build_verdict/DefaultTestListener.kt
sergioboi: How does this class complicate this PR?

subprojects/gradle/build-verdict/src/main/kotlin/com/avito/android/build_verdict/BuildVerdictPluginExtension.kt
sergioboi: Change. I thought that the Directory is Property

subprojects/gradle/build-verdict/src/main/kotlin/com/avito/android/build_verdict/BuildVerdictPlugin.kt
sergioboi: I don't see problems with changing implementation for internal field
But I could wrap this for you

subprojects/gradle/build-verdict/src/test/kotlin/com/avito/android/build_verdict/BuildVerdictPluginTest.kt
sergioboi: We have already discussed that topic and I didn't get why this pattern is bad
I could change the naming for you but if you want me to make this in the future I propose to create documentation about test naming with motivation and discussed all pros and cons there

subprojects/gradle/build-verdict/src/test/kotlin/com/avito/android/build_verdict/BuildVerdictPluginTest.kt
sergioboi: Why? As for me, it's easier to read function when it has anchors for the eyes

subprojects/gradle/build-verdict/src/test/kotlin/com/avito/android/build_verdict/BuildVerdictPluginTest.kt
dsvoronin: it's just about consistency; we need single pattern for all tests in repository; if you're introducing new one - make sure it's changed across repo
i dont' mind one or another way, i'm just for single pattern, it's easier to read

subprojects/gradle/build-verdict/src/test/kotlin/com/avito/android/build_verdict/BuildVerdictPluginTest.kt
dsvoronin: it was in https://kotlinlang.org/docs/reference/coding-conventions.html#naming-rules
and we chose ` - ` for anchors

build.gradle.kts
dsvoronin: it's kinda gradle's design decision

https://discuss.gradle.org/t/how-to-apply-plugin-based-on-build-type-android/15265/2?u=jffiorillo

some part of script logic depends on it; like kotlin dsl won't compile if you're using generated accessors

subprojects/gradle/build-verdict/build.gradle.kts
dsvoronin: in this case, it will be better to clarify `because` message: 
something like: `TestProjectGenerator depends on agp in plugin's classpath` or even link to documentation about this case (better one)

subprojects/gradle/build-verdict/src/main/kotlin/com/avito/android/build_verdict/BuildVerdictPlugin.kt
dsvoronin: yep, it's not a problem in that perspective;

it's about design:
you pass implementation details around (mutable map), but imagine someone enabled `debug` logging level and we'll maybe encounter OOM's here while parsing logs, decided to change it to some buffered streaming;
i'd prefer to make an abstraction right away. backed by mutable map in naive implementation

and please, nothing here is "for me", i just sharing how i can improve this code :)

subprojects/gradle/build-verdict/src/main/kotlin/com/avito/android/build_verdict/DefaultTestListener.kt
dsvoronin: just the question of scope; like empty tests;
it's easy to change context after merge, decide about another implementation and forget about this class forever )

subprojects/gradle/build-verdict/src/main/kotlin/com/avito/android/build_verdict/BuildVerdict.kt
dsvoronin: yes i know that you need custom serialization; i prefer adapter here even if i don't like it's API, trigger for me is naming, if i can't justify name - something is wrong with abstraction 
so what about specific library in data class name? (why? nothing about gson here, it could be Jacksonable or Moshiable), leaked detail in naming

subprojects/gradle/build-verdict/src/main/kotlin/com/avito/android/build_verdict/BuildFailureListener.kt
```diff
@@ -0,0 +1,73 @@
...
+        )
+    }
+
+
+    private val Throwable.stringStackTrace: String
+        get() {
+            val stringWriter = StringWriter()
+            printStackTrace(PrintWriter(stringWriter))
+            return stringWriter.toString()
+        }
```
eugene-krivobokov: Duplicates StackTraceFormatter

subprojects/gradle/build-verdict/src/test/kotlin/com/avito/android/build_verdict/BuildVerdictPluginTest.kt
eugene-krivobokov: >  it's easier to read function when it has anchors for the eyes

The question is not about naming. It is about changing agreements.
In all other tests, we have the same convention. If we want to change it, this is fine. But it's better to do it explicitly and fix it somewhere.

build.gradle.kts
dsvoronin: it's not the better citation i found, not really about this case; but i seen better somewhere

subprojects/gradle/build-verdict/src/main/kotlin/com/avito/android/build_verdict/BuildVerdictPlugin.kt
```diff
@@ -0,0 +1,36 @@
...
+
+class BuildVerdictPlugin : Plugin<ProjectInternal> {
+
+    override fun apply(target: ProjectInternal) {
+        require(target.isRoot()) {
+            "build-verdict plugin must be applied to the root project"
+        }
+        val logs = mutableMapOf<TaskPath, StringBuilder>()
+        val extension = target.extensions.create<BuildVerdictPluginExtension>("buildVerdict")
+        target.gradle.addListener(TaskErrorOutputCaptureExecutionListener(logs))
```
eugene-krivobokov: It may cause a memory leak because of a daemon.
Probably, we should remove this listener.
See example in `GradleCollector.cleanup`.

Also, I know an option to reveal this problem.
Run multiple builds with a daemon. Collect memory profile and check multiple instances of listeners.
It can be done easier by gradle-profiler:
```
my_scenario {
    tasks = ["<any simple task>"]
    // To reveal memory-leaks
    warm-ups = 10
}
```

eugene-krivobokov: Also, there is an unobvious behavior of `buildFinished`. It's invoked multiple times with different results.
We had a memory leak after IDE synchronization because of it (MBS-5738).

eugene-krivobokov: This is why I think I'd good to adapt and reuse existing abstractions from the `gradle-profile` module.

subprojects/gradle/build-verdict/build.gradle.kts
sergioboi: http://links.k.avito.ru/mbs-9695

subprojects/gradle/build-verdict/src/main/kotlin/com/avito/android/build_verdict/BuildVerdict.kt
sboishtyan: Jsonable

sergioboi: rename to `Error` without adapters

subprojects/gradle/build-verdict/src/main/kotlin/com/avito/android/build_verdict/BuildVerdictPlugin.kt
sergioboi: > but imagine someone enabled debug logging level

We already capture only error level outputs, so we don't care about logging level

> nothing here is "for me"

I want to clarify my opinion that is cheaper to make "for you" then you could change my opinion in this design decision

sergioboi: I thought and didn't get how LogCapturer would differ from Map
It would need read-write methods like a MutableMap.
The current code already segregates read and write operation. So I don't see how LogCapturer will improve the code

subprojects/gradle/build-verdict/src/main/kotlin/com/avito/android/build_verdict/DefaultTestListener.kt
sergioboi: It's already used so I don't get how I can forget about this ))

subprojects/gradle/build-verdict/src/main/kotlin/com/avito/android/build_verdict/BuildFailureListener.kt
sergioboi: Extract a new module

subprojects/gradle/build-verdict/src/main/kotlin/com/avito/android/build_verdict/BuildVerdictPlugin.kt
sergioboi: I won't check if this is true. But I will add `removeListener` for both: taskOutputListner and buildFinishedAction
@eugene-krivobokov Take a look at the `buildFinishedAction` it seems to behave not like the same listener, but I didn't check

gradle.properties
```diff
@@ -13,3 +13,4 @@ avito.git.state=local
 # Without it failed on studio sync when ci=true. It happens because studio makes eager configuration of each register task.
 avito.repo.ssh.url=
 useCompositeBuild=true
+avito.build-verdict.enabled=false
```
eugene-krivobokov, line 16: What about an extension instead of properties?
It'll be more convenient for clients

subprojects/gradle/build-verdict/src/main/kotlin/com/avito/android/build_verdict/BuildVerdictPlugin.kt
eugene-krivobokov: Please, comment on decisions for resolved comments.
In this example, It's not clear why `org.gradle.util.Path` is not suitable.

subprojects/gradle/build-verdict/src/main/kotlin/com/avito/android/build_verdict/BuildVerdictPlugin.kt
```diff
@@ -0,0 +1,55 @@
...
+                    )
+                )
+            }
+        }
+    }
+
+    internal val Project.pluginIsEnabled: Boolean
+        get() = providers
+            .gradleProperty(enabledProp)
+            .forUseAtConfigurationTime()
```
eugene-krivobokov, line 48: :+1:

gradle.properties
eugene-krivobokov: Example:
```kotlin
buildVerdict {
   enabled = myCustomFunction()
}
```

sergioboi: Agree with that in common cases extension is preferable then property, but for the current case, the extension is an overhead. Only we and only once need to enable it on CI and disable for local purposes. So property will be enough

subprojects/gradle/build-verdict/src/main/kotlin/com/avito/android/build_verdict/BuildVerdictPluginExtension.kt
```diff
@@ -0,0 +1,15 @@
...
+import org.gradle.api.file.ProjectLayout
+import org.gradle.api.model.ObjectFactory
+import org.gradle.api.provider.Property
+
+@Suppress("UnstableApiUsage")
+abstract class BuildVerdictPluginExtension(
+    objects: ObjectFactory,
+    layout: ProjectLayout
+) {
+    val buildVerdictOutputDir: Property<Directory> =
```
eugene-krivobokov: How about just `outputDir`?

```kotlin
buildVerdict {
   outputDir = "custom/dir"
}

buildVerdict {
   buildVerdictOutputDir = "custom/dir"
}
```

subprojects/gradle/build-verdict/src/test/kotlin/com/avito/android/build_verdict/BuildVerdictPluginTest.kt
```diff
@@ -0,0 +1,231 @@
...
+            errorOutput = listOf(
+                """FAILED tests:
+	AppTest.test assert true
+	AppTest.test runtime exception"""
+            )
+        )
+    }
+
+    @Test
+    fun `build success - no verdict file`() {
```
eugene-krivobokov, line 154: In the future, it may be convenient to have a verdict with general information. At least for the sake of consistency.

subprojects/settings.gradle.kts
```diff
@@ -86,6 +87,7 @@ include(":common:junit-utils")
 include(":common:retrace")
 include(":common:truth-extensions")
 include(":common:composite-exception")
+include(":common:throwable-utils")
```
eugene-krivobokov, line 90: As for me, It's too little for a standalone module. It seems like "common:lang"

eugene-krivobokov (approved):

subprojects/common/throwable-utils/src/main/kotlin/com/avito/android/util/StackTraceFormatter.kt
```diff
@@ -4,9 +4,9 @@ import org.apache.commons.text.TextStringBuilder
 import java.io.PrintWriter
 
 /**
- * Prepare stacktrace for report
+ * Get stacktrace as list of lines from throwable
```
eugene-krivobokov, line 7: In Kotlin 1.4 we can use `Throwable.stackTraceToString()` instead of this function.

PR #604, opened 2020-10-20
MBS-8523 Create new abstraction DeviceProvider to hide all device creation and reservation stuff

sergioboi:


subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/reservation/devices/provider/DevicesProviderFactory.kt
```diff
@@ -0,0 +1,77 @@
...
+        ): DevicesProvider {
+            val androidDebugBridge = AndroidDebugBridge(
+              logger = { logger.info(it) }
+            )
+            val emulatorsLogsReporter = EmulatorsLogsReporter(
+              outputFolder = output,
+              logcatTags = executionParameters.logcatTags,
+              logcatDir = logcatDir
+            )
+            val devicesManages = AdbDevicesManager(logger = commonLogger(logger))
```
eugene-krivobokov: a typo - manager

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/reservation/devices/provider/KubernetesDevicesProvider.kt
```diff
@@ -0,0 +1,29 @@
...
+  private val adbDevicesManager: AdbDevicesManager,
+  private val logger: CILogger
+) : DevicesProvider {
+
+    override suspend fun provideFor(reservations: Collection<Reservation.Data>): ReceiveChannel<Device> {
+        val claim = client.claim(reservations)
+        // TODO parallel device getting
+        return claim.serials.map { serial ->
+            adbDevicesManager.findDevice(serial)
+                .orElseGet { throw IllegalStateException("Can't find device connected adb device $serial") }
```
eugene-krivobokov, line 23: What will it fail? Only the one reading from the channel?

eugene-krivobokov (approved):

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/reservation/devices/provider/KubernetesDevicesProvider.kt
sergioboi: This code is moved from deleted class `ExternalIntentionBasedDevicesObserver`

sergioboi: It will break the whole task

eugene-krivobokov: > It will break the whole task

This is strange. If one emulator hangs why the whole task should fail? 
It looks like we should restart this one like from DeviceWorker

PR #605, opened 2020-10-21
MBS-8523 Delete GlobalScope in TestRunner.kt

sergioboi:
Delete GlobalScope
Now runner tests are fairer: they can find leaks

### DIFF
--- a/subprojects/gradle/runner/client/src/main/kotlin/com/avito/runner/scheduler/runner/client/TestExecutionClient.kt
+++ b/subprojects/gradle/runner/client/src/main/kotlin/com/avito/runner/scheduler/runner/client/TestExecutionClient.kt
@@ -5,7 +5,8 @@ import com.avito.runner.scheduler.runner.client.model.ClientTestRunResult
 import com.avito.runner.scheduler.runner.scheduler.TestExecutionState
 import com.avito.runner.service.IntentionExecutionService
 import com.avito.runner.service.model.intention.Intention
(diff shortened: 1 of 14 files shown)


subprojects/gradle/runner/client/src/main/kotlin/com/avito/runner/scheduler/runner/client/TestExecutionClient.kt
```diff
@@ -18,27 +19,30 @@ class TestExecutionClient {
...
-        // TODO: Don't use global scope. Unconfined coroutines lead to leaks
-        GlobalScope.launch {
-            for (request in requests) {
-                statesMapping[request.intention] = request.state
-                executionServiceCommunication.intentions.send(request.intention)
+    suspend fun start(
+        executionServiceCommunication: IntentionExecutionService.Communication,
+        scope: CoroutineScope
+    ): Communication {
+        scope.launch(Dispatchers.Unconfined) {
```
eugene-krivobokov: Does it matter what to use when we only work with channels?

eugene-krivobokov (approved):

subprojects/gradle/runner/client/src/main/kotlin/com/avito/runner/scheduler/runner/client/TestExecutionClient.kt
sergioboi: You are right. I pretend that I understand coroutines

PR #606, opened 2020-10-23
Delete pod in k8s when DeviceWorker dies

sergioboi:


buildscript/src/main/kotlin/Dependencies.kt
```diff
@@ -11,6 +11,7 @@ object Dependencies {
         val espresso = "3.2.0"
         val mockito = "3.3.3"
         val detekt = "1.10.0"
+        val coroutines = "1.3.7"
```
dsvoronin: I feel that version upgrade should be in separate commit (PR better, because of squash);
at least mention it in resulting commit message (+PR title)

subprojects/gradle/runner/service/src/main/kotlin/com/avito/runner/service/worker/device/adb/Adb.kt
```diff
@@ -0,0 +1,16 @@
+package com.avito.runner.service.worker.device.adb
+
+class Adb {
+    val adb: String
```
dsvoronin: It looks like `adbBinaryPath` suits better

subprojects/gradle/runner/service/src/main/kotlin/com/avito/runner/service/worker/device/adb/Adb.kt
```diff
@@ -0,0 +1,16 @@
...
+
+    init {
+        val androidHome: String = requireNotNull(System.getenv("ANDROID_HOME")) {
```
dsvoronin, line 8: Please consider using [AndroidSdk.kt](https://github.com/avito-tech/avito-android/blob/08d8bf607c449683b0b8e56bd43da386c1bae45c/subprojects/gradle/android/src/main/kotlin/com/avito/android/AndroidSdk.kt#L27) as a single place for sdk binaries access

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/reservation/devices/provider/KubernetesDevicesProvider.kt
```diff
@@ -3,27 +3,44 @@ package com.avito.instrumentation.reservation.devices.provider
...
 
     override fun provideFor(reservations: Collection<Reservation.Data>, scope: CoroutineScope): ReceiveChannel<Device> {
         val claim = client.claim(reservations, scope)
         // TODO parallel device getting
-        return claim.serials.map { serial ->
-            adbDevicesManager.findDevice(serial)
-                .orElseGet { throw IllegalStateException("Can't find device connected adb device $serial") }
+        return claim.deviceCoordinates.map { coordinate ->
+            val adbDeviceParams = adbDevicesManager.findDevice(coordinate.serial)
+                .orElseGet { throw IllegalStateException("Can't find device connected adb device ${coordinate.serial}") }
```
dsvoronin, line 28: How do we know about this exception? Will it fail the whole build?

dsvoronin (approved):

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/reservation/devices/provider/KubernetesDevicesProvider.kt
sergioboi: Yes, it should fail the whole build now

subprojects/gradle/runner/service/src/main/kotlin/com/avito/runner/service/worker/device/adb/Adb.kt
sergioboi: It is hard to reuse AndroidSdk.kt. I add a todo for further refactorings

eugene-krivobokov: It seems that this knowledge should be passed to runner/service from outside.

PR #612, opened 2020-10-31
MBS-7970 remove :gradle:utils module

dsvoronin:


eugene-krivobokov (approved):

subprojects/gradle/signer/src/main/kotlin/com/avito/plugin/SignServicePlugin.kt
```diff
@@ -198,3 +199,14 @@ class SignServicePlugin : Plugin<Project> {
         }
     }
 }
+
+@OptIn(ExperimentalContracts::class)
+private fun String?.hasContent(): Boolean {
```
sergioboi, line 204: Hello strange function)

PR #624, opened 2020-11-10
MBS-9781 register separate LintReportToSlackTask per slack channel

dsvoronin:


subprojects/gradle/lint-report/src/main/kotlin/com/avito/android/lint/slack/LintReportToSlackTaskFactory.kt
```diff
@@ -0,0 +1,63 @@
...
+                token = token,
+                workspace = workspace
+            ) as SlackClient
+        }
+    }
+
+    /**
+     * To be used in CiStep, because slack channel only known from there
+     */
+    fun registerLintReportToSlackTask(slackChannel: String): TaskProvider<LintSlackReportTask> {
```
eugene-krivobokov, line 37: Explain please, why don't we just send the same report from one task.
To make each interaction independent (loggable, cacheable, ...)?

eugene-krivobokov (approved):

subprojects/gradle/lint-report/src/main/kotlin/com/avito/android/lint/slack/LintReportToSlackTaskFactory.kt
dsvoronin: previously we had single task in graph for main app, like `:app:lintReportToSlack` it was configured multiple times from different `builds` in CiSteps plugin;

`fullCheck` configures and sets input property `slackChannel` in `:app:lintReportToSlack` to `#android-dev`
`regress` configures and sets input property `slackChannel` in `:app:lintReportToSlack` to `""`

now `fullCheck` configuration is broken and property set to `""`

eugene-krivobokov: Initially, I thought about multiple destinations in one build.
Thanks, I see the conflict between different builds.
It seems we need to add configuration tests to address this problem in all steps.

subprojects/gradle/lint-report/src/main/kotlin/com/avito/android/lint/slack/LintReportToSlackTaskFactory.kt
```diff
@@ -0,0 +1,63 @@
...
+
+        val taskName = "lintReportTo${slackChannel.validInGradleTaskName()}"
+
+        var taskProvider = project.tasks.typedNamedOrNull<LintSlackReportTask>(taskName)
+
+        if (taskProvider == null) {
+            logger.info("LintCheck: task $taskName already created in another ciStep; multiple reports are possible")
```
sergioboi, line 44: I don't understand the message because `taskProvider` is null but the message says `taskName already created`

PR #626, opened 2020-11-11
Hotfix: Return deployment deletion

sergioboi:


eugene-krivobokov (approved):

dsvoronin (approved):

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/reservation/client/kubernetes/KubernetesReservationClient.kt
```diff
@@ -163,6 +163,12 @@ class KubernetesReservationClient(
                             }
                         }
                     }
+
+                    logger.debug("Deleting deployment: $deploymentName...")
+                    removeEmulatorsDeployment(
+                        deploymentName = deploymentName
+                    )
+                    logger.debug("Deployment: $deploymentName deleted")
                 }
```
eugene-krivobokov, line 172: Can we cover it by integration tests?

sergioboi: We use `kubernetesClient` for deletion and there is no result of this, so I see one opportunity for testing it, only by wrap `kubernetesClient` by an interface and test it's fake/mock/stub

eugene-krivobokov: Worth to try: https://github.com/fabric8io/kubernetes-client#mocking-kubernetes

PR #629, opened 2020-11-11
MBS-9086 First preparations for AGP 4.1

eugene-krivobokov:
Minor preparations for AGP 4.1 before breaking changes.

### DIFF
--- a/subprojects/gradle/impact-shared/src/main/java/com/avito/impact/configuration/SimpleConfiguration.kt
+++ b/subprojects/gradle/impact-shared/src/main/java/com/avito/impact/configuration/SimpleConfiguration.kt
@@ -1,10 +1,13 @@
 package com.avito.impact.configuration
 
 import com.android.build.gradle.api.AndroidSourceSet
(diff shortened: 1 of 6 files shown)


subprojects/gradle/impact-shared/src/main/java/com/avito/impact/configuration/SimpleConfiguration.kt
```diff
@@ -79,3 +83,22 @@ abstract class SimpleConfiguration(val module: InternalModule) : Equality {
...
+                .withType(DefaultProjectDependency::class.java)
+        }
+        .toSet()
+        .map {
+            it.dependencyProject
+                .internalModule
+                .implementationConfiguration
+        }
+        .toMutableSet().apply {
+            remove(this@dependencies)
```
eugene-krivobokov, line 102: Somehow, in AGP 4.1 it founds configuration itself. Whatever the reason, we don't need it in dependencies.

dsvoronin (approved):

PR #633, opened 2020-11-14
Mbs 9937 lintlink

dsvoronin:

PR #634, opened 2020-11-14
Mbs 9936 lint crashes in report

dsvoronin:
[breaking] have to set `slackChannelToReportLintBugs` property 

waiting for #633 as based on its branch

### DIFF
--- a/subprojects/gradle/lint-report/src/main/kotlin/com/avito/android/lint/slack/LintSlackReportTask.kt
+++ b/subprojects/gradle/lint-report/src/main/kotlin/com/avito/android/lint/slack/LintSlackReportTask.kt
@@ -42,7 +42,7 @@ abstract class LintSlackReportTask : DefaultTask() {
 
         createLintSlackAlert().report(
             lintReport = models,
-            channelForDevs = SlackChannel(slackReportChannel.get()),
+            channel = SlackChannel(slackReportChannel.get()),
             channelForLintBugs = SlackChannel(slackChannelForLintBugs.get()),
             buildUrl = teamcityBuildLinkAccessor.getBuildUrl()
         )
--- a/subprojects/gradle/lint-report/src/main/kotlin/com/avito/android/lint/slack/LintSlackReportTask.kt
+++ b/subprojects/gradle/lint-report/src/main/kotlin/com/avito/android/lint/slack/LintSlackReportTask.kt
@@ -18,6 +18,9 @@ abstract class LintSlackReportTask : DefaultTask() {
     @get:Input
     abstract val slackReportChannel: Property<String>
 
+    @get:Input
+    abstract val slackChannelForLintBugs: Property<String>
+
     @get:InputFile
     abstract val lintXml: RegularFileProperty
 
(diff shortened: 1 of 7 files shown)


subprojects/gradle/lint-report/src/main/kotlin/com/avito/android/lint/slack/LintSlackReportTask.kt
```diff
@@ -39,7 +42,8 @@ abstract class LintSlackReportTask : DefaultTask() {
 
         createLintSlackAlert().report(
             lintReport = models,
-            channel = SlackChannel(slackReportChannel.get()),
+            channelForDevs = SlackChannel(slackReportChannel.get()),
```
eugene-krivobokov: `ForDevs` - is a use case.
It's better to avoid such details in the plugin and focus on the content.

dsvoronin: i agree, fixed

subprojects/gradle/lint-report/src/main/kotlin/com/avito/android/lint/slack/LintSlackReporter.kt
```diff
@@ -24,44 +26,113 @@ interface LintSlackReporter {
...
+                    val (fatalErrors, regularErrors) = errors.partition { it.isFatal }
+
+                    var isMessageSent = false
+
+                    if (regularErrors.isNotEmpty()) {
+                        sendReport(
+                            channel = channelForDevs,
+                            message = buildSlackMessageToDevs(
+                                projectPath = lintReport.projectRelativePath,
+                                errors = regularErrors,
```
eugene-krivobokov, line 46: In this case, we hide some of the errors. 
I'm afraid it can be misleading to see different information in a message and in a full report.
I propose to add information about fatals but mark them.

subprojects/gradle/lint-report/src/main/kotlin/com/avito/android/lint/slack/LintSlackReporter.kt
```diff
@@ -25,43 +27,112 @@ interface LintSlackReporter {
...
+                            message = buildSlackMessageAboutLintBugs(
+                                projectPath = lintReport.projectRelativePath,
+                                fatalErrors = fatalErrors,
+                                unknownErrors = unknowns,
+                                buildUrl = buildUrl
+                            ),
+                            htmlReport = lintReport.htmlFile
+                        )
+
+                        isMessageSent = true
```
eugene-krivobokov, line 68: Why do we need this state?
Can we avoid it or store it somewhere else (task output, check by API in slack, ...)?

subprojects/gradle/lint-report/src/main/kotlin/com/avito/android/lint/slack/LintSlackReporter.kt
dsvoronin: probably, let's see how it's going, it will be easy to fill another type of errors in separate line here

eugene-krivobokov (approved):

PR #632, opened 2020-11-12
MBS-9086 Basic AGP 4.1 support

eugene-krivobokov:
- Overview: https://medium.com/androiddevelopers/new-apis-in-the-android-gradle-plugin-f5325742e614
- Samples: https://github.com/android/gradle-recipes

New lifecycle in AGP tasks:

- Build scripts are run, allowing the build and plugins to configure the Android Gradle Plugin DSL objects as well as registering variant callbacks.   
- The Android Gradle Plugin combines build types and product flavors to create variants and test components.   
- The `onVariants` API is invoked for each variant, allowing the customization of variant settings.
     - NB: It doesn’t use properties since the values are used at configuration time.   
- The `onVariantProperties` API is invoked for each variant that was enabled allowing the registration of tasks that consume or modify intermediates of the build.
     - NB: Didn't find alternative for test variants. Asking for help: https://gradle-community.slack.com/archives/CJYS1DAP5/p1605196130059800
- The Android Gradle Plugin registers tasks for the variants after both the `onVariants` and the `onVariantProperties` callbacks have been called.   
- The previous variant API (`applicationVariants`, `testVariants` and so on) is called for each variant, using the registered tasks.
     - NB: These abstractions changed a bit.  
- Gradle calculates the task graph, and the build can begin executing.

---
- [x] check integration

PR #635, opened 2020-11-15
MBS-9086 Fix broken screenshots tests recording

eugene-krivobokov:
blocker for #632


subprojects/gradle/design-screenshots/src/main/kotlin/com/avito/plugin/DeviceSearch.kt
```diff
@@ -1,35 +0,0 @@
...
-    ) : DeviceSearch {
-        override fun getDevice(): Either<Exception, AdbDevice> {
-            adbDevicesManager.connectedDevices().let { set ->
-                val either = when (set.size) {
-                    0 -> {
-                        val exception = Exception("There are no connected devices")
-                        Either.left(exception)
-                    }
-                    1 -> {
-                        Either.right(set.first() as AdbDevice)
```
eugene-krivobokov, line 23: Here was the unsafe cast:
```
Execution failed for task ':design-app:recordScreenshots'.
> class com.avito.runner.service.worker.device.adb.AdbDeviceParams cannot be cast to class com.avito.runner.service.worker.device.adb.AdbDevice (com.avito.runner.service.worker.device.adb.AdbDeviceParams and com.avito.runner.service.worker.device.adb.AdbDevice are in unnamed module of loader org.gradle.internal.classloader.VisitableURLClassLoader @eca88ce)
```

subprojects/gradle/design-screenshots/src/main/kotlin/com/avito/plugin/DeviceProvider.kt
```diff
@@ -0,0 +1,50 @@
...
+    private val adbDevicesManager: DevicesManager,
+    private val logger: CILogger
+) : DeviceProvider {
+
+    override fun getDevice(): AdbDevice {
+        val adbDeviceParams = connectedDeviceParams()
+
+        val serial: Serial.Local = when {
+            adbDeviceParams.id is Serial.Local -> adbDeviceParams.id as Serial.Local
+            adbDeviceParams.id is Serial.Remote ->  Serial.Local(adbDeviceParams.id.value) // TODO: support model of locally connected device
```
eugene-krivobokov, line 28: Didn't change in this PR, but it seems a valid case.
It's helpful to connect emulator from local container. Especially for screenshots test.

dsvoronin (approved):

subprojects/gradle/design-screenshots/src/main/kotlin/com/avito/plugin/DeviceProvider.kt
sergioboi: As I know, screenshot generating works only at the `Local emulator`

sergioboi:
infra version of screenshot tests is not used now. So we could not be blocked

subprojects/gradle/design-screenshots/src/main/kotlin/com/avito/plugin/DeviceProvider.kt
eugene-krivobokov: The term "local" is confusing. 
I mean local emulator inside Docker container. It is connected to adb as "localhost:5555".
This is the easiest way for me to use screenshots tests locally.

sergioboi: As I know @DEADMC made screenshots by the emulator without docker on the local machine

PR #639, opened 2020-11-17
MBS-8436 add detekt plugin; apply and fix UnusedPrivateMember rule

dsvoronin:
image


subprojects/build.gradle.kts
```diff
@@ -359,3 +378,21 @@ fun Project.configureJunit5Tests() {
...
+    include("**/*.kts")
+    exclude("**/resources/**")
+    exclude("**/build/**")
+    reports {
+        xml.enabled = false
+        html.enabled = false
+    }
+}
+
+tasks.named("check").dependsOn(detektAll)
```
dsvoronin, line 411: wiring to use standard `build` and `check` tasks chains

Makefile
```diff
@@ -83,6 +83,12 @@ unit_tests:
 compile_tests:
 	./gradlew -p subprojects compileTestKotlin $(log_level)
 
+check:
+	./gradlew -p subprojects check
+
+detekt:
```
dsvoronin, line 89: convenient `make detekt` for local iterations

subprojects/detekt.yml
```diff
@@ -0,0 +1,689 @@
...
+
+output-reports:
+    active: true
+    exclude:
+    # - 'TxtOutputReport'
+    # - 'XmlOutputReport'
+    # - 'HtmlOutputReport'
+
+comments:
+    active: false
```
dsvoronin, line 43: All is copied from detekt default config and disabled (except UnusedPrivateMember);

Proposed pattern is to enable and fix valuable(for us) rules one by one and not use baseline mechanism

subprojects/build.gradle.kts
```diff
@@ -359,3 +378,21 @@ fun Project.configureJunit5Tests() {
...
+    buildUponDefaultConfig = false
+    setSource(files(projectDir))
+    config.setFrom(files(project.rootDir.resolve("detekt.yml")))
+    include("**/*.kt")
+    include("**/*.kts")
+    exclude("**/resources/**")
+    exclude("**/build/**")
+    reports {
+        xml.enabled = false
+        html.enabled = false
```
eugene-krivobokov, line 407: Would it be easier to read html report in a build verdict?

subprojects/build.gradle.kts
```diff
@@ -18,10 +20,12 @@ buildscript {
 }
 
 plugins {
+    base
```
sergioboi, line 27: What is the base?

subprojects/common/okhttp/src/main/kotlin/com/avito/http/RequestBodies.kt
```diff
@@ -34,6 +34,7 @@ fun isPlaintext(buffer: Buffer): Boolean {
         val prefix = Buffer()
         val byteCount = if (buffer.size < 64) buffer.size else 64
         buffer.copyTo(prefix, 0, byteCount)
+        @Suppress("UnusedPrivateMember")
```
sergioboi, line 37: Didn't get the problem with that code

subprojects/detekt.yml
eugene-krivobokov: How about adding this knowledge to this file?

Also, it'd be helpful to add a link to the reference: 
https://github.com/detekt/detekt/blob/master/detekt-core/src/main/resources/default-detekt-config.yml

subprojects/detekt.yml
```diff
@@ -0,0 +1,689 @@
...
+    excludeCorrectable: false
+    weights:
+    # complexity: 2
+    # LongParameterList: 1
+    # style: 1
+    # comments: 1
+
+config:
+    validation: true
+    warningsAsErrors: false
```
sergioboi: True?

subprojects/detekt.yml
```diff
@@ -0,0 +1,689 @@
...
+    ThrowsCount:
+        active: false
+        max: 2
+    TrailingWhitespace:
+        active: false
+    UnderscoresInNumericLiterals:
+        active: false
+        acceptableDecimalLength: 5
+    UnnecessaryAbstractClass:
+        active: false
```
sergioboi, line 639: could we use active = false by default?
I want to see what is enabled ASAP

subprojects/detekt.yml
```diff
@@ -0,0 +1,689 @@
...
+        parameterPattern: '[a-z][A-Za-z0-9]*'
+        privateParameterPattern: '[a-z][A-Za-z0-9]*'
+        excludeClassPattern: '$^'
+        ignoreOverridden: true
+    EnumNaming:
+        active: true
+        excludes: ['**/test/**', '**/androidTest/**', '**/commonTest/**', '**/jvmTest/**', '**/jsTest/**', '**/iosTest/**']
+        enumEntryPattern: '[A-Z][_a-zA-Z0-9]*'
+    ForbiddenClassName:
+        active: false
```
sergioboi, line 375: As I understand If on the top level of `naming` we have active=false then we don't need to duplicate active=false

subprojects/build.gradle.kts
eugene-krivobokov: https://docs.gradle.org/current/userguide/base_plugin.html

subprojects/build.gradle.kts
dsvoronin: don't know how to add it there; just using logs for now

subprojects/common/okhttp/src/main/kotlin/com/avito/http/RequestBodies.kt
dsvoronin: `i` is unused

subprojects/detekt.yml
dsvoronin: done (not affecting anything right now)

subprojects/detekt.yml
dsvoronin: as far as i know detekt doesn't allow this

subprojects/detekt.yml
dsvoronin: i only changed top level one's;
this is detekt defaults

subprojects/common/okhttp/src/main/kotlin/com/avito/http/RequestBodies.kt
sergioboi: Could we change `i` to `_` and remove `Suppress`?

sergioboi (approved):

subprojects/common/okhttp/src/main/kotlin/com/avito/http/RequestBodies.kt
dsvoronin: nope,

dsvoronin: nope, 
image

eugene-krivobokov: It shows that we don't need an index here at all:

```kotlin
repeat(15) { ... }
```

PR #643, opened 2020-11-20
Enable more detekt rules

dsvoronin:


subprojects/gradle/upload-cd-build-result/src/test/kotlin/com/avito/cd/UploadCdBuildResultTaskActionTest.kt
```diff
@@ -21,6 +21,52 @@ class UploadCdBuildResultTaskActionTest {
...
+            "${mockWebServer.url("")}/apps-release-local/appA-android/appA/1-1-100/1.json"
+        )
+    )
+    private val cdBuildConfig = CdBuildConfig(
+        schemaVersion = schemaVersion,
+        project = NupokatiProject.AVITO,
+        releaseVersion = releaseVersion,
+        outputDescriptor = CdBuildConfig.OutputDescriptor("${mockWebServer.url("")}$outputPath", true),
+        deployments = emptyList()
+    )
```
eugene-krivobokov, line 69: Such ordering is unusual for me but I could get used to it.

subprojects/detekt.yml
```diff
@@ -565,6 +565,8 @@ style:
     ExpressionBodySyntax:
         active: false
         includeLineWrapping: true
+    # https://detekt.github.io/detekt/style.html#forbiddencomment
+    # we use todos a lot, maybe there is a better way
```
eugene-krivobokov, line 571: Maybe not in this PR but I'd love to use `STOPSHIP` or similar alternatives.

While coding I often use TODO in two different cases:
- Intentional tech debt. This is a valid usage, nothing to discuss.
- Marker for myself to implement something later but before PR. 
I don't want to miss it but not in all cases I can break the code for the fail-fast approach.
If I miss todo in the second case, it'll be harder to find.

subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/util/DeviceSettingsChecker.kt
```diff
@@ -32,7 +32,7 @@ class DeviceSettingsChecker(private val context: Context) {
     }
 
     private fun checkSupportedApi() {
-        if (Build.VERSION.SDK_INT >= 30) {
+        if (Build.VERSION.SDK_INT > Build.VERSION_CODES.Q) {
```
eugene-krivobokov, line 35: (just grunting) It'll be easier for me to see in Android documentation just numbers instead of these meaningless and hardly comparable names

subprojects/common/report-viewer/src/main/kotlin/com/avito/report/model/TestName.kt
```diff
@@ -20,24 +20,22 @@ data class TestName(val className: String, val methodName: String) {
...
     }
 
 // todo убрать определение features из пакета, будут явно указаны в аннотации к тесту
 val TestName.features: List<String>
     get() = when {
-        packageName.startsWith(domofondPrefix) -> {
+        packageName.startsWith(domofondPrefix) ->
             packageName.substringAfter("$domofondPrefix.")
                 .split(delimiter)
-        }
```
eugene-krivobokov, line 40: I have doubts. I find it's harder to read in cases with long chains or just big statements (try, ...)

dsvoronin: I understood the reason behind that rule as you misguided, especially in big blocks of code, that it's not a single expression.

Also encourages you to extract functions

subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/util/DeviceSettingsChecker.kt
dsvoronin: yep, that's really less clear than before
it's magic number rule, and could be suppressed; what do you think?

subprojects/detekt.yml
dsvoronin: yes, i use this a lot too, but don't like it
i'd prefer separate tooling for such a debt like [codestream](https://www.codestream.com/), but it should be a process around this tool in team

another point is to comment about known issue that may help to debug it in future;

eugene-krivobokov: > i'd prefer separate tooling for such a debt like codestream

Do you mean to use it for intentional tech debt and forbid TODOs at all?
I see no contradictions. We could choose something for the second case (STOPSHIP). We won't see such comments in PR, we won't merge such code. This comment wouldn't bother anyone at all who doesn't use it.

subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/util/DeviceSettingsChecker.kt
eugene-krivobokov: In this case, I'd want to see 30, because it's our logic.

In common Android libraries, version names are valid because they are referenced in the documentation.

up to you

subprojects/common/report-viewer/src/main/kotlin/com/avito/report/model/TestName.kt
eugene-krivobokov: > Also encourages you to extract functions

Yeah, this is a good argument for it.

eugene-krivobokov (approved):

subprojects/detekt.yml
dsvoronin: yep, we don't have a process to keep todo's count close to zero, so it's only purpose now is to give a warning for next change about possible quirks
It could be just a regular comment

eugene-krivobokov: Ok, I'll enable STOPSHIP'like stopper in an additional PR

eugene-krivobokov: #956

PR #644, opened 2020-11-23
Fix RecyclerViewInteractionContext.kt

sergioboi:
merge cell and child matcher because cellmatcher wasn't be propagated when `provideChildContext`


dsvoronin:
Do you have plans to merge this PR? maybe close it for now?

PR #648, opened 2020-11-24
Remove unused code

sergioboi:


subprojects/android-test/ui-testing-core/src/main/kotlin/com/avito/android/test/InteractionContext.kt
```diff
@@ -86,11 +86,6 @@ class RecyclerViewInteractionContext(
             needScroll = needScroll
         )
 
-        ActionInterceptor.Proxy(
```
dsvoronin, line 89: it's definitely unused code, but i'd rather do:

```
val proxyedAction = ActionInterceptor.Proxy(
            source = actionOnItem,
            interceptors = UITestConfig.actionInterceptors
        )

interactionContext.perform(proxyedAction)
```

sergioboi: It will be proxied below when execute `interactionContext.perform(actionOnItem)`

dsvoronin (approved):

PR #650, opened 2020-11-25
remove kotlin root plugin

dsvoronin:
hard to maintain in different repositories; cause problems with version desync


eugene-krivobokov:
Need to remove mentions from the documentation: `Infrastructure.md`

eugene-krivobokov:
- [ ] Merge only after new implementation in the internal repository

eugene-krivobokov (approved):

PR #654, opened 2020-11-26
add ImpactMetrics ciStep

dsvoronin:


subprojects/gradle/cd/src/main/kotlin/com/avito/ci/steps/ImpactMetrics.kt
```diff
@@ -0,0 +1,22 @@
...
+import com.avito.kotlin.dsl.isRoot
+import com.avito.kotlin.dsl.namedOrNull
+import org.gradle.api.Project
+import org.gradle.api.Task
+import org.gradle.api.tasks.TaskProvider
+
+class ImpactMetrics(context: String, name: String) : BuildStep(context, name) {
+
+    override fun registerTask(project: Project, rootTask: TaskProvider<out Task>) {
+        require(project.isRoot()) { failMessage("should be configured for root project") }
```
sergioboi, line 12: How could we add this step to the root project?

eugene-krivobokov: I don't see restrictions for the root project. It will add a task to it, e.g. `:fastCheck`

subprojects/gradle/cd/src/main/kotlin/com/avito/ci/steps/ImpactMetrics.kt
```diff
@@ -0,0 +1,22 @@
...
+
+        val impactMetricsTask = requireNotNull(project.tasks.namedOrNull("impactMetrics")) {
+            failMessage("impactMetrics task not registered for root project")
+        }
+
+        rootTask.configure { it.dependsOn(impactMetricsTask) }
+    }
```
eugene-krivobokov, line 19: This build step shows a common simple pattern - add `dependsOn` to a custom task.
In this scenario, we know nothing about the peculiarities of the task.
Thus, we can make it reusable. From a client, we need only a task provider.

How it may look:

```
register("fastCheck") {
    task("impactMetrics")
    task(taskProvider)
}
```

Or course, there may be other points for extensibility: impact analysis, ...

I know that we want to reconsider CI steps architecture. But it seems a reasonable temporary compromise to move some existing steps to the client-side.

eugene-krivobokov (approved):

PR #668, opened 2020-12-07
MBS-9988 Fix Kotlin caching for infra clients

eugene-krivobokov:


subprojects/gradle/instrumentation-tests/build.gradle.kts
```diff
@@ -36,7 +36,7 @@ dependencies {
     implementation(project(":gradle:kotlin-dsl-support"))
     implementation(project(":gradle:process"))
     implementation(project(":gradle:runner:client"))
-    implementation(project(":gradle:runner:shared-test"))
```
eugene-krivobokov, line 40: This dependency changes inputs in compile tasks and breaks caching.

Couldn't find the exact reasons. I've suspected `random` functions but couldn't prove it. I've tried to remove all unstable values but only removing test dependencies helped.
Anyway, we don't want to have test dependencies in build scripts.

Discussion in Gradle slack: [link](https://gradle-community.slack.com/archives/CAD95CR62/p1606304726060600)

subprojects/gradle/runner/fake/build.gradle.kts
```diff
@@ -0,0 +1,13 @@
...
+    id("com.jfrog.bintray")
+}
+
+extra["artifact-id"] = "runner-fake"
+
+dependencies {
+    implementation(project(":common:report-viewer"))
+    implementation(project(":gradle:runner:service"))
+    implementation(project(":gradle:runner:shared"))
+}
```
eugene-krivobokov, line 13: Made a new module for fakes for two reasons:
- It shouldn't depend on tests
- It needs to know about the test report. Runner itself shouldn't.

dsvoronin (approved):

PR #669, opened 2020-12-07
fix not final class

dsvoronin:


eugene-krivobokov (approved):

subprojects/gradle/instrumentation-tests-dex-loader/src/main/kotlin/com/avito/android/LoadTestsFromApkTask.kt
```diff
@@ -13,16 +16,16 @@ import javax.inject.Inject
 
 @CacheableTask
 @Suppress("UnstableApiUsage")
-class LoadTestsFromApkTask @Inject constructor(
+open class LoadTestsFromApkTask @Inject constructor(
```
eugene-krivobokov, line 19: Gradle uses `abstract` to show an intention

sergioboi: Why didn't make it abstract?

dsvoronin: i missed your comment, will do in next PR

PR #672, opened 2020-12-09
Change dependencies from unsafe String to the type safe decaration

sergioboi:


subprojects/build.gradle.kts
```diff
@@ -400,6 +400,7 @@ fun Project.configureJunit5Tests() {
     tasks.withType<Test> {
         useJUnitPlatform()
         maxParallelForks = 8
+        setForkEvery(1)
```
sergioboi: By default, the Test task doesn't fork the execution process
Now on each Test class, we will start a new process with the maximum amount equals to 8

dsvoronin: Nice,
check memory consumption though, because i did no tuning here and gradle test kit inherit daemon settings by default

subprojects/gradle/test-project/src/main/kotlin/com/avito/test/gradle/dependencies/GradleDependency.kt
```diff
@@ -0,0 +1,75 @@
+package com.avito.test.gradle.dependencies
+
+import com.avito.test.gradle.dependencies.GradleDependency.Coordinate.Maven
+import com.avito.test.gradle.dependencies.GradleDependency.Coordinate.Platform
+import com.avito.test.gradle.dependencies.GradleDependency.Coordinate.Project
+import org.funktionale.either.Either
+import org.gradle.util.Path
+
+data class GradleDependency(
```
dsvoronin: With all respect to intention (hate dynamic things), i honestly don't like the final result, and that's why:

1. i need to learn another api around gradle that is less configurable and can possibly block some uncommon things if future
2. strings was chosen exactly because of this ^^ , as lesser evil
3. new api is misleading: what is maven? afaik it's artifacts coordinates notation, introduced by gradle, compatible with maven/ivy etc

subprojects/build.gradle.kts
dsvoronin: Check memory consumption, i did no tuning for gradle test kit parameters and its inherited from daemon settings

eugene-krivobokov: Why do we need a new process for each test? Do we have any problems?

subprojects/gradle/test-project/src/main/kotlin/com/avito/test/gradle/dependencies/GradleDependency.kt
eugene-krivobokov: plus for maven

subprojects/gradle/test-project/src/main/kotlin/com/avito/test/gradle/dependencies/GradleDependency.kt
```diff
@@ -0,0 +1,75 @@
...
+            configuration: CONFIGURATION = CONFIGURATION.IMPLEMENTATION,
+            mavenCoordinate: String
+        ): GradleDependency = GradleDependency(configuration, Maven(mavenCoordinate))
+
+        fun project(
+            configuration: CONFIGURATION = CONFIGURATION.IMPLEMENTATION,
+            projectPath: String
+        ): GradleDependency = GradleDependency(configuration, Project(Path.path(projectPath)))
+
+        fun platformProject(
```
eugene-krivobokov:
```suggestion
        fun platform(
```

The closer to the original API, the better

```
api(platform("com.fasterxml.jackson:jackson-bom:2.9.8"))
api(platform(project(":module")))
```

subprojects/gradle/test-project/src/main/kotlin/com/avito/test/gradle/dependencies/GradleDependency.kt
sergioboi: > i need to learn another api around gradle that is less configurable

Don't agree that we need "super all cases Gradle API". Only we need is a simple type-safe way to declare dependencies for all our cases. I prefer cut API to String parameter

> strings was chosen exactly because of this ^^ , as lesser evil

I don't like String because I spent time digging into the TestProjectGenerator implementation to realize how it actually will be pasted to `build.gradle`

> new api is misleading: what is maven?

that only point I agree with. Could you help me choose correct name?

subprojects/gradle/test-project/src/test/kotlin/com/avito/test/gradle/TestProjectGeneratorTest.kt
```diff
@@ -109,11 +110,11 @@ class TestProjectGeneratorTest {
...
             "app",
-            dependencies = """
-            implementation project(':parent:empty:library')
-            implementation project(':parent:empty:kotlin')
-            implementation project(':parent:empty:platform')
-        """.trimIndent()
+            dependencies = setOf(
+                project(projectPath = ":parent:empty:library"),
+                project(projectPath = ":parent:empty:kotlin"),
+                project(projectPath = ":parent:empty:platform")
```
eugene-krivobokov: How about just `path`?

```kotlin
project(path = ":parent:empty:library"),
project(path = ":parent:empty:kotlin"),
project(path = ":parent:empty:platform")
```

I don't see reasons to duplicate this knowledge.

subprojects/gradle/test-project/src/main/kotlin/com/avito/test/gradle/dependencies/GradleDependency.kt
```diff
@@ -0,0 +1,75 @@
...
+
+    companion object {
+        fun maven(
+            configuration: CONFIGURATION = CONFIGURATION.IMPLEMENTATION,
+            mavenCoordinate: String
+        ): GradleDependency = GradleDependency(configuration, Maven(mavenCoordinate))
+
+        fun project(
+            configuration: CONFIGURATION = CONFIGURATION.IMPLEMENTATION,
+            projectPath: String
```
eugene-krivobokov: How about moving this argument to the first position?

```kotlin
project(":parent:empty:library"),
project(":parent:empty:kotlin"),
project(":parent:empty:platform")
```

subprojects/gradle/test-project/src/main/kotlin/com/avito/test/gradle/dependencies/GradleDependency.kt
eugene-krivobokov: Possible option: external dependency (https://docs.gradle.org/current/javadoc/org/gradle/api/artifacts/dsl/DependencyHandler.html)

subprojects/build.gradle.kts
sergioboi: Check locally it changes nothing
Removed

subprojects/gradle/test-project/src/main/kotlin/com/avito/test/gradle/dependencies/GradleDependency.kt
sergioboi: Stay without changes

subprojects/gradle/test-project/src/main/kotlin/com/avito/test/gradle/dependencies/GradleDependency.kt
```diff
@@ -0,0 +1,81 @@
...
+            TEST_IMPLEMENTATION("testImplementation"),
+            ANDROID_TEST_IMPLEMENTATION("androidTestImplementation"),
+            API("api");
+
+            override fun getScriptRepresentation() = representation
+        }
+
+        sealed class Coordinate : GradleScriptCompatible {
+
+            data class Maven(val mavenCoordinates: String) : Coordinate() {
```
eugene-krivobokov: We've discussed a more generic name for external dependencies

subprojects/gradle/impact/src/test/kotlin/com/avito/impact/ImpactAnalysisTest.kt
```diff
@@ -178,10 +184,10 @@ class ImpactAnalysisTest {
                 AndroidAppModule("standalone_app"),
                 AndroidAppModule(
                     name = "app",
-                    dependencies = """
-                        implementation project(':feature_a')
-                        implementation project(':feature_b')
-                    """
+                    dependencies = setOf(
+                        project(path = ":feature_a", configuration = IMPLEMENTATION),
```
dsvoronin: do you need to write specific "IMPLEMENTATION" configuration here? maybe omit it as default one?
or it makes sense for current tests readability?

subprojects/gradle/test-project/src/main/kotlin/com/avito/test/gradle/dependencies/GradleDependency.kt
```diff
@@ -0,0 +1,81 @@
...
+                override fun getScriptRepresentation() = "'$mavenCoordinates'"
+            }
+
+            data class Project(val projectPath: Path) : Coordinate() {
```
eugene-krivobokov:
```suggestion
            data class Project(val path: Path) : Coordinate() {
```
?

subprojects/gradle/test-project/src/main/kotlin/com/avito/test/gradle/dependencies/GradleDependency.kt
```diff
@@ -0,0 +1,82 @@
...
+
+            fun platformProject(
+                path: String,
+                configuration: CONFIGURATION = CONFIGURATION.IMPLEMENTATION
+            ): Safe = Safe(
+                configuration,
+                Platform(Either.right(Project(Path.path(path))))
+            )
+
+            fun platformMaven(
```
eugene-krivobokov:
```suggestion
            fun platformExternal(
```
?

subprojects/gradle/test-project/src/main/kotlin/com/avito/test/gradle/dependencies/GradleDependency.kt
eugene-krivobokov: Why?

subprojects/gradle/test-project/src/main/kotlin/com/avito/test/gradle/dependencies/GradleDependency.kt
sergioboi: Done

eugene-krivobokov (approved):

PR #680, opened 2020-12-16
MBS-9407 configurable memory request for k8s emulators

dsvoronin:
remove default 3.5Gi setting


eugene-krivobokov (approved):

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/InstrumentationTestsPlugin.kt
```diff
@@ -94,7 +95,23 @@ class InstrumentationTestsPlugin : Plugin<Project> {
                     group = ciTaskGroup
                 }
 
-                val instrumentationTask = project.tasks.register<InstrumentationTestsTask>(
+                if (instrumentationConfiguration.impactAnalysisPolicy is On.RunChangedTests) {
+                    if (project.plugins.hasPlugin(InstrumentationChangedTestsFinderApi.pluginId)) {
```
eugene-krivobokov: Why not `withPlugin`?

dsvoronin: i created separate PR for these changes

sergioboi:
Sorry, but I don't get the intention of this PR. Why did we remove default values and make properties nullable?

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/reservation/client/kubernetes/ReservationDeploymentFactory.kt
```diff
@@ -125,13 +125,23 @@ class ReservationDeploymentFactory(
...
+                        }
+
+                        requests = mutableMapOf<String, Quantity>().apply {
+                            if (!emulator.cpuCoresRequest.isNullOrBlank()) {
+                                plusAssign("cpu" to Quantity(emulator.cpuCoresRequest))
+                            }
+                            if (!emulator.memoryRequest.isNullOrBlank()) {
+                                plusAssign("memory" to Quantity(emulator.memoryRequest))
+                            }
+                        }
```
eugene-krivobokov, line 144: Do we have an implicit contract for requests and limits?
E.g. are pods without limits and requests make sense and valid?

eugene-krivobokov: If I see an original error from k8s, it will be enough.

dsvoronin: Yes and no; We could create it for some test purposes, but in real load it should be set

dsvoronin:
> Sorry, but I don't get the intention of this PR. Why did we remove default values and make properties nullable?

this PR introduces memoryRequest property, we couldn't set it before and it's needed for additional tuning
nullability is perfectly valid from k8s standpoint and default value also solves problem with backward compatibility

PR #682, opened 2020-12-18
MBS-9407 fix impact analysis config for changed tests

dsvoronin:


eugene-krivobokov (approved):

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/InstrumentationTestsPlugin.kt
```diff
@@ -94,7 +95,23 @@ class InstrumentationTestsPlugin : Plugin<Project> {
                     group = ciTaskGroup
                 }
 
-                val instrumentationTask = project.tasks.register<InstrumentationTestsTask>(
+                if (instrumentationConfiguration.impactAnalysisPolicy is On.RunChangedTests) {
```
sergioboi, line 98: Sorry but I forgot everything you did.
If you don't mind could you explain the difference between: On.RunChangedTests, On.RunAffectedTests, On.RunNewTests, On.RunModifiedTests
And do we need all these types now?

PR #683, opened 2020-12-18
Slack client fix for nullable fields

dsvoronin:
`username` and `botId` could be legally null
Integration test started to fail because some messages in integration test channel had empty author

### DIFF
--- a/subprojects/gradle/slack/src/main/kotlin/com/avito/slack/SlackClient.kt
+++ b/subprojects/gradle/slack/src/main/kotlin/com/avito/slack/SlackClient.kt
@@ -34,9 +34,9 @@ interface SlackClient : SlackMessageSender, SlackFileUploader {
         private val methodsClient: MethodsClient = Slack.getInstance().methods()
 
         /**
-         * для [findMessage] фетчим [messagesLookupCount] сообщений вверх, число просто взято "с запасом"
+         * fetch [messagesLookupCount] messages back though history
          */
-        private val messagesLookupCount = 100
+        private val messagesLookupCount = 25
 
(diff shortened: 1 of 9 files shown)


subprojects/gradle/slack/src/main/kotlin/com/avito/slack/SlackClient.kt
```diff
@@ -34,9 +34,9 @@ interface SlackClient : SlackMessageSender, SlackFileUploader {
         private val methodsClient: MethodsClient = Slack.getInstance().methods()
 
         /**
-         * для [findMessage] фетчим [messagesLookupCount] сообщений вверх, число просто взято "с запасом"
+         * fetch [messagesLookupCount] messages back though history
          */
-        private val messagesLookupCount = 100
+        private val messagesLookupCount = 25
```
dsvoronin, line 39: It just speedup things a bit with no impact on stability in current cases

subprojects/gradle/slack/src/main/kotlin/com/avito/slack/model/FoundMessage.kt
```diff
@@ -2,9 +2,9 @@ package com.avito.slack.model
 
 data class FoundMessage(
     val text: String,
-    val botId: String,
+    val botId: String?,
     val timestamp: String,
-    val author: String,
+    val author: String?,
```
dsvoronin, line 7: This could be service messages, and user sees integration hook "name" as username, but in fact message author fetched as null

same for botId

eugene-krivobokov (approved):

PR #684, opened 2020-12-18
Increase test timeouts

dsvoronin:
Some gradle test kit tests goes up to 8 min according to statistics
We should tune that, but right now just removing blocker


subprojects/build.gradle.kts
```diff
@@ -287,7 +287,7 @@ subprojects {
 
                 systemProperty("isTest", true)
 
-                systemProperty("junit.jupiter.execution.timeout.default", MINUTES.toSeconds(5))
+                systemProperty("junit.jupiter.execution.timeout.default", MINUTES.toSeconds(10))
```
eugene-krivobokov, line 290: Will you plan investigation?

dsvoronin: created internal MBS-10204 for this

subprojects/gradle/instrumentation-tests/src/integTest/kotlin/com/avito/instrumentation/executing/TestExecutorIntegrationTest.kt
```diff
@@ -19,16 +19,13 @@ import com.avito.utils.logging.CILogger
...
-import org.junit.jupiter.api.Timeout
 import org.junit.jupiter.api.io.TempDir
 import java.io.File
-import java.util.concurrent.TimeUnit.SECONDS
 
 internal class TestExecutorIntegrationTest {
 
     private val logger = FakeCILogger()
 
-    @Timeout(10, unit = SECONDS)
```
sergioboi, line 31: Why did you remove those?

subprojects/build.gradle.kts
sergioboi: Where did you see the statistics?

dsvoronin: it's in task description, i don't have one, just seeing this time to time on Teamcity's "Tests" tab

subprojects/gradle/instrumentation-tests/src/integTest/kotlin/com/avito/instrumentation/executing/TestExecutorIntegrationTest.kt
dsvoronin: because i believe global timeouts is enough, and we don't need to tune it per test

subprojects/build.gradle.kts
eugene-krivobokov: How many tests require 10 min? Can we override timeout only for them?

dsvoronin: i just don't know; seems like it's random with GradleTestKit one's
we didn't have any of timeouts, now added 10 min timeout and planned to add some statistic and fix it
i see it as an improvement worth merging,
because in current state we only introduce flakiness

subprojects/gradle/instrumentation-tests/src/integTest/kotlin/com/avito/instrumentation/executing/TestExecutorIntegrationTest.kt
sergioboi: I agree in common but in that Test `Timeout` is an intention of our testing we fix timeouts.

dsvoronin: i found that sometimes this test timings are also fluctuate and could run over 30sec.
I don't know what is the correct value here, so just removed it in favor of global timeouts

sergioboi: In my opinion, this test should take 1 sec maximum. We could pairly fix this strange behavior

eugene-krivobokov (approved):

subprojects/gradle/instrumentation-tests/src/integTest/kotlin/com/avito/instrumentation/executing/TestExecutorIntegrationTest.kt
```diff
@@ -19,16 +19,13 @@ import com.avito.utils.logging.CILogger
 import com.avito.utils.logging.StubCILogger
 import com.avito.utils.logging.commonLogger
 import org.junit.jupiter.api.Test
-import org.junit.jupiter.api.Timeout
 import org.junit.jupiter.api.io.TempDir
 import java.io.File
-import java.util.concurrent.TimeUnit.SECONDS
 
 internal class TestExecutorIntegrationTest {
```
eugene-krivobokov, line 25: Please, make an issue about testing in minikube/kind/... like environment.
It will be easier to cover edge-cases more hermetically.

PR #685, opened 2020-12-18
rename fake,mock -> stub

dsvoronin:
see discussion: #679


sergioboi (approved):

sergioboi:
Let's make a documentation page about that decision for future

PR #689, opened 2020-12-21
MBS-10164 Check unique R classes in all configurations

eugene-krivobokov:
Workaround for https://issuetracker.google.com/issues/175936041

### DIFF
--- a/subprojects/gradle/buildchecks/src/main/kotlin/com/avito/android/plugin/build_param_check/UniqueRClassesTask.kt
+++ b/subprojects/gradle/buildchecks/src/main/kotlin/com/avito/android/plugin/build_param_check/UniqueRClassesTask.kt
@@ -13,11 +13,16 @@ abstract class UniqueRClassesTask : DefaultTask() {
     fun check() {
         val apps = project.subprojects.filter { it.isAndroidApp() }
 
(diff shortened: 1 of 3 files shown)


dsvoronin (approved):

subprojects/gradle/buildchecks/src/main/kotlin/com/avito/android/plugin/build_param_check/UniqueRClassesTask.kt
```diff
@@ -13,11 +13,16 @@ abstract class UniqueRClassesTask : DefaultTask() {
     fun check() {
         val apps = project.subprojects.filter { it.isAndroidApp() }
 
-        apps.forEach { app ->
+        apps.stream().parallel().forEach { app ->
```
sergioboi, line 16: Did you check that `forEech` gave increasion?

eugene-krivobokov: The whole task works longer becuase of additional configurations to check: 1 min become 3 min.
Avito app is on critical path, other applications much faster.

Without parallelization the task tooks 6 min.

PR #696, opened 2020-12-29
Add ability to add <a/> and other html tags to html version of a build verdict

sergioboi:


subprojects/gradle/build-verdict-tasks-api/src/main/kotlin/com/avito/android/build_verdict/span/SpannedString.kt
```diff
@@ -0,0 +1,29 @@
+package com.avito.android.build_verdict.span
+
+import java.net.URL
+
+data class SpannedString(
```
sergioboi, line 5: This is a new type for presenting HTML parts in the build verdict by users
I tried many solutions but choose the current version. It is flexible, extensible but not a type-safe
Not type-safe means that users could set any String like an HTML and we could fail on rendering.
If we will face problems with the wrong HTML we may change API

subprojects/gradle/build-verdict/src/main/kotlin/com/avito/android/build_verdict/BuildVerdictPluginExtension.kt
```diff
@@ -35,11 +36,11 @@ abstract class BuildVerdictPluginExtension(
         )
     }
 
-    fun onTaskFailure(name: String, producer: (Task) -> String) {
+    fun onTaskFailure(name: String, producer: (Task) -> SpannedString) {
```
sergioboi, line 39: This PR is incompatible with the old API and will break current users because I don't see a reason to make this change in two steps

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/InstrumentationTestsTask.kt
```diff
@@ -120,8 +129,15 @@ abstract class InstrumentationTestsTask @Inject constructor(
     private val verdictFile = objects.fileProperty().convention(output.file("verdict.json"))
 
     @get:Internal
-    override val verdict: String
-        get() = verdictFile.asFile.get().readText()
+    override val verdict: SpannedString
+        get() {
+            val verdictRaw = verdictFile.asFile.get().reader()
+            val verdict = gson.fromJson<Verdict>(verdictRaw)
+            return link(
```
sergioboi, line 141: This is how the link creation looks on the user's side

subprojects/gradle/build-verdict/src/main/kotlin/com/avito/android/build_verdict/BuildVerdictPluginExtension.kt
dsvoronin: Some more context to this (as i went to our project to find current clients):

```
    onTaskFailure("detektAll") {
        File(detekt.reportsDir, "detekt.txt").readText()
    }
```

So this lambda will be red because of changed return value;

look ok to me, minor thing to remember on version update

subprojects/gradle/build-verdict-tasks-api/src/main/kotlin/com/avito/android/build_verdict/span/SpannedString.kt
```diff
@@ -0,0 +1,29 @@
...
+    val plain: String,
+    val html: String
+) {
+    constructor(value: String) : this(value, value)
+
+    companion object {
+
+        @JvmStatic
```
dsvoronin, line 13: this is for groovy, right?

subprojects/gradle/build-verdict/src/main/kotlin/com/avito/android/build_verdict/interna/span/SpannedStringBuilder.kt
```diff
@@ -0,0 +1,33 @@
+package com.avito.android.build_verdict.interna.span
```
dsvoronin: package name typo: `interna` vs `internal`

does not add as a github suggestion, because it required refactoring in IDE

subprojects/gradle/build-verdict/src/test/kotlin/com/avito/android/build_verdict/BuildVerdictPluginExecutionPhaseTest.kt
```diff
@@ -215,9 +216,13 @@ class BuildVerdictPluginExecutionPhaseTest : BaseBuildVerdictTest() {
             ),
             buildGradleExtra = """
 import com.avito.android.build_verdict.TaskVerdictProducer
+import com.avito.android.build_verdict.span.SpannedString
 import org.gradle.api.Task
+
 buildVerdict {
-    onTaskFailure("customTask", { "User added verdict" } as TaskVerdictProducer)
+    onTaskFailure("customTask", { 
+        SpannedString.link("https://www.google.com/", "User added verdict")
```
dsvoronin, line 224: named arguments would be nice here

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/InstrumentationTestsActionFactory.kt
```diff
@@ -50,4 +50,8 @@ interface InstrumentationTestsActionFactory {
 
         override fun provideFinalizer() = finalizerFactory.create()
     }
+
+    companion object {
+        val gson by lazy { GsonBuilder().setPrettyPrinting().create() }
```
dsvoronin, line 55: can it be private?

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/InstrumentationTestsTask.kt
dsvoronin: i like not statically imported `SpannedString.link()` more, easier to read where it comes from

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/InstrumentationTestsActionFactory.kt
dsvoronin, line 55: it looks like this gson producer used everywhere in plugin, so should be passed around as an argument or placed on package level, not in concrete `InstrumentationTestsActionFactory` companion

dsvoronin (approved):
LGTM, i want you to answer my comments and questions, but could be merged in this state

subprojects/gradle/build-verdict-tasks-api/src/main/kotlin/com/avito/android/build_verdict/span/SpannedString.kt
dsvoronin: Maybe we should restrict clients to return only Strings for now? I mean `onTaskFailure` methods
Just convert `String` to `SpannedString` on plugin side then
Do we really need that freedom? And if yes for what particular task?
Additional benefit: backward compatibility

subprojects/gradle/build-verdict/src/main/kotlin/com/avito/android/build_verdict/BuildVerdictPluginExtension.kt
dsvoronin: see: https://github.com/avito-tech/avito-android/pull/696#discussion_r549892299

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/InstrumentationTestsActionFactory.kt
sergioboi: I agree, it's common question: How to share common components like the `gson` in the plugin.
The ideal world is when we set everything via object constructors but we live with Gradle and to do this I have to write much code.
Dependency Inversion solves two problems: reusing and simple changing dependencies for tests for example.

In that case, I don't see any reason for refactoring that code because I don't need to change implementations now.
And I see `InstrumentationTestsActionFactory` as a normal place for `gson` constant which used through the plugin.

Could you tell me how you image `gson` at the package level?

subprojects/gradle/build-verdict/src/test/kotlin/com/avito/android/build_verdict/BuildVerdictPluginExecutionPhaseTest.kt
sergioboi: It's a groovy script and I can't

subprojects/gradle/build-verdict-tasks-api/src/main/kotlin/com/avito/android/build_verdict/span/SpannedString.kt
sergioboi: Yes

subprojects/gradle/build-verdict/src/main/kotlin/com/avito/android/build_verdict/BuildVerdictPluginExtension.kt
sergioboi: Yes, I want to break compatibility and fix it in the release commit

subprojects/gradle/build-verdict-tasks-api/src/main/kotlin/com/avito/android/build_verdict/span/SpannedString.kt
sergioboi: I've already restricted clients from returning `String`

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/InstrumentationTestsTask.kt
eugene-krivobokov: Will you describe the usage of the plugin in documentation?

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/InstrumentationTestsActionFactory.kt
dsvoronin: Just some singleton(object) for global val's for instrumentation plugin

sergioboi: I can't choose the name for that object

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/InstrumentationTestsTask.kt
sergioboi: I see it selfdocumented because I wrote it
Could you give me an example of cases you want to see in the documentation

eugene-krivobokov: I mean the plugin in general, as others

- Getting started
- How to add a custom report

sergioboi: I will create a task for this

PR #701, opened 2021-01-12
explicit api turned in warning mode

dsvoronin:


dsvoronin:
Will enabled it module-per-module

PR #702, opened 2021-01-13
Speedup UniqueRClasses build check

eugene-krivobokov:
Build check for [unique package names](https://avito-tech.github.io/avito-android/docs/projects/buildchecks/#unique-r-classes) (R classes) was too slow. It took ~3min for Avito application.
The problem is in impact analysis. To find all manifests to merge we had to resolve all dependencies.
For now, we can't do impact analysis inside Gradle workers.

I've switched to consuming manifests directly from merging tasks. 
It gives caching and mostly instant execution - up to 0.1s for the check itself. Merging manifests also fast.

### DIFF
--- a/subprojects/gradle/buildchecks/src/main/kotlin/com/avito/android/plugin/build_param_check/UniqueRClassesTask.kt
+++ b/subprojects/gradle/buildchecks/src/main/kotlin/com/avito/android/plugin/build_param_check/UniqueRClassesTask.kt
@@ -1,43 +1,83 @@
 package com.avito.android.plugin.build_param_check
 
-import com.avito.android.isAndroid
 import com.avito.android.isAndroidApp
-import com.avito.impact.configuration.internalModule
-import com.avito.impact.util.AndroidManifest
+import com.avito.impact.util.AndroidManifestPackageParser
 import org.gradle.api.DefaultTask
+import org.gradle.api.file.FileCollection
+import org.gradle.api.file.RegularFile
+import org.gradle.api.file.RegularFileProperty
+import org.gradle.api.model.ObjectFactory
+import org.gradle.api.provider.ListProperty
+import org.gradle.api.provider.Property
+import org.gradle.api.tasks.CacheableTask
+import org.gradle.api.tasks.Input
+import org.gradle.api.tasks.InputFiles
+import org.gradle.api.tasks.OutputFile
+import org.gradle.api.tasks.PathSensitive
+import org.gradle.api.tasks.PathSensitivity
 import org.gradle.api.tasks.TaskAction
+import javax.inject.Inject
 
-abstract class UniqueRClassesTask : DefaultTask() {
+@CacheableTask
+abstract class UniqueRClassesTask @Inject constructor(
+    objects: ObjectFactory
+) : DefaultTask() {
+
+    @get:InputFiles
+    @get:PathSensitive(PathSensitivity.RELATIVE)
+    abstract val appManifest: RegularFileProperty
+
+    @get:InputFiles
+    @get:PathSensitive(PathSensitivity.RELATIVE)
+    abstract val librariesManifests: Property<FileCollection>
+
+    @get:InputFiles
+    @get:PathSensitive(PathSensitivity.RELATIVE)
+    abstract val testManifests: Property<FileCollection>
+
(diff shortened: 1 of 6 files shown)


subprojects/gradle/buildchecks/src/main/kotlin/com/avito/android/plugin/build_param_check/UniqueRClassesTaskFactory.kt
```diff
@@ -0,0 +1,63 @@
...
+            allowedNonUniquePackageNames.set(config.allowedNonUniquePackageNames)
+
+            val processAppManifest = project.tasks.withType(ProcessApplicationManifest::class.java)
+                .first { it.variantName == appVariant.name }
+
+            val processTestManifest = project.tasks.withType(ProcessTestManifest::class.java).first()
+
+            appManifest.set(processAppManifest.mergedManifest)
+            librariesManifests.set(processAppManifest.getManifests())
+            testManifests.set(processTestManifest.getManifests())
```
eugene-krivobokov, line 57: > I've switched to consuming manifests directly from merging tasks.

dsvoronin (approved):

subprojects/gradle/buildchecks/src/main/kotlin/com/avito/android/plugin/build_param_check/UniqueRClassesTask.kt
```diff
@@ -1,43 +1,83 @@
...
                     Please, make packages unique.
                     """
-                    ).toString()
-                )
-            }
+                ).toString()
+            )
         }
+
+        output.get().asFile.writeText("1")
```
sergioboi, line 80: Let's replace "1" with "No duplicates"

subprojects/gradle/buildchecks/src/main/kotlin/com/avito/android/plugin/build_param_check/UniqueRClassesTaskFactory.kt
```diff
@@ -0,0 +1,63 @@
...
+
+            dependsOn(processAppManifest)
```
sergioboi, line 59: As I understand Gradle correct you don't need a dependsOn here because Gradle could inference that from properties set

subprojects/gradle/buildchecks/src/main/kotlin/com/avito/android/plugin/build_param_check/UniqueRClassesTaskFactory.kt
```diff
@@ -0,0 +1,63 @@
...
+    private val config: Check.UniqueRClasses
+) {
+
+    init {
+        check(rootProject.isRoot()) {
+            "Project ${rootProject.path} must be root"
+        }
+    }
+
+    fun register(rootTask: TaskProvider<Task>) {
```
sergioboi, line 27: Combination of className `UniqueRClassesTaskFactory`, function name `register` and param name `rootTask` is unclear for me
I don't know how to rename the factory but propose rename register to the dependsOn or something similar

And tasks creation is implementation details

eugene-krivobokov: Renamed to this in #781 :

```kotlin
UniqueRClassesTaskProvider(...)
                .dependsOn(rootTask)
```

subprojects/gradle/buildchecks/src/main/kotlin/com/avito/android/plugin/build_param_check/UniqueRClassesTaskFactory.kt
eugene-krivobokov: Thanks, removed them in #781 
It's an implicit behavior. I need to get used to it.

subprojects/gradle/buildchecks/src/main/kotlin/com/avito/android/plugin/build_param_check/UniqueRClassesTask.kt
eugene-krivobokov: Fixed in #781

PR #703, opened 2021-01-13
release branch cherrypick (logger refactor leftovers)

dsvoronin:
bump version

### DIFF
--- a/subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/runner/InHouseInstrumentationTestRunner.kt
+++ b/subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/runner/InHouseInstrumentationTestRunner.kt
@@ -286,7 +286,7 @@ abstract class InHouseInstrumentationTestRunner :
             report.registerIncident(AppCrashException(incident))
             report.reportTestCase()
         } catch (t: Throwable) {
-            logger.critical("Can't register and report unexpected incident", t)
+            logger.warn("Can't register and report unexpected incident", t)
         }
     }
 
--- a/subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/runner/InHouseInstrumentationTestRunner.kt
+++ b/subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/runner/InHouseInstrumentationTestRunner.kt
@@ -2,7 +2,6 @@ package com.avito.android.runner
 
 import android.annotation.SuppressLint
 import android.os.Bundle
-import android.util.Log
 import androidx.annotation.CallSuper
 import androidx.test.espresso.Espresso
 import androidx.test.platform.app.InstrumentationRegistry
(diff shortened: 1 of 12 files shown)


subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/runner/InHouseInstrumentationTestRunner.kt
```diff
@@ -280,13 +281,12 @@ abstract class InHouseInstrumentationTestRunner :
...
 
-    @SuppressLint("LogNotTimber")
-    fun tryToReportUnexpectedIncident(incident: Throwable, tag: String) {
+    fun tryToReportUnexpectedIncident(incident: Throwable) {
         try {
             report.registerIncident(AppCrashException(incident))
             report.reportTestCase()
         } catch (t: Throwable) {
-            Log.w(tag, "Can't register and report unexpected incident", t)
+            logger.critical("Can't register and report unexpected incident", t)
```
eugene-krivobokov: Is it a failure for tests running?

subprojects/android-test/test-report/src/main/kotlin/com/avito/android/test/report/ArgsProvider.kt
```diff
@@ -49,4 +51,14 @@ class BundleArgsProvider(
             return result
         }
     }
+
+    override fun getOptionalIntArgument(name: String): Int? {
+        val defaultValue = Int.MIN_VALUE
+        val result = bundle.getInt(name, defaultValue)
+        return if (result == defaultValue) {
```
eugene-krivobokov, line 58: Why don't we check it by `containsKey`?

subprojects/build.gradle.kts
```diff
@@ -325,7 +325,7 @@ fun Project.configureBintray(vararg publications: String) {
         //   (https://github.com/bintray/gradle-bintray-plugin/issues/325)
         // - Could not upload to 'https://api.bintray.com/...':
         //   HTTP/1.1 405 Not Allowed 405 Not Allowed405 Not Allowednginx
-        override = false
+        override = true
```
eugene-krivobokov: Are you sure to make it default?
It can override the previous release and break all clients.

dsvoronin: nope, thanks

subprojects/android-test/test-report/src/main/kotlin/com/avito/android/test/report/ArgsProvider.kt
dsvoronin: i think it is a bit more safe that way

from bundle.getInt
```
 catch (ClassCastException e) {
            typeWarning(key, o, "Integer", defaultValue, e);
            return defaultValue;
        }
```

and no default value means default is 0, so we will get 0 result if pass something not convertable to int

subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/runner/InHouseInstrumentationTestRunner.kt
dsvoronin: i reverted to warning, first look was: yes, it's absolutely critical, but then i remember all this multithread calls to onException, and that we will get "false-criticals" here

eugene-krivobokov (approved):

PR #693, opened 2020-12-23
MBS-10013 Validate tests state before running

eugene-krivobokov:
## Context

Sometimes we get flaky component tests because of test rules with interactions with external systems: resource manager, real network.

## Changes

Added a contract between the type of test and available test rules.
The validation is on two levels: 

1. Inside test rule in client code

```kotlin
class MyRule : SimpleRule() {

  override fun validatePreconditions() {
        requireTestType(Kind.E2E)
}
```

2. On `AndroidJUnitRunner` level when we don't have a suitable test rule.
At this level, all test rules are already validated and prepared for the test application.

A user adds desired checks:

```kotlin
BaseAvitoInstrumentTestRunner {

override val testMetadataValidator = CompositeTestMetadataValidator(validators = listOf(
        NetworkIsMockedValidator()
    ))
```

PR #706, opened 2021-01-13
Remove excessive test rule method

eugene-krivobokov:
The method was added in #693. The idea was to force check in all test rules.

```kotlin
class MyRule : SimpleRule() {

  override fun validatePreconditions() {
        requireTestType(Kind.E2E)
  }
  ```

However, I found it excessive. There are only two cases of check in dozens of test rules. Forcing implementation doesn't help noticeably.

```kotlin
class MyRule : SimpleRule() {

  override fun validatePreconditions() {
        // no requirements
  }
  ```
I think that `before` will be enough and serve the same purpose:

```kotlin
class MyRule : SimpleRule() {

  override fun before() {
        requireTestType(Kind.E2E)
  }
  ```


dsvoronin (approved):
I agree, LGTM

PR #709, opened 2021-01-14
Logger leftovers

dsvoronin:
see commit messages for changes details


subprojects/common/elastic/src/main/kotlin/com/avito/android/elastic/ElasticClientFactory.kt
```diff
@@ -0,0 +1,43 @@
...
+import com.avito.time.DefaultTimeProvider
+import com.avito.time.TimeProvider
+import okhttp3.OkHttpClient
+import java.util.concurrent.TimeUnit
+
+object ElasticClientFactory {
+
+    private const val defaultTimeoutSec = 10L
+
+    private val timeProvider: TimeProvider by lazy { DefaultTimeProvider() }
```
eugene-krivobokov: Is it so expensive to make it lazy?

subprojects/common/time/src/main/kotlin/com/avito/time/DefaultTimeProvider.kt
```diff
@@ -12,9 +14,17 @@ import java.util.concurrent.TimeUnit
...
 
-    override val timeZone: TimeZone = TimeZone.getTimeZone("Europe/Moscow")
+    private val logger = loggerFactory.create<DefaultTimeProvider>()
+
+    private val locale: Locale = Locale.getDefault()
+
+    override val timeZone: TimeZone = TimeZone.getDefault()
+
+    init {
+        logger.debug("Creating new time provider: Locale = $locale; TimeZone = $timeZone")
```
eugene-krivobokov, line 26: Even debug seems too verbose. Time provider is a lightweight interface about standard classes: TimeZone, Locale, DateFormat, ... Don't see reasons to log it.

subprojects/common/okhttp/src/main/kotlin/com/avito/http/RetryInterceptor.kt
```diff
@@ -42,11 +44,17 @@ class RetryInterceptor constructor(
 
             try {
                 prepareForRetry(response)
+                val request = if (tryCount > 1) {
```
eugene-krivobokov, line 47: Do we modify only the second retry? Why?

subprojects/detekt.yml
```diff
@@ -675,8 +675,10 @@ style:
     RedundantHigherOrderMapUsage:
         active: false
     # https://detekt.github.io/detekt/style.html#redundantvisibilitymodifierrule
+    # todo don't know about kotlin strict mode
+    # possibly fixed in 1.15 https://github.com/detekt/detekt/issues/3125
```
eugene-krivobokov: Can we just update to 1.15?

subprojects/common/time/src/main/kotlin/com/avito/time/DefaultTimeProvider.kt
dsvoronin: It's in fact crucial for this PR; i want to log defaults for locale and timezone for all environments: gradle, emulators

subprojects/common/okhttp/src/main/kotlin/com/avito/http/RetryInterceptor.kt
dsvoronin: on second try; we don't modify only original request processing

subprojects/detekt.yml
dsvoronin: yep, i thought it would be better in separate pr but whatever

subprojects/common/elastic/src/main/kotlin/com/avito/android/elastic/ElasticClientFactory.kt
dsvoronin: nope, in fact it maybe more expensive to make it lazy here; i'll fix it

subprojects/detekt.yml
dsvoronin: fix in 1.15 https://github.com/detekt/detekt/issues/3125 only works per module, not in our detektAll task
because of how strict api detection works

subprojects/common/time/src/main/kotlin/com/avito/time/DefaultTimeProvider.kt
eugene-krivobokov: > i want to log defaults for locale and timezone for all environments

Ok, that seems reasonable. I propose to do it one time while starting the environment. 
I found it excessive to log the environment on each creation of this simple util.

The reason is simple. If we want to check the environment, it will be good to do it explicitly and as early as possible.

subprojects/common/okhttp/src/main/kotlin/com/avito/http/RetryInterceptor.kt
eugene-krivobokov: Sorry, missed that we do all `proceed` in the loop.

eugene-krivobokov (approved):

PR #710, opened 2021-01-14
update android-builder image

dsvoronin:
key changes:
 - build-tools;30.0.3 (kept 29.0.3 for migration)
 - updated commandline tools (installs in another location, fixed package_file parameter, installs patcher and platform-tools by default)
 - no need to pass ENV GRADLE_USER_HOME anymore

https://hub.docker.com/layers/avitotech/android-builder/334a3381e7/images/sha256-c9582e9f344a38d047e87bf6d3709c91c8802ae9e6d0e2aba674c1bdd5320fcf?context=explore

### DIFF
--- a/ci/docker/android-builder/packages.txt
+++ b/ci/docker/android-builder/packages.txt
@@ -1,0 +1,6 @@
+build-tools;30.0.3
+build-tools;29.0.3
+platforms;android-29
+tools
+extras;google;google_play_services
+extras;google;m2repository
--- a/_main.sh
+++ b/_main.sh
@@ -2,7 +2,7 @@
 
 set -e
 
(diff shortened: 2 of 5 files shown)


ci/docker/android-builder/packages.txt
```diff
@@ -0,0 +1,6 @@
+build-tools;30.0.3
```
eugene-krivobokov, line 1: Do we have plans to migrate to 30 in this quarter?

eugene-krivobokov (approved):

ci/docker/android-builder/packages.txt
dsvoronin: no i don't think so, 
but we could try to use higher build-tools version for building 29 api

PR #713, opened 2021-01-15
MBS-10348 filter our non-kotlin files for kotlin class analysis

dsvoronin:
fails on AndroidManifest.xml changes
right now changed tests finder is oversimplified and doesn't take many things into account
it's main purpose is to filter changed tests to run on PR with more strict retry rules (like 3 of 3 success instead of 1 of 2 success)

changes in manifest can lead to new problems in tests, but more likely even 1 of 2 will fail on these changes

changed retry policy runs on top of standard one, so no tests will be unintentionally skipped


subprojects/gradle/instrumentation-changed-tests-finder/src/main/kotlin/com/avito/android/FindChangedTestsAction.kt
```diff
@@ -45,6 +45,7 @@ abstract class FindChangedTestsAction : WorkAction<FindChangedTestsAction.Params
         ).map { changedFiles ->
             changedFiles.asSequence()
                 .filter { it.changeType in changeTypes }
+                .filter { it.file.extension == KotlinClassesFinderImpl.KOTLIN_FILE_EXTENSION }
```
eugene-krivobokov, line 48: What about integration tests?

eugene-krivobokov (approved):

subprojects/gradle/instrumentation-changed-tests-finder/src/main/kotlin/com/avito/android/FindChangedTestsAction.kt
dsvoronin: Yep, need to add some

PR #714, opened 2021-01-15
Mbs 10311 runner metrics

dsvoronin:
Added some basic metrics for test runner:

- initial delay
- end delay
- median tests queue time (from test suite started to moment test claimed a device)
- median installation time (from moment test claimed a device to actual test start)
- suite time
- total time (suite time with both delays)

(no execution time for now)

image


subprojects/gradle/runner/service/src/main/kotlin/com/avito/runner/service/listener/TestMetricsSender.kt
```diff
@@ -0,0 +1,66 @@
...
+import com.avito.android.stats.GaugeMetric
+import com.avito.android.stats.StatsDSender
+import com.avito.logger.LoggerFactory
+import com.avito.logger.create
+import com.avito.runner.service.model.TestCase
+import com.avito.runner.service.model.TestCaseRun
+import com.avito.runner.service.worker.device.Device
+import com.avito.time.TimeProvider
+
+class TestMetricsSender(
```
eugene-krivobokov:
```suggestion
internal class TestMetricsSender(
```
?

subprojects/gradle/runner/service/src/main/kotlin/com/avito/runner/service/listener/TestMetricsSender.kt
```diff
@@ -0,0 +1,66 @@
...
+
+import com.avito.android.stats.GaugeMetric
+import com.avito.android.stats.StatsDSender
+import com.avito.logger.LoggerFactory
+import com.avito.logger.create
+import com.avito.runner.service.model.TestCase
+import com.avito.runner.service.model.TestCaseRun
+import com.avito.runner.service.worker.device.Device
+import com.avito.time.TimeProvider
+
```
eugene-krivobokov: Describe please the concept. It's harder to get the idea from implementation details.

subprojects/gradle/runner/service/src/main/kotlin/com/avito/runner/service/listener/TestMetricsSender.kt
```diff
@@ -0,0 +1,66 @@
...
+        result: TestCaseRun.Result,
+        durationMilliseconds: Long,
+        executionNumber: Int
+    ) {
+        val key = Key(test, executionNumber)
+
+        val startedMark = startedMarks[key]
+
+        if (startedMark != null) {
+            statsDSender.send(prefix, GaugeMetric("execution", timeProvider.nowInMillis() - startedMark))
```
eugene-krivobokov: How many unique metrics and events on each PR do you expect?
I'm worried about the workload.

dsvoronin: it's 2 metrics for each test run (with retries), so about 6000 per PR; seems ok

subprojects/gradle/runner/service/src/main/kotlin/com/avito/runner/service/listener/TestMetricsSender.kt
dsvoronin: yep, thanks; time to make module explicit )

subprojects/gradle/runner/service/src/main/kotlin/com/avito/runner/service/listener/TestMetricsSender.kt
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    override fun started(device: Device, targetPackage: String, test: TestCase, executionNumber: Int) {
+        val key = Key(test, executionNumber)
+        startedMarks[key] = timeProvider.nowInMillis()
+
+        val intendedMark = intendedMarks[key]
+
+        if (intendedMark != null) {
+            statsDSender.send(prefix, GaugeMetric("queue", timeProvider.nowInMillis() - intendedMark))
```
eugene-krivobokov: I didn't get what do we want to collect here.
It sends the same gauge metric `runner.test.duration.queue` on each execution. 
If I'm not mistaken, only the last value within a bucket (30 sec by default) will be preserved.
https://statsd.readthedocs.io/en/v3.2.1/types.html#gauges

dsvoronin: yep, i added test coordinates to the metric

subprojects/gradle/runner/service/src/main/kotlin/com/avito/runner/service/listener/TestMetricsSender.kt
dsvoronin: I tried and it becomes implementation details ))

- `DeviceWorker` calls `TestListener.intended()`
- `TestMetrics` saves this `intended` timestamp by key (testcase, executionumber)
- DeviceWorker` calls `TestListener.started()`
- `TestMetrics` saves this `started` timestamp by key (testcase, executionumber)
- `TestMetrics` sends delta between now and intended
- DeviceWorker` calls `TestListener.finished()`
- `TestMetrics` sends delta between now and started

subprojects/gradle/runner/service/src/main/kotlin/com/avito/runner/service/listener/TestListener.kt
```diff
@@ -6,6 +6,12 @@ import com.avito.runner.service.worker.device.Device
 
 interface TestListener {
 
+    fun intended(
```
sergioboi: Let's add KDoc because intended is unclear by itself

subprojects/gradle/runner/service/src/main/kotlin/com/avito/runner/service/listener/TestMetricsSender.kt
```diff
@@ -0,0 +1,81 @@
...
+import com.avito.runner.service.worker.device.Device
+import com.avito.time.TimeProvider
+
+internal class TestMetricsSender(
+    private val statsDSender: StatsDSender,
+    private val timeProvider: TimeProvider,
+    loggerFactory: LoggerFactory
+) : TestListener {
+
+    private val prefix = "runner.test.duration"
```
sergioboi: Describe full names for all metrics

subprojects/gradle/runner/service/src/main/kotlin/com/avito/runner/service/listener/TestMetricsSender.kt
```diff
@@ -0,0 +1,81 @@
...
+                    path = keyMetric("execution", key),
+                    value = timeProvider.nowInMillis() - startedMark
+                )
+            )
+        } else {
+            logger.warn("Can't send test metric: started mark not found")
+        }
+    }
+
+    private fun keyMetric(name: String, key: Key): String =
```
sergioboi: What is the name? Prefer to see metric as Enum

subprojects/gradle/runner/service/src/main/kotlin/com/avito/runner/service/listener/TestMetricsSender.kt
```diff
@@ -0,0 +1,81 @@
...
+        "${key.test.deviceName}.${key.executionNumber}.${key.test.testName}.$name"
```
sboishtyan, line 76: Let's make the format "testName.deviceName.executionNumber.metricName"

sboishtyan: As I said it will produce a lot of unique metrics and it will load graphite a lot.

As I understand we don't need specific test data we need only average time for all tests

Let's simplify metric sending and remove testName, device and execution number and Statsd will produce a unique point for each 30s

subprojects/gradle/runner/service/src/main/kotlin/com/avito/runner/service/listener/TestMetricsSender.kt
```diff
@@ -0,0 +1,82 @@
...
```
sergioboi: As I said it will produce a lot of unique metrics and it will load graphite a lot.

As I understand we don't need specific test data we need only average time for all tests

Let's simplify metric sending and remove testName, device and execution number and Statsd will produce a unique point for each 30s

sergioboi (changes requested):

subprojects/gradle/runner/service/src/main/kotlin/com/avito/runner/service/listener/TestMetricsSender.kt
eugene-krivobokov: Thanks, I meant only part about metrics layout.

subprojects/gradle/runner/service/src/main/kotlin/com/avito/runner/service/listener/TestMetricsSender.kt
eugene-krivobokov: Screenshot from 2021-01-15 20-21-36

We've discussed the main idea.
Detalization is a subject of discussion.

subprojects/gradle/runner/service/src/main/kotlin/com/avito/runner/service/listener/TestMetricsSender.kt
dsvoronin: yep i agree, let's compute all on client

subprojects/gradle/runner/client/src/main/kotlin/com/avito/runner/scheduler/listener/ArtifactsTestListener.kt
```diff
@@ -10,13 +10,21 @@ import java.io.File
...
 
-class ArtifactsTestListener(
+internal class ArtifactsTestListener(
     private val lifecycleListener: TestLifecycleListener,
     loggerFactory: LoggerFactory
 ) : TestListener {
 
     private val logger = loggerFactory.create<ArtifactsTestListener>()
 
+    override fun onDevice(
```
eugene-krivobokov, line 20: From this interface, the lifecycle is not clear.
I'd good to describe it.

subprojects/gradle/runner/client/src/main/kotlin/com/avito/runner/scheduler/metrics/TestKey.kt
```diff
@@ -0,0 +1,8 @@
+package com.avito.runner.scheduler.metrics
+
+import com.avito.runner.service.model.TestCase
+
+internal data class TestKey(
```
eugene-krivobokov: The "key" looks like a usage case. What if I don't use it as a key?
Maybe it's better to express the content itself: `TestCaseExecution` or similar.

subprojects/gradle/runner/client/src/main/kotlin/com/avito/runner/scheduler/metrics/TestMetricsSender.kt
```diff
@@ -0,0 +1,55 @@
...
+import com.avito.android.stats.GaugeMetric
+import com.avito.android.stats.StatsDSender
+
+internal class TestMetricsSender(
+    private val statsDSender: StatsDSender,
+    buildId: String,
+    instrumentationConfigName: String
+) {
+
+    private val prefix = "testrunner.$buildId.$instrumentationConfigName"
```
eugene-krivobokov, line 12: > buildId

Are you sure that we need to make unique metrics for each build?

eugene-krivobokov: I see, we have gauge metrics. It helps to avoid overriding between builds.

subprojects/gradle/runner/client/src/main/kotlin/com/avito/runner/scheduler/metrics/TestMetricsSender.kt
```diff
@@ -0,0 +1,55 @@
+package com.avito.runner.scheduler.metrics
+
+import com.avito.android.stats.GaugeMetric
+import com.avito.android.stats.StatsDSender
+
+internal class TestMetricsSender(
```
eugene-krivobokov, line 6: I'd expect to see a description of metrics in the product documentation.

subprojects/gradle/runner/client/src/main/kotlin/com/avito/runner/scheduler/metrics/TestMetricsListenerImpl.kt
```diff
@@ -0,0 +1,97 @@
...
+            testSuiteEndedTime = timeProvider.nowInMillis(),
+            onDeviceMarks = onDeviceMarks,
+            startedMarks = startedMarks,
+            finishedMarks = finishedMarks
+        )
+
+        with(testMetricsSender) {
+            metricsComputer.computeInitialDelay()
+                ?.let { sendInitialDelay(it) }
+                ?: logger.warn("Not sending initial delay, no data")
```
eugene-krivobokov, line 73: When will it happen? On empty test suite? In such a case can we check it one time but not for all metrics?

subprojects/gradle/runner/client/src/main/kotlin/com/avito/runner/scheduler/metrics/TestMetricsListenerImpl.kt
```diff
@@ -0,0 +1,97 @@
...
+
+internal class TestMetricsListenerImpl(
+    private val testMetricsSender: TestMetricsSender,
+    private val timeProvider: TimeProvider,
+    private val loggerFactory: LoggerFactory
+) : TestMetricsListener {
+
+    private val logger = loggerFactory.create<TestMetricsListener>()
+
+    private val onDeviceMarks = ConcurrentHashMap<TestKey, Long>()
```
eugene-krivobokov: `TestKey, Long` - can it be a named model?
We duplicate this `onDeviceMarks: Map<TestKey, Long>` in two classes. Data shouldn't be described by property name, but rather by type.

subprojects/gradle/runner/client/src/main/kotlin/com/avito/runner/scheduler/metrics/TestMetricsComputer.kt
```diff
@@ -0,0 +1,69 @@
...
+            } else {
+                value - onDeviceMark
+            }
+        }
+        .filterNotNull()
+
+    /**
+     * null is possible if no tests
+     */
+    fun computeInitialDelay(): Long? = firstTestStarted?.let { it - testSuiteStartedTime }
```
eugene-krivobokov: I think we can delete `compute` from the interface.
These are not heavy computations and clients don't need to know it.
Also, the client knows about this from the class name.

```kotlin
computer.computeInitialDelay()

computer.initialDelay()
```

subprojects/gradle/runner/client/src/main/kotlin/com/avito/runner/scheduler/metrics/TestMetricsComputer.kt
```diff
@@ -0,0 +1,69 @@
+package com.avito.runner.scheduler.metrics
+
+import com.avito.logger.LoggerFactory
+import com.avito.logger.create
+
+internal class TestMetricsComputer(
+    loggerFactory: LoggerFactory,
+    private val testSuiteStartedTime: Long,
+    private val testSuiteEndedTime: Long,
+    private val onDeviceMarks: Map<TestKey, Long>,
```
eugene-krivobokov: https://github.com/avito-tech/avito-android/pull/714/files#r560060644

subprojects/gradle/runner/client/src/main/kotlin/com/avito/runner/scheduler/metrics/TestMetricsListenerImpl.kt
dsvoronin: i see that it leads to coupling (more data leaks from computation to sending)

what is the problem here? 
code is harder to read? or extra log messages?

eugene-krivobokov: As I understood, we duplicate checks and errors for the same reason - empty test suite.
This idea can be expressed by the code:

```kotlin
class TestSuiteExecutionProfile {
  val totalTime: Long
  val testsExecution: TestsExecution? // <--- one mark to describe the situation
  
  class TestsExecution(
     val initialDelay: Long,
     ...
     val endDelay: Long
  )
}

val profile: TestExecutionProfile = testMetricsComputer.compute()
```

Not a big deal, up to you

docs/content/test_runner/Metrics.md
```diff
@@ -0,0 +1,31 @@
+# Test runner metrics
+
+Metrics available at:
```
eugene-krivobokov, line 3: Maybe not in this PR but here must be a part about enabling and configuring metrics.

docs/content/test_runner/Metrics.md
```diff
@@ -0,0 +1,31 @@
...
+
+### `suite`
+
+Time from first test execution start to last test execution finished
+
+### `total`
+
+Total job time (suite time with both delays)
+
+![Metrics](https://user-images.githubusercontent.com/1105133/105228737-fb467f00-5b73-11eb-801a-da494182f431.png)
```
eugene-krivobokov, line 31: Is it possible to draw it in mermaid?
Rendered images are hard to change.

docs/mkdocs.yml
```diff
@@ -84,14 +84,16 @@ nav:
...
     -   Testing:
-            -   Test Runner: test/Runner.md
             -   Flaky annotation: test/FlakyAnnotation.md
             -   Toast: test/Toast.md
     -   Test framework:
             - test_framework/TestFramework.md
             -   Snackbar: test_framework/Snackbar.md
             -   Screen checks: test_framework/ScreenChecks.md
             -   Internals: test_framework/Internals.md
+    -   Test runner:
```
eugene-krivobokov, line 94: How about the `projects` page?
We used it for ready to use plugins

subprojects/gradle/runner/client/src/main/kotlin/com/avito/runner/scheduler/metrics/TestMetricsAggregatorImpl.kt
```diff
@@ -0,0 +1,59 @@
...
+
+    private val lastTestEnded = testTimestamps.values
+        .mapNotNull { it.finished }
+        .maxOrNull()
+
+    private val queueTimes: List<Long> = testTimestamps.values
+        .mapNotNull { it.onDevice }
+        .map { it - testSuiteStartedTime }
+
+    private val installationTimes: List<Long> = testTimestamps.values
```
eugene-krivobokov, line 25: Storing filtered values in properties seems premature optimization.
They are needed only one time and can be easily calculated in place or by request.

docs/mkdocs.yml
dsvoronin: I know, but it's not ready to use for now

subprojects/gradle/runner/client/src/main/kotlin/com/avito/runner/scheduler/metrics/TestMetricsAggregatorImpl.kt
dsvoronin: i can wrap it in `lazy {}`, but i wanted to extract this calculation into functions anyways, and why do it as a function if it's value doesn't change?

docs/content/test_runner/Metrics.md
dsvoronin: i couldn't find a sutable diagram type

docs/content/test_runner/Metrics.md
dsvoronin: i agree, but better in next PR

subprojects/common/math/src/main/kotlin/com/avito/math/Median.kt
```diff
@@ -0,0 +1,22 @@
...
+ * see https://stackoverflow.com/a/28822243/981330
+ */
+fun <N : Number> Sequence<N>.median(): Double {
+
+    val sorted = map { it.toDouble() }.sorted().toList()
+
+    require(sorted.isNotEmpty()) { "Cannot calculate median of no elements" }
+
+    return if (sorted.size % 2 == 0) {
+        (sorted[sorted.size / 2] + sorted[sorted.size / 2 - 1]) / 2
```
sergioboi, line 18: Sorry but `(sorted[sorted.size / 2] + sorted[sorted.size / 2 - 1])` equal to `sorted.size - 1`

subprojects/gradle/instrumentation-tests/src/integTest/kotlin/com/avito/instrumentation/executing/TestExecutorIntegrationTest.kt
```diff
@@ -73,6 +74,8 @@ internal class TestExecutorIntegrationTest {
     private fun createTestExecutor(
         loggerFactory: LoggerFactory,
         configurationName: String = "",
+        buildId: String = "",
```
sergioboi, line 77: prefer to see "integration-test-build-id"

subprojects/gradle/runner/client/src/main/kotlin/com/avito/runner/scheduler/metrics/TestMetricsAggregator.kt
```diff
@@ -0,0 +1,16 @@
+package com.avito.runner.scheduler.metrics
+
+internal interface TestMetricsAggregator {
+
+    fun initialDelay(): Long?
```
sergioboi, line 5: Why they could be nullable?

subprojects/gradle/runner/client/src/main/kotlin/com/avito/runner/scheduler/metrics/TestMetricsAggregatorImpl.kt
eugene-krivobokov: Of course, in this class with tests, we shouldn't see a difference. I'm talking about the general approach.

The more state we have, the more complexity we add.
Storing something is needed only for caching when a computation is long. It is in fact a performance technique.
We consume more memory for faster access to results.
I prefer to see caching when we have reasons to think that calculation in place / by request is not enough.

subprojects/common/math/src/main/kotlin/com/avito/math/Median.kt
```diff
@@ -0,0 +1,22 @@
+package com.avito.math
```
eugene-krivobokov, line 1: I've seen an external library in previous commits?
I think this is totally ok for well-known issues.

subprojects/gradle/runner/client/src/main/kotlin/com/avito/runner/scheduler/metrics/TestMetricsAggregatorImpl.kt
dsvoronin: i see, i believe it make sense here, but i agree in general

subprojects/common/math/src/main/kotlin/com/avito/math/Median.kt
dsvoronin: no, it's not )

subprojects/gradle/runner/client/src/main/kotlin/com/avito/runner/scheduler/metrics/TestMetricsAggregator.kt
dsvoronin: could be a lot of reasons:
- something it broken in runner logic and some listener calls not invoked
- empty test suite
- maybe some case in integration tests

It will be very hard to troubleshoot with error message from metrics sender, so i prefer defensive approach here, but log a warning that metrics was not sended

subprojects/common/math/src/main/kotlin/com/avito/math/Median.kt
dsvoronin: i removed external library and added inhouse implementation of median

subprojects/gradle/runner/client/src/main/kotlin/com/avito/runner/scheduler/metrics/TestMetricsAggregator.kt
sergioboi: ok

subprojects/common/math/src/main/kotlin/com/avito/math/Median.kt
sergioboi: I'm dumb

subprojects/common/math/src/main/kotlin/com/avito/math/Median.kt
eugene-krivobokov: It was a small and tested library.
Why do we want an own implementation?

eugene-krivobokov (approved):

subprojects/common/math/src/main/kotlin/com/avito/math/Median.kt
dsvoronin: you mean kotlin-statistics?
it is mostly a wrapper on top of commons-math, which jar is > 2mb

sergioboi (approved):

PR #718, opened 2021-01-20
MBS-9285 use system properties to align versions

dsvoronin:
use system properties to align versions between plugins and dependencies for kotlin, agp and detekt

use kotlin bom to achieve same version for all kotlin dependencies

use forUseAtConfigurationTime to read system properties and not break configuration cache

### DIFF
--- a/subprojects/gradle.properties
+++ b/subprojects/gradle.properties
@@ -18,3 +18,7 @@ android.useAndroidX=true
 projectVersion=2021.3
 # Disable console output https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/issues/202
 systemProp.dependency.analysis.silent=true
+# Versions
+systemProp.kotlinVersion=1.4.21
+systemProp.detektVersion=1.15.0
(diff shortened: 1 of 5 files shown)


subprojects/gradle.properties
```diff
@@ -18,3 +18,7 @@ android.useAndroidX=true
 projectVersion=2021.3
 # Disable console output https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/issues/202
 systemProp.dependency.analysis.silent=true
+# Versions
+systemProp.kotlinVersion=1.4.21
+systemProp.detektVersion=1.15.0
+systemProp.androidGradlePluginVersion=4.1.1
```
eugene-krivobokov, line 24: Can't we use project properties instead of system properties? Are there any restrictions?

eugene-krivobokov (approved):

subprojects/gradle.properties
dsvoronin: don't know, i just reverted it back

PR #719, opened 2021-01-20
Mbs 10393 elastic time fix

dsvoronin:
send to elastic without timezone and with default locale
added an additional checker to find locale problem: DateFormatChecker

send via slf4j if elastic disabled


subprojects/common/elastic/src/main/kotlin/com/avito/android/elastic/DateFormatCheckerImpl.kt
```diff
@@ -0,0 +1,21 @@
+package com.avito.android.elastic
+
+import java.util.Locale
+
+internal class DateFormatCheckerImpl : DateFormatChecker {
```
eugene-krivobokov:
```suggestion
internal class ElasticDateFormatChecker : DateFormatChecker {
```
`Impl` -> `Elastic` to express the purpose

subprojects/common/elastic/src/main/kotlin/com/avito/android/elastic/DateFormatCheckerImpl.kt
```diff
@@ -0,0 +1,21 @@
...
+
+    private val datePattern = Regex("[0-9]{4}-[0-9]{2}-[0-9]{2}")
+
+    private val dateCheckCache = mutableSetOf<String>()
+
+    override fun check(formattedDate: String) {
+        if (!dateCheckCache.contains(formattedDate)) {
```
eugene-krivobokov: It seems to me like premature optimization.

Why do you think that this regex will be a problem? Do we have proofs?
Storing dates for each event can bloat memory.

subprojects/common/elastic/src/main/kotlin/com/avito/android/elastic/HttpElasticClient.kt
```diff
@@ -60,9 +61,13 @@ internal class HttpElasticClient(
 
             params.putAll(metadata)
 
+            val formattedDate = isoDate.format(now)
+
+            dateFormatChecker.check(formattedDate)
```
eugene-krivobokov, line 70: We've discussed feature toggles for new temporary fail-fast checks
up to you

subprojects/common/time/src/main/kotlin/com/avito/time/DefaultTimeProvider.kt
```diff
@@ -50,9 +40,7 @@ class DefaultTimeProvider(loggerFactory: LoggerFactory) : TimeProvider {
             cal1.get(Calendar.DAY_OF_YEAR) == cal2.get(Calendar.DAY_OF_YEAR)
     }
 
-    override fun formatter(pattern: String): DateFormat {
-        return SimpleDateFormat(pattern, locale).apply {
-            timeZone = this@DefaultTimeProvider.timeZone
-        }
+    override fun utcFormatter(pattern: String): DateFormat = SimpleDateFormat(pattern, locale).apply {
+        timeZone = TimeZone.getTimeZone("UTC")
     }
```
eugene-krivobokov: I propose to remove this function from TimeProvider.
I don't see reasons to have it TimeProvider's interface.

We can create SimpleDateFormat anywhere, it's a simple class:

 ```kotlin
val utcFormat: DateFormat = SimpleDateFormat(pattern, Locale.getDefault()).apply {
        timeZone = TimeZone.getTimeZone("UTC")
    }
 ```

subprojects/common/time/src/main/kotlin/com/avito/time/DefaultTimeProvider.kt
```diff
@@ -12,20 +10,12 @@ import java.util.concurrent.TimeUnit
...
 
     private val locale: Locale = Locale.getDefault()
 
-    override val timeZone: TimeZone = TimeZone.getDefault()
-
-    init {
-        logger.debug("Creating new time provider: Locale = $locale; TimeZone = $timeZone")
-    }
-
     override fun isSameDay(date1: Date, date2: Date): Boolean {
```
eugene-krivobokov: Another example. It seems like an extension function for Date:

```kotlin
date1.isSameDay(data2)
```

I don't see reasons to have this function in a big interface and pass it to all usages.

subprojects/common/elastic/src/main/kotlin/com/avito/android/elastic/HttpElasticClient.kt
```diff
@@ -60,9 +61,13 @@ internal class HttpElasticClient(
...
+
             elasticApi.log(
                 indexPattern = indexPattern,
-                date = isoDate.format(nowWithTimeZone),
+                date = formattedDate,
```
sergioboi, line 74: sad that elastic can't accept a date in Long format

subprojects/common/elastic/src/main/kotlin/com/avito/android/elastic/DateFormatCheckerImpl.kt
dsvoronin: this date must have a `yyyy-MM-dd` format
i see some invalid format evidence:

```
2021.1.20
2021.01.0020
2021.0001.20
2021.0001.0020
```

Don't know it could happen, so this check is to find bad locale of something

i don't want to calcualte regex match on every log call, so something like `2020.01.20` will be in cache for whole build

subprojects/common/elastic/src/main/kotlin/com/avito/android/elastic/HttpElasticClient.kt
dsvoronin: it is index in url path

subprojects/common/time/src/main/kotlin/com/avito/time/DefaultTimeProvider.kt
dsvoronin: fixed

subprojects/common/time/src/main/kotlin/com/avito/time/DefaultTimeProvider.kt
dsvoronin: fixed

subprojects/common/elastic/src/main/kotlin/com/avito/android/elastic/HttpElasticClient.kt
dsvoronin: flag added

eugene-krivobokov (approved):

PR #721, opened 2021-01-21
MBS-10416 Remove BOM dependencies from plugins

eugene-krivobokov:
They broke plugins resolution on clients
Will investigate reasons and possible workarounds additionally.
This PR is a hotfix.

### DIFF
--- a/buildscript/src/main/kotlin/Dependencies.kt
+++ b/buildscript/src/main/kotlin/Dependencies.kt
@@ -3,6 +3,7 @@
 object Dependencies {
 
     object Versions {
+        const val okhttp = "4.9.0"
         const val sentry = "1.7.23"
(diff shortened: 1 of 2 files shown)


subprojects/gradle/cd/build.gradle.kts
```diff
@@ -36,6 +36,13 @@ dependencies {
     testImplementation(project(":gradle:test-project"))
 }
 
+configurations.all {
+    resolutionStrategy {
+        // docker client тянет 4.x версию
+        force(Dependencies.okhttp)
```
dsvoronin: i think it is not necessary even without bom

eugene-krivobokov: Removed

dsvoronin (approved):

PR #723, opened 2021-01-22
MBS-9088 configuration cache preparations

dsvoronin:
Fix some blockers
Add tests for configuration cache compatibility
Most of them ignored with blocker on AGP 4.2 #725

### DIFF
--- a/subprojects/gradle/artifactory-app-backup/src/main/kotlin/com/avito/android/plugin/artifactory/ArtifactoryAppBackupPlugin.kt
+++ b/subprojects/gradle/artifactory-app-backup/src/main/kotlin/com/avito/android/plugin/artifactory/ArtifactoryAppBackupPlugin.kt
@@ -40,7 +40,7 @@ class ArtifactoryAppBackupPlugin : Plugin<Project> {
         backup: Backup
     ): PublishToMavenRepository {
         return tasks.withType(PublishToMavenRepository::class.java).matching {
-            it.repository.name == artifactoryRepositoryName
+            it.repository.name == artifactoryRepositoryName &&
             it.publication.name == backup.name
         }.first()
     }
--- a/subprojects/gradle/build-metrics/src/test/kotlin/com/avito/android/plugin/build_metrics/ConfigurationCacheCompatibilityTest.kt
+++ b/subprojects/gradle/build-metrics/src/test/kotlin/com/avito/android/plugin/build_metrics/ConfigurationCacheCompatibilityTest.kt
@@ -1,0 +1,37 @@
+package com.avito.android.plugin.build_metrics
+
+import com.avito.test.gradle.TestProjectGenerator
+import com.avito.test.gradle.TestResult
+import com.avito.test.gradle.gradlew
+import com.avito.test.gradle.module.AndroidAppModule
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.io.TempDir
+import java.io.File
+
+internal class ConfigurationCacheCompatibilityTest {
+
+    @Test
+    fun `configuration with applied plugin - ok`(@TempDir projectDir: File) {
+        TestProjectGenerator(
+            name = "rootapp",
+            plugins = listOf("com.avito.android.build-metrics"),
+            modules = listOf(
+                AndroidAppModule(name = "app")
+            )
+        ).generateIn(projectDir)
+
+        runTask(projectDir).assertThat().buildSuccessful()
+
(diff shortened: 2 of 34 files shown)


subprojects/gradle/artifactory-app-backup/src/main/kotlin/com/avito/android/plugin/artifactory/ArtifactoryAppBackupPlugin.kt
```diff
@@ -40,7 +40,7 @@ class ArtifactoryAppBackupPlugin : Plugin<Project> {
         backup: Backup
     ): PublishToMavenRepository {
         return tasks.withType(PublishToMavenRepository::class.java).matching {
-            it.repository.name == artifactoryRepositoryName
+            it.repository.name == artifactoryRepositoryName &&
```
eugene-krivobokov, line 43: Ooops :facepalm:

subprojects/gradle/build-metrics/src/test/kotlin/com/avito/android/plugin/build_metrics/ConfigurationCacheCompatibilityTest.kt
```diff
@@ -0,0 +1,42 @@
...
+            name = "rootapp",
+            plugins = listOf("com.avito.android.build-metrics"),
+            modules = listOf(
+                AndroidAppModule(name = "app")
+            )
+        ).generateIn(projectDir)
+
+        runTask(projectDir).assertThat().buildSuccessful()
+
+        runTask(projectDir).assertThat().buildSuccessful().configurationCachedReused()
```
eugene-krivobokov, line 30: We duplicate the same pattern of the test in all plugins.

```
ConfigurationCacheCompatibilityTest

<generate project> // <---

runTask(projectDir).assertThat().buildSuccessful()

runTask(projectDir).assertThat().buildSuccessful().configurationCachedReused()

<run task> // <---
```

Is it possible to reuse this contract and define only specific parts?
The issue is not in code duplication but rather in the common contract.
Another example: test for caching.

eugene-krivobokov (approved):

subprojects/gradle/build-metrics/src/test/kotlin/com/avito/android/plugin/build_metrics/ConfigurationCacheCompatibilityTest.kt
dsvoronin: o thought about inheritance of `@Test` annotation from abstract classes, but  junit5 doesn't support

eugene-krivobokov: One probable option: [junit 5 extensions](https://medium.com/decisionbrain/factor-test-code-the-junit-5-way-that-is-without-inheritance-9ba2784d34)
It looks overcomplicated for such simple cases without a complex shared state.

subprojects/gradle/design-screenshots/src/main/kotlin/com/avito/plugin/ScreenshotsPlugin.kt
```diff
@@ -10,15 +10,17 @@ class ScreenshotsPlugin : Plugin<Project> {
...
-                project.tasks.register<PullScreenshotsTask>("recordScreenshots") {
-                    group = "design"
-                    variant.set(applicationVariant)
-                    description = "Create and pull screenshots from device"
-                }
-                project.tasks.register<ClearScreenshotsTask>("clearScreenshots") {
-                    group = "design"
-                    variant.set(applicationVariant)
-                    description = "Clear screenshots on device"
+                if (applicationVariant.name != "release") {
```
sergioboi, line 13: Why do we skip the release application variant?

dsvoronin: this plugin was written without variants in mind, it fails on basic integration test because trying to register task with same name for every variant;
instead fixing it proper way (i don't know all applications so post poned that) i fixed it the way it's actually used (run manually for debug or stagingAutoTest) variant

eugene-krivobokov: @JaverSingleton , cc

PR #725, opened 2021-01-22
AGP 4.2 

dsvoronin:
https://developer.android.com/studio/releases/gradle-plugin#4-2-0

### Failing tests in signer plugin
```
Circular dependency between the following tasks:
:app:bundleRelease
+--- :app:produceReleaseBundleIdeListingFile
|    \--- :app:signBundleViaServiceRelease
|         \--- :app:bundleRelease (*)
\--- :app:signBundleViaServiceRelease (*)
```

### Failed to apply plugin 'com.android.internal.application' in multiple tests 

https://github.com/gradle/gradle/issues/16774

Using in memory project generator https://github.com/avito-tech/avito-android/blob/752fbc219b69352e6aa7a00f81126bfed3daadd9/subprojects/gradle/test-project/src/main/kotlin/com/avito/test/gradle/InMemoryProjectGenerator.kt

Could not create plugin of type 'AppPlugin'

Cannot create service of type DefaultBuildEventsListenerRegistry using DefaultBuildEventsListenerRegistry constructor as required service of type BuildEventListenerFactory for parameter #1 is not available.

### DIFF
--- a/libraries/src/main/kotlin/com/avito/android/ExternalLibrariesExtension.kt
+++ b/libraries/src/main/kotlin/com/avito/android/ExternalLibrariesExtension.kt
@@ -32,7 +32,7 @@ abstract class ExternalLibrariesExtension @Inject constructor(private val provid
      * We use exact version to provide consistent environment and avoid build cache issues
      * (AGP tasks has artifacts from build tools)
      */
-    val buildToolsVersion = "29.0.3"
+    val buildToolsVersion = "30.0.3"
 
     val androidGradlePluginVersion = systemProperty("androidGradlePluginVersion").get()
     val androidLintVersion = systemProperty("androidLintVersion").get()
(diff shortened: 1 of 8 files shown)


libraries/src/main/kotlin/com/avito/android/ExternalLibrariesExtension.kt
```diff
@@ -32,7 +32,7 @@ abstract class ExternalLibrariesExtension @Inject constructor(private val provid
      * We use exact version to provide consistent environment and avoid build cache issues
      * (AGP tasks has artifacts from build tools)
      */
-    val buildToolsVersion = "29.0.3"
+    val buildToolsVersion = "30.0.3"
```
eugene-krivobokov, line 35: Do we need to update the builder image accordingly?

dsvoronin: https://github.com/avito-tech/avito-android/blob/94959be2a2641b12bb603f8c1e125765193bbe9c/ci/docker/android-builder/packages.txt#L1-L2

It already supports both

dsvoronin:
Declined in favor of separate PR's

PR #729, opened 2021-01-22
MBS-7650 Fix test filters

sergioboi:


subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/TestsFilter.kt
```diff
@@ -9,19 +9,19 @@ import com.avito.android.test.annotations.UIComponentStub
 import com.avito.android.test.annotations.UIComponentTest
 
 private val manualAnnotations = setOf(
-    ManualTest::javaClass.name,
-    UIComponentStub::javaClass.name,
-    E2EStub::javaClass.name
+    ManualTest::class.java.name,
```
sergioboi, line 12: I don't know a reason.
I ran the instrumentationDynamic and see `javaClass` but the annotation class name
Screenshot 2021-01-22 at 21 15 29

And my hypothesis that `E2EStub::javaClass.name` works incorrectly without dependency on the kotlin-reflection

dsvoronin (approved):

PR #731, opened 2021-01-25
fix accessing system properties and it's nullability

dsvoronin:


subprojects/gradle/buildchecks/src/main/kotlin/com/avito/android/plugin/build_param_check/SystemValuesAccessor.kt
```diff
@@ -0,0 +1,47 @@
+package com.avito.android.plugin.build_param_check
+
+import org.gradle.api.provider.Provider
+import org.gradle.api.provider.ProviderFactory
+
+@Suppress("UnstableApiUsage")
+internal class SystemValuesAccessor(private val providers: ProviderFactory) {
```
eugene-krivobokov: This class has multiple responsibilities:
- Hides reading of system properties. It can be done without grouping them in one god interface.
- Hides formatting and default values. They are not just "system values", but formatted values for specific clients. E.g. `javaInfo`

Maybe renaming to something kinda "BuildEnvironmentInfo" will show an intention better.

eugene-krivobokov (approved):

subprojects/gradle/buildchecks/src/main/kotlin/com/avito/android/plugin/build_param_check/SystemValuesAccessor.kt
eugene-krivobokov: This is ok as the internal class for hotfix

dsvoronin: i renamed it

subprojects/gradle/buildchecks/src/main/kotlin/com/avito/android/plugin/build_param_check/BuildEnvironmentInfo.kt
```diff
@@ -0,0 +1,47 @@
...
+    private val javaVersion: String?
+        get() = getSystemProperty("java.version")
+
+    private val javaVendor: String?
+        get() = getSystemProperty("java.vendor")
+
+    val javaHome: String?
+        get() = getEnvironmentVariable("JAVA_HOME")
+
+    // todo it's avito-specific, how to get version generic way?
```
eugene-krivobokov, line 21: We can remove it. Don't see the value.

PR #733, opened 2021-01-26
MBS-9048 filter critical log level

dsvoronin:
- don't report every test error in AbstractLaunchRule
- don't report "Error while disconnecting UI Automation"
- report Can't register and report unexpected incident

### DIFF
--- a/subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/monitoring/CompositeTestIssuesMonitor.kt
+++ b/subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/monitoring/CompositeTestIssuesMonitor.kt
@@ -66,7 +66,7 @@ class CompositeTestIssuesMonitor(
         return EventBuilder()
             .withTag(
                 "test_name",
-                "${testRunEnvironment.testMetadata.className}.${testRunEnvironment.testMetadata.methodName}"
+                testRunEnvironment.testMetadata.testName
             )
             .withTag("build_id", testRunEnvironment.teamcityBuildId.toString())
             .withTag("build_branch", testRunEnvironment.buildBranch)
--- a/subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/runner/InHouseInstrumentationTestRunner.kt
+++ b/subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/runner/InHouseInstrumentationTestRunner.kt
@@ -81,9 +81,11 @@ abstract class InHouseInstrumentationTestRunner :
     val testRunEnvironment: TestRunEnvironment by lazy { createRunnerEnvironment(instrumentationArguments) }
 
     override val loggerFactory by lazy {
+        val runEnvironment = testRunEnvironment.asRunEnvironmentOrThrow()
         AndroidLoggerFactory(
             elasticConfig = elasticConfig,
-            sentryConfig = sentryConfig
+            sentryConfig = sentryConfig,
+            testName = runEnvironment.testMetadata.testName
         )
     }
 
@@ -310,7 +312,7 @@ abstract class InHouseInstrumentationTestRunner :
      */
     private fun initApplicationCrashHandling() {
         val currentHandler = Thread.getDefaultUncaughtExceptionHandler()
-        
+
         Thread.setDefaultUncaughtExceptionHandler(
             ReportUncaughtHandler(
                 loggerFactory = loggerFactory,
--- a/subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/runner/InHouseInstrumentationTestRunner.kt
+++ b/subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/runner/InHouseInstrumentationTestRunner.kt
@@ -292,7 +292,7 @@ abstract class InHouseInstrumentationTestRunner :
             report.registerIncident(AppCrashException(incident))
             report.reportTestCase()
         } catch (t: Throwable) {
-            logger.warn("Can't register and report unexpected incident", t)
+            logger.critical("Can't register and report unexpected incident", t)
         }
     }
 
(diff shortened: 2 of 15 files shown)


subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/runner/InHouseInstrumentationTestRunner.kt
```diff
@@ -292,7 +292,7 @@ abstract class InHouseInstrumentationTestRunner :
             report.registerIncident(AppCrashException(incident))
             report.reportTestCase()
         } catch (t: Throwable) {
-            logger.warn("Can't register and report unexpected incident", t)
+            logger.critical("Can't register and report unexpected incident", t)
```
sergioboi, line 297: I think an incident report fail is not critical because this is not lead to test failures

subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/test/AbstractLaunchRule.kt
```diff
@@ -35,7 +35,7 @@ abstract class AbstractLaunchRule : TestRule {
                 beforeTest()
                 base.evaluate()
             } catch (t: Throwable) {
-                logger.critical("Error during test completion", t)
+                logger.warn("Error during test completion", t)
```
sergioboi: We could delete this whole catch block. It looks useless because we throw an error further

subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/runner/ReportUncaughtHandler.kt
```diff
@@ -5,18 +5,19 @@ import com.avito.logger.create
...
-
-        InHouseInstrumentationTestRunner.instance.tryToReportUnexpectedIncident(
-            incident = e
-        )
-
-        globalExceptionHandler?.uncaughtException(t, e)
+        if (e.message in nonCriticalErrorMessages) {
+            logger.warn("Non critical error caught by ReportUncaughtHandler", e)
+        } else {
+            logger.critical("Application crash captured by global handler", e)
```
sergioboi:
```suggestion
            logger.critical("Application crashed", e)
```
We don't need info about `captured by` because it could be got by logger TAG

subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/runner/InHouseInstrumentationTestRunner.kt
dsvoronin: why so? i thought if we haven't report here test is probably lost

subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/test/AbstractLaunchRule.kt
dsvoronin: we don't want this test failures in kibana? report viewer is enough?

subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/runner/InHouseInstrumentationTestRunner.kt
sergioboi: I thought the incident means that test is ERROR. But If it's LOST behavior why we simple handle that better to avoid a LOST

subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/test/AbstractLaunchRule.kt
sergioboi: As I understand we will without that log because the test fails further

subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/runner/InHouseInstrumentationTestRunner.kt
dsvoronin: We're trying to make it an error, sending incident, but if something unexpected happens (e.g. this catch closure), we need to know about it and handle somehow

That's why this message promoted to critical;
plan is to look for this criticals and fix it's reasons

subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/test/AbstractLaunchRule.kt
dsvoronin: Can you please point me to another place where we log this error?

sergioboi: UncaughtReportHandler?

dsvoronin: If i understand this correctly this exception will be delivered as correct result with `finish()` function, not as an unexpected result

see:
https://github.com/android/android-test/blob/ec8526115902abec5ab0f66af0147874729080f4/runner/android_junit_runner/java/androidx/test/runner/AndroidJUnitRunner.java#L442

and:
https://github.com/android/android-test/blob/ec8526115902abec5ab0f66af0147874729080f4/runner/android_junit_runner/java/androidx/test/internal/runner/TestExecutor.java#L56

sergioboi: Ok
```suggestion
                logger.warn("Test execution failed", t)
```

subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/monitoring/CompositeTestIssuesMonitor.kt
```diff
@@ -66,7 +66,7 @@ class CompositeTestIssuesMonitor(
         return EventBuilder()
             .withTag(
                 "test_name",
```
sergioboi, line 68: What does `test_name` mean?

sergioboi: It's a key for value?

dsvoronin: `test_name` is a "className.$methodName" - unique test identifier
it's a key, yes

sergioboi (approved):

PR #739, opened 2021-01-27
add logger kdoc

dsvoronin:
based on https://github.com/avito-tech/avito-android/discussions/698

### DIFF
--- a/subprojects/common/logger/src/main/kotlin/com/avito/logger/Logger.kt
+++ b/subprojects/common/logger/src/main/kotlin/com/avito/logger/Logger.kt
@@ -1,8 +1,27 @@
 package com.avito.logger
 
+/**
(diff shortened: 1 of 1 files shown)


subprojects/common/logger/src/main/kotlin/com/avito/logger/Logger.kt
```diff
@@ -1,8 +1,27 @@
...
+     */
     fun info(msg: String)
+
+    /**
+     * unwanted state but can proceed
+     */
     fun warn(msg: String, error: Throwable? = null)
+
+    /**
+     * could not recover from this state
```
eugene-krivobokov, line 24:
```suggestion
     * Could not recover from this state
```
Sentences with capital letters are easier to read

sergioboi (approved):

PR #743, opened 2021-01-29
MBS-9048 Fix some lost test status cases

dsvoronin:


subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/runner/InHouseInstrumentationTestRunner.kt
```diff
@@ -291,8 +297,10 @@ abstract class InHouseInstrumentationTestRunner :
 
     fun tryToReportUnexpectedIncident(incident: Throwable) {
         try {
-            report.registerIncident(AppCrashException(incident))
-            report.reportTestCase()
+            if (!report.isWritten) {
+                report.registerIncident(AppCrashException(incident))
+                report.reportTestCase()
+            }
```
eugene-krivobokov, line 303: What else? Will we lost the incident?

subprojects/gradle/runner/client/src/main/kotlin/com/avito/runner/scheduler/listener/ArtifactsTestListener.kt
```diff
@@ -67,26 +68,33 @@ internal class ArtifactsTestListener(
...
 
     @Suppress("SdCardPath") // android API's are unavailable here
     private fun testMetadataFullDirectory(targetPackage: String, test: TestCase): File =
-        File("/sdcard/Android/data/$targetPackage/files/runner/${testMetadataDirectoryPath(test)}")
+        File("/sdcard/Android/data/$targetPackage/files/$RUNNER_OUTPUT_FOLDER/${testMetadataDirectoryPath(test)}")
 
     private fun testMetadataDirectoryPath(test: TestCase): String = "${test.className}#${test.methodName}"
+
+    companion object {
+        // todo should be passed with instrumentation params, see [ExternalStorageTransport]
```
eugene-krivobokov, line 89: Why does anyone need to customize this path?

subprojects/android-test/test-report/src/main/kotlin/com/avito/android/test/report/transport/OutputFileProvider.kt
```diff
@@ -0,0 +1,36 @@
+package com.avito.android.test.report.transport
+
+import androidx.test.platform.app.InstrumentationRegistry
+import java.io.File
+
+internal class OutputFileProvider {
+
+    sealed class Result {
```
eugene-krivobokov, line 8: `kotlin.Result` ?

dsvoronin: It's usage disabled bu default, i think we should decide it's usage outside of this PR context

subprojects/gradle/runner/client/src/main/kotlin/com/avito/runner/scheduler/listener/ArtifactsTestListener.kt
dsvoronin: it doesn't, but it's not clear if is it safe to change
you mean comment could be enough here?

subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/runner/InHouseInstrumentationTestRunner.kt
dsvoronin: It is basically any exceptions thrown after test report already written
these exceptions already logged in caller methods

subprojects/gradle/runner/client/src/main/kotlin/com/avito/runner/scheduler/listener/ArtifactsTestListener.kt
eugene-krivobokov: I see two options:
- Use value from constants. No one needs to customize it
- Client needs to customize the path

The question was about intention itself "should be passed with instrumentation params". 
Why we decided that this value should be customizable?

eugene-krivobokov (approved):

subprojects/gradle/runner/client/src/main/kotlin/com/avito/runner/scheduler/listener/ArtifactsTestListener.kt
dsvoronin: yep, i agree
these two modules don't have a dependency from each other, so such constants need to be placed somewhere, and that's the right way;
i'll try to find such place and anyway rephrase the comment

subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/runner/ReportUncaughtHandler.kt
```diff
@@ -15,7 +15,7 @@ internal class ReportUncaughtHandler(
         if (e.message in nonCriticalErrorMessages) {
             logger.warn("Non critical error caught by ReportUncaughtHandler", e)
         } else {
-            logger.critical("Application crashed", e)
+            logger.warn("Application crashed", e)
```
sergioboi, line 18: Is it common error?

subprojects/android-test/test-report/src/main/kotlin/com/avito/android/test/report/transport/ExternalStorageTransport.kt
```diff
@@ -1,67 +1,46 @@
...
 
-        return File(runnerDirectory, testFolderName).apply {
-            mkdirs()
+        when (outputFileResult) {
+            is Success ->
+                try {
+                    val json = gson.toJson(testRuntimeDataBuilder.fromState(state))
+                    val file = outputFileResult.file
+                    file.writeText(json)
+                    logger.info("Wrote report for test to file: ${file.absolutePath}")
```
sergioboi, line 37: Debug

subprojects/android-test/test-report/src/main/kotlin/com/avito/android/test/report/transport/ExternalStorageTransport.kt
```diff
@@ -1,67 +1,46 @@
...
+                } catch (e: Throwable) {
+                    logger.warn("Can't write report runtime data; leads to LOST test", e)
```
sergioboi, line 39: Could we write this status to the file system and then read from the runner for making decisions?

subprojects/android-test/test-report/src/main/kotlin/com/avito/android/test/report/transport/OutputFileProvider.kt
```diff
@@ -0,0 +1,36 @@
+package com.avito.android.test.report.transport
+
+import androidx.test.platform.app.InstrumentationRegistry
+import java.io.File
+
+internal class OutputFileProvider {
```
sergioboi, line 6: It's a too abstract name for the class which provides a `ReportFile`.

I want consistency. You could rename class or function ie:
- FileProvider and get(): Result<File>
- ReportProvider and get(): Result<Report>

subprojects/common/report-viewer/src/main/kotlin/com/avito/report/model/Incident.kt
```diff
@@ -11,9 +11,15 @@ data class Incident(
 ) {
     enum class Type {
 
+        /**
+         * abnormal test execution
```
sergioboi, line 15: What does `abnormal` mean?

subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/runner/ReportUncaughtHandler.kt
dsvoronin: Not really, but it will lead to reported test with `Error` status, which is not much different than any other runtime crash, which is treated more like a developer error, not an infrastructure one

subprojects/android-test/test-report/src/main/kotlin/com/avito/android/test/report/transport/ExternalStorageTransport.kt
dsvoronin: it is literally catch closure of file system write attempt
maybe it should be critical though

subprojects/android-test/test-report/src/main/kotlin/com/avito/android/test/report/transport/OutputFileProvider.kt
dsvoronin: yep, i agree; i'll add this comment to Report extraction story

subprojects/common/report-viewer/src/main/kotlin/com/avito/report/model/Incident.kt
dsvoronin: not normal :)

PR #744, opened 2021-01-29
MBS-10413 use artifactory proxy if available

dsvoronin:
using inhouse artifactory proxy for dependencies to improve stability

project also works without it (point to publically available repositories), but it is not tested in CI

gradle distribution downloads from public repo for now (not as simple to override it conditionally, and so far there was only one episode with it)

also update kaspresso to 1.2.0


subprojects/settings.gradle.kts
```diff
@@ -122,10 +122,29 @@ include(":ci:k8s-deployments-cleaner")
...
+
+    fun MavenArtifactRepository.setUrlOrProxy(repositoryName: String, originalRepo: String) {
+        if (artifactoryUrl.isNullOrBlank()) {
+            name = repositoryName
+            setUrl(originalRepo)
+        } else {
+            name = "Proxy for $repositoryName: $originalRepo"
+            artifactoryUrl(repositoryName)
+        }
+    }
```
eugene-krivobokov, line 140: Is there any way to reuse these functions?

subprojects/settings.gradle.kts
```diff
@@ -174,13 +195,32 @@ pluginManagement {
...
 
+val artifactoryUrl: String? by settings
+
+fun MavenArtifactRepository.artifactoryUrl(repositoryName: String) {
+    setUrl("$artifactoryUrl/$repositoryName")
+    isAllowInsecureProtocol = true
+}
+
+fun MavenArtifactRepository.setUrlOrProxy(repositoryName: String, originalRepo: String) {
+    if (artifactoryUrl.isNullOrBlank()) {
```
eugene-krivobokov, line 206: How should I know about this property?

Let's add it to the documentation. I see it as a part of "Contibuting / How to start"

eugene-krivobokov (approved):

subprojects/settings.gradle.kts
dsvoronin: NOPE :(

PR #745, opened 2021-01-29
MBS-10393 fix logger dateformatter concurrent problem

dsvoronin:
remove checker


subprojects/common/elastic/src/main/kotlin/com/avito/android/elastic/HttpElasticClient.kt
```diff
@@ -96,7 +93,16 @@ internal class HttpElasticClient(
...
     }
 
-    private fun utcFormatter(pattern: String): DateFormat = SimpleDateFormat(pattern, Locale.getDefault()).apply {
-        timeZone = TimeZone.getTimeZone("UTC")
+    /**
+     * from javadoc:
+     *
+     * It is recommended to create separate format instances for each thread.
+     * If multiple threads access a format concurrently, it must be synchronized
+     * externally.
```
eugene-krivobokov, line 101: Another good reason to use date-time API.
`java.time.format.TimeFormatter` is thread-safe.
We can use the original API in Gradle and desugar or backported version in Android code.

eugene-krivobokov (approved):

subprojects/common/elastic/src/main/kotlin/com/avito/android/elastic/HttpElasticClient.kt
dsvoronin: yep, but we have to evaluate desugaring impact on build times

subprojects/common/elastic/src/main/kotlin/com/avito/android/elastic/HttpElasticClient.kt
```diff
@@ -96,7 +93,16 @@ internal class HttpElasticClient(
...
+     */
+    private fun utcFormatter(pattern: String) = object : ThreadLocal<DateFormat>() {
```
sergioboi, line 103: ThreadLocal.withInitial { }

dsvoronin: This code used in ui tests env on 22 API emulators:
`Call requires API level 26 (current min is 21): java.lang.ThreadLocal#withInitial`
and it is not supported for now even with desugaring: https://developer.android.com/studio/write/java8-support-table

PR #757, opened 2021-02-05
MBS-10560 remove test issues monitor and http tracker

dsvoronin:
because it duplicates new logger work, that also uses sentry and kibana to deliver warnings and errors

httptracker was not used lately


subprojects/android-test/test-instrumentation-runner/src/main/kotlin/com/avito/android/runner/OrchestratorDelegate.kt
```diff
@@ -0,0 +1,10 @@
+package com.avito.android.runner
+
+import android.os.Bundle
+
+interface OrchestratorDelegate {
```
eugene-krivobokov, line 5: I don't understand the difference in responsibility between this abstraction and `InstrumentationDelegate`.
Would you mind explaining that?

subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/runner/TestRunEnvironment.kt
```diff
@@ -96,75 +95,68 @@ sealed class TestRunEnvironment {
...
     apiUrlParameterKey: String = "unnecessaryUrl",
     mockWebServerUrl: String = "localhost",
-    argumentsProvider: ArgsProvider
+    argumentsProvider: ArgsProvider,
 ): TestRunEnvironment {
-
-    val isFakeOrchestratorRun =
-        argumentsProvider.getOptionalArgument(FAKE_ORCHESTRATOR_RUN_ARGUMENT) != null
-
-    return when (isFakeOrchestratorRun) {
```
sboishtyan, line 112: Why did you delete this check?

subprojects/android-test/test-instrumentation-runner/src/main/kotlin/com/avito/android/runner/OrchestratorDelegate.kt
dsvoronin: i just delegated logic around orchestrator from InstrumentationTestRunner code
InstrumentationDelegate is a logic delegated depending on orchestrator argument, that's why its separated

eugene-krivobokov: What I see:
`test-instrumentation-runner` module doesn't know about avito test runner.
`InstrumentationDelegate` is in this module. It is a common delegate for `AndroidJUnitRunner`
`OrchestratorDelegate` is also in this module.
I didn't get how to decide which delegate to use when I will need to extract something else.

Is it correct that `OrchestratorDelegate` should encapsulate knowledge about AndroidX test orchestrator?

eugene-krivobokov (approved):

subprojects/android-test/test-instrumentation-runner/src/main/kotlin/com/avito/android/runner/OrchestratorDelegate.kt
dsvoronin: yes; i see it like this

eugene-krivobokov: Ok, maybe naming it as `AndroidXOrchestratorDelegate` will show it more obviously.
It took time to realize it because "orchestration" for me is a general term.

PR #764, opened 2021-02-08
MBS-10571 introduce strongly typed API for Graphite series name

dsvoronin:
to encapsulate time series constraints into more predictable API


subprojects/common/statsd/src/test/kotlin/com/avito/android/stats/SeriesNameTest.kt
```diff
@@ -0,0 +1,77 @@
...
+
+    @Test
+    fun `create series name - keeps minus`() {
+        val result = SeriesName.create("one-a")
+
+        assertThat(result.toString()).isEqualTo("one-a")
+    }
+
+    @Test
+    fun `create series name - replaces unsupported symbol (percent) with underscore`() {
```
eugene-krivobokov, line 44: Does regex replaces all invalid consecutive symbols by one `_`?
If so, we can check it also in tests

eugene-krivobokov (approved):

sergioboi (approved):

subprojects/common/statsd/src/test/kotlin/com/avito/android/stats/SeriesNameTest.kt
dsvoronin: thanks, new testcase added

eugene-krivobokov: I think we can merge these cases into one:
`create series name - replaces unsupported symbols with an underscore`
It includes implicitly `replaces unsupported symbol with underscore`

PR #766, opened 2021-02-10
MBS-10469 Add type safe models for plugins in test project generator

eugene-krivobokov:
It's needed to customize versions and disable plugins in generated projects


subprojects/gradle/test-project/src/test/kotlin/com/avito/test/gradle/PluginsSpecTest.kt
```diff
@@ -0,0 +1,78 @@
+package com.avito.test.gradle
+
+import com.avito.test.gradle.plugin.plugins
+import com.google.common.truth.Truth.assertThat
+import org.junit.Test
+
+class PluginsSpecTest {
+
```
eugene-krivobokov, line 8: Use cases for the new model

dsvoronin (approved):
We are closer and closer to general purpose `gradle-poet`)

PR #771, opened 2021-02-11
MBS-10556 use testFixtures for :common:report-viewer module

dsvoronin:


eugene-krivobokov:
Are `testFixtures` working correctly in IDEA \ AS? I don't remember the exact issues about them.

eugene-krivobokov (approved):

dsvoronin:
> Are `testFixtures` working correctly in IDEA \ AS? I don't remember the exact issues about them.

so far no problems with non-android modules
also no problem with android module as consumer
android module as producer not supported yet and honestly no real need for now

there was a problems with AS sync in presence of test fixtures in modules, looks like it was fixed

PR #773, opened 2021-02-11
Hide KuberentesClient usages by KubernetesApi interface

sergioboi:
This task is huge, so I need a write unit tests for reservation state management
So I decide to hide real KubernetesClient usages for easy stubing and faking


subprojects/gradle/runner/device-provider/src/main/kotlin/com/avito/android/runner/devices/internal/kubernetes/KubernetesApi.kt
```diff
@@ -0,0 +1,153 @@
...
+            try {
+                logger.debug("Deleting deployment: $deploymentName")
+                kubernetesClient.apps().deployments().withName(deploymentName).delete()
+                logger.debug("Deployment: $deploymentName deleted")
+            } catch (t: Throwable) {
+                logger.warn("Failed to delete deployment $deploymentName", t)
+            }
+        }
+
+        override suspend fun createDeployment(deployment: Deployment) {
```
dsvoronin, line 65: It better for testing to return value here and throw exception on another level

Same for other `return Unit` functions

dsvoronin (approved):

subprojects/gradle/runner/device-provider/src/main/kotlin/com/avito/android/runner/devices/internal/kubernetes/KubernetesApi.kt
```diff
@@ -0,0 +1,153 @@
...
+import io.fabric8.kubernetes.client.KubernetesClient
+
+internal interface KubernetesApi {
+    val namespace: String
+    suspend fun deletePod(podName: String): Boolean
+    suspend fun getPodLogs(podName: String): String
+    suspend fun getPod(podName: String): Pod?
+    suspend fun getPodDescription(podName: String): String
+    suspend fun deleteDeployment(deploymentName: String)
+    suspend fun createDeployment(deployment: Deployment)
```
eugene-krivobokov, line 17: This API exposes `io.fabric8.kubernetes` models. 
Do you want to hide them in the future or this abstraction is enough?

eugene-krivobokov:
LGTM

PR #774, opened 2021-02-12
Mbs 10605 use convention plugins

dsvoronin:
Used convention plugins instead of `subprojects {}` blocks.
Should speed up configuration with single eval->config pass per module,
but more importantly decouple project configuration with meaningful scripts (could be further improved)

Configure on demand not working for now (see comment in gradle.properties)

This will also be a working primer for my upcoming presentation on a real project, not another sample

Checklist:
- [x] publication
- [x] integration checks
- [x] colleagues checkout to check correct IDE behavior

Reference project: https://github.com/jjohannes/idiomatic-gradle


subprojects/gradle/upload-to-googleplay/build.gradle.kts
```diff
@@ -1,13 +1,13 @@
 plugins {
-    id("kotlin")
-    `maven-publish`
-    id("com.jfrog.bintray")
+    id("com.avito.android.kotlin-jvm")
+    id("com.avito.android.publish-kotlin-library")
+    id("com.avito.android.libraries")
 }
```
eugene-krivobokov, line 5: I'm thinking of common naming patterns for convention plugins:
- To distinguish easier plugins from buildSrc from "external" plugins
- To understand what it applies behind

As I understand, the package name `com.avito.android` is used in public repositories to avoid clashing. In convention plugins, we have more control and can omit it.

When I see `com.avito.android...` plugin it's not clear is it from a public repository or not. 

Possible options:
- `id(convention.kotlin-jvm)`
- `id(convention.com.android.library)`
- ...

What do you think?

build-logic/kotlin-convention/build.gradle.kts
```diff
@@ -0,0 +1,23 @@
+plugins {
+    `kotlin-dsl`
+    id("com.avito.android.libraries")
```
eugene-krivobokov, line 3: The grouping or maybe just naming is not clear for me.

- Directory: `kotlin-convention` - expect to see what Kotlin needs
- Plugin id: `.libraries` - kinda "all libraries"?
- Dependencies: kotlin, AGP?, buildlogic

How about this?

```kotlin
id("<>.kotlin")

dependencies {
    implementation("com.avito.android.buildlogic:libraries")
    implementation(platform("com.avito.android.infra:platform"))
    implementation(libs.kotlinPlugin)
    // without AGP
}
```

build-logic/kotlin-convention/build.gradle.kts
```diff
@@ -0,0 +1,23 @@
...
+}
+
+group = "com.avito.android.buildlogic"
+
+dependencies {
+    implementation("com.avito.android.buildlogic:libraries")
+    implementation(platform("com.avito.android.infra:platform"))
+    implementation(libs.kotlinPlugin)
+    implementation(libs.androidGradlePlugin)
```
eugene-krivobokov, line 12: This `kotlin-convention` directory seems more `android-convention`

settings.gradle.kts
```diff
@@ -1,5 +1,14 @@
...
 
+// Convenience to provide constants for library GA coordinates
+includeBuild("libraries")
+
+// Platform for dependency versions shared by 'main build' and 'build-logic'
+includeBuild("platform")
+
+// Not traditional 'buildSrc', but 'build-logic' as a normal included build
+includeBuild("build-logic")
+
```
eugene-krivobokov, line 11: How about grouping them into one folder?
As for me, these are implementation details and don't need at a root level.
E.g. `libraries` sounds like regular modules.

dsvoronin: you mean all inside `build-logic`?
i'd prefer to merge it like this, test if all working correctly and then move it as we feel right, because it's not in a public api

build-logic/kotlin-convention/build.gradle.kts
dsvoronin: more like `conventions` )

subprojects/gradle/upload-to-googleplay/build.gradle.kts
dsvoronin: yes, i like this idea; but let me rename it afterwards

eugene-krivobokov: ok

settings.gradle.kts
eugene-krivobokov: Ok

build.gradle.kts
```diff
@@ -1,382 +1,40 @@
...
+    id("com.jfrog.bintray") apply false
 }
 
 buildscript {
     configurations.classpath {
         resolutionStrategy {
             // com.autonomousapps.dependency-analysis depends on older version of okio, and it's resolved for
             // our instrumentation-tests plugin in subprojects in runtime
-            force(Dependencies.okio)
+            force("com.squareup.okio:okio:2.7.0")
```
sergioboi, line 25: mmm
`Gradle` : "we make plugins for reusing versions"

subprojects/android-lib/proxy-toast/build.gradle.kts
```diff
@@ -1,10 +1,9 @@
...
-    `maven-publish`
-    id("com.jfrog.bintray")
+    id("com.avito.android.kotlin-android-library")
+    id("com.avito.android.libraries")
+    id("com.avito.android.publish")
 }
 
 dependencies {
-    implementation(Dependencies.androidAnnotations)
+    implementation(libs.androidAnnotations)
```
sergioboi, line 8: What do you not like about using `Dependencies`

gradle.properties
```diff
@@ -2,6 +2,8 @@ kotlin.code.style=official
 org.gradle.jvmargs=-Xmx4g -XX:+UseGCOverheadLimit -XX:GCTimeLimit=10 -Dfile.encoding=UTF-8
 org.gradle.parallel=true
 org.gradle.caching=true
+# TODO https://github.com/gradle/gradle/issues/4823
+org.gradle.configureondemand=false
```
sergioboi, line 6: sadfrod

libraries/src/main/kotlin/com/avito/android/ExternalLibrariesExtension.kt
```diff
@@ -0,0 +1,131 @@
+package com.avito.android
+
+import org.gradle.api.JavaVersion
+import org.gradle.api.provider.Provider
+import org.gradle.api.provider.ProviderFactory
+import javax.inject.Inject
+
+abstract class ExternalLibrariesExtension @Inject constructor(private val providers: ProviderFactory) {
```
sergioboi, line 8: Prefer to see a structure that was before

build.gradle.kts
dsvoronin: i'll fix it later

eugene-krivobokov (approved):

subprojects/android-lib/proxy-toast/build.gradle.kts
dsvoronin: please take a look on MBS-10605 task description

PR #782, opened 2021-02-14
add GRADLE_USER_HOME env, fixes cache and wrapper volumes mounts

### LINKED ISSUE #778: Downloading gradlew each docker run of android-builder
Wrote a similar Dockerfile and now I see downloading of gradlew despite I have volumes for gradle and gradlew caches
image

Glad if you can check if this issue is reproducible in your infrastructure
image

dsvoronin:
Fixed #778


ci/_main.sh
```diff
@@ -105,6 +105,7 @@ function runInBuilder() {
         --env TZ="Europe/Moscow" \
         --env LOCAL_USER_ID="$USER_ID" \
         --env BINTRAY_GPG_PASSPHRASE="$BINTRAY_GPG_PASSPHRASE" \
+        --env GRADLE_USER_HOME=/gradle \
```
eugene-krivobokov, line 108: This env is optional. Should we consider it?

dsvoronin: You mean we should fail on it's abscence in entrypoint?

eugene-krivobokov: This is an option. 
The alternative is to repeat default behavior: use `$USER_HOME/.gradle` if not set.

dsvoronin: There is a problem with mounting into folder of not yet created user; 
i didn't go deeper, please check `entrypoint.sh` changes

eugene-krivobokov (approved):

PR #784, opened 2021-02-15
fix buildchecks publishing

dsvoronin:


eugene-krivobokov (approved):

subprojects/gradle/build-checks/build.gradle.kts
```diff
@@ -28,7 +28,7 @@ dependencies {
 gradlePlugin {
     plugins {
         create("buildChecks") {
-            id = "com.avito.android.buildchecks"
+            id = "com.avito.android.build-checks"
```
eugene-krivobokov, line 31: It will break backward compatibility for clients.

Also, need to change it in docs and tests

dsvoronin: changed

PR #787, opened 2021-02-17
Attach threads dump to RV on Incedent errors

sergioboi:


subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/runner/InHouseInstrumentationTestRunner.kt
```diff
@@ -283,6 +284,7 @@ abstract class InHouseInstrumentationTestRunner :
     fun tryToReportUnexpectedIncident(incident: Throwable) {
         try {
             if (!report.isWritten) {
+                report.addHtml("Threads dump", ThreadDumper.getThreadDump(), false)
```
sergioboi, line 287: @eugene-krivobokov @dsvoronin I need your opinion,
Should we attach threads dump on each failed test or on incidents would be enough?

dsvoronin: I think it's enough even for `could not launch intent` case for now

dsvoronin (approved):

PR #790, opened 2021-02-18
MBS-10559 add missing library variant configs

dsvoronin:
add android.buildFeatures config
add compileAll task

### DIFF
--- a/build-logic/android-convention/src/main/kotlin/convention.android-base.gradle.kts
+++ b/build-logic/android-convention/src/main/kotlin/convention.android-base.gradle.kts
@@ -1,5 +1,4 @@
 import com.android.build.gradle.BaseExtension
-import org.gradle.kotlin.dsl.configure
 
 plugins {
     id("convention.libraries")
@@ -30,5 +29,17 @@ configure<BaseExtension> {
         isWarningsAsErrors = true
         textReport = true
         isQuiet = true
+    }
+
+    @Suppress("UnstableApiUsage")
+    with(buildFeatures) {
+        aidl = false
+        compose = false
+        buildConfig = false
+        prefab = false
+        renderScript = false
+        resValues = false
+        shaders = false
+        viewBinding = false
     }
 }
--- a/build-logic/android-convention/src/main/kotlin/convention.kotlin-android-app.gradle.kts
+++ b/build-logic/android-convention/src/main/kotlin/convention.kotlin-android-app.gradle.kts
@@ -4,3 +4,26 @@ plugins {
     id("convention.kotlin-base")
     id("convention.android-base")
 }
+
+android {
+
+    testBuildType = "debug"
+
+    /**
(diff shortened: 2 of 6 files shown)


build-logic/android-convention/src/main/kotlin/convention.android-base.gradle.kts
```diff
@@ -31,4 +30,16 @@ configure<BaseExtension> {
...
+    with(buildFeatures) {
+        aidl = false
+        compose = false
+        buildConfig = false
+        prefab = false
+        renderScript = false
+        resValues = false
+        shaders = false
+        viewBinding = false
+    }
```
eugene-krivobokov, line 44: Just thoughts out loud. 
We have defaults:

```properties
android.defaults.buildfeatures.aidl=false
android.defaults.buildfeatures.buildconfig=false
...
```
But it's easy to miss renamings. Configuration by code is more robust.

eugene-krivobokov (approved):

build-logic/android-convention/src/main/kotlin/convention.kotlin-android-app.gradle.kts
```diff
@@ -4,3 +4,26 @@ plugins {
...
+    /**
+     * Disable all buildTypes except testing
+     * to avoid confusing errors in IDE if wrong build variant is selected
+     */
+    variantFilter {
+        if (name != testBuildType) {
+            ignore = true
+            logger.debug("Build variant $name is omitted for module: $path")
+        }
+    }
```
orcchg, line 21: Could you please explain the reason for that? Why do you only allow "debug" build types for app-modules? Do you actually use dynamic-feature-modules? If you actually do, then my question is: consider if you have :my-dynamic-feature and its base module :app, which is 'com.android.application'. Given only a "debug" build type for :app, we can not build "release" application using :my-dynamic-feature unless we specify matching fallbacks for the latter. But it seems to be weird, that "release" :my-dynamic-feature falls back to use "debug" version of :app.
Does it make sense? If we do use dynamic-feature modules, should we permit "release" build types for app-modules? Or you'd suggest enabling that build type only for "base" (for dynaminc-feature) module (which is normally :app module) directly in its build.gradle file? Thank you!

dsvoronin: We don't use dynamic-feature-modules in any of our projects, and so has no real world experience on how to deal with it.

This config only applies to project you can see, so it's not the same way production avito app works
We simple don't need anything except debug for ui testing of app modules belongs to this repository;

i can add clarification to this code comment if you think it could be helpful.

orcchg: @dsvoronin Thank you for explanation, this is quite clear for me now. A clarifying comment would be great in any way :D

PR #791, opened 2021-02-18
MBS-10666 new test runner gradle plugin; gradle plugin publication poc

dsvoronin:
Plan is to apply one or another plugin based on feature flag
Both plugin will reuse same logic extracted to `runner-client` module


subprojects/gradle/cd/build.gradle.kts
```diff
@@ -1,6 +1,6 @@
 plugins {
     id("convention.kotlin-jvm")
-    id("convention.publish-gradle-plugin")
+    id("convention.publish-gradle-plugin-legacy")
```
sergioboi, line 3: I don't get what is the purpose of that plugin?

dsvoronin: convention for publication of gradle plugins;

problem is here: https://github.com/avito-tech/avito-android/pull/791/files#diff-e7b95c7356270a4bf48fea167dfbb47a8fa67e7a8ab35930fce58a7e9c047f2cR12

trying to publish new plugin correctly with new convention along with legacy one

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/android/testrunner/service/TestService.kt
```diff
@@ -0,0 +1,36 @@
+package com.avito.android.testrunner.service
+
+import com.avito.android.testrunner.service.TestService.Params
+import com.avito.utils.gradle.KubernetesCredentials
+import org.gradle.api.provider.Property
+import org.gradle.api.services.BuildService
+import org.gradle.api.services.BuildServiceParameters
+
+@Suppress("UnstableApiUsage")
+public abstract class TestService : BuildService<Params>, AutoCloseable {
```
eugene-krivobokov, line 10: It'd easier to see this change in another PR. It doesn't related to plugins publication

eugene-krivobokov (approved):

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/android/testrunner/service/TestService.kt
dsvoronin: ok, i'll extract publication logic

dsvoronin:
will do another one

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/android/testrunner/service/TestService.kt
eugene-krivobokov: Thanks

PR #794, opened 2021-02-20
MBS-10675 skip lambda toString reflection

dsvoronin:
sometimes leads to KotlinReflectionInternalError; see [KT-40666](https://youtrack.jetbrains.com/issue/KT-40666)

promote LOST causes to Critical
catch Throwables not Exceptions

### DIFF
--- a/subprojects/android-test/test-report/src/main/kotlin/com/avito/android/test/report/transport/PreTransportMappers.kt
+++ b/subprojects/android-test/test-report/src/main/kotlin/com/avito/android/test/report/transport/PreTransportMappers.kt
@@ -2,7 +2,9 @@ package com.avito.android.test.report.transport
 
 import com.avito.android.test.report.model.DataSet
 import com.avito.android.test.report.model.StepResult
+import com.avito.android.util.isLambda
 import com.avito.report.model.Step
+import java.lang.reflect.Field
 
 internal interface PreTransportMappers {
 
@@ -20,12 +22,15 @@ internal interface PreTransportMappers {
     fun DataSet.serialize(): Map<String, String> {
         return javaClass.declaredFields
             .filter { it.name != "number" }
-            .map { field ->
-                field.name to javaClass.methods
-                    .find { it.name == "get${field.name.capitalize()}" }
-                    ?.invoke(this)
-                    .toString()
-            }
+            .filter { !it.type.isLambda() }
+            .map { field -> field.name to getFieldStringValue(field) }
             .toMap()
+    }
+
+    private fun Any.getFieldStringValue(field: Field): String {
+        return javaClass.methods
+            .find { it.name == "get${field.name.capitalize()}" }
+            ?.invoke(this)
+            .toString()
     }
 }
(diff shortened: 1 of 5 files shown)


subprojects/android-test/test-report/src/main/kotlin/com/avito/android/test/report/transport/TestRuntimeDataBuilder.kt
```diff
@@ -19,7 +19,7 @@ class TestRuntimeDataBuilder(private val timeProvider: TimeProvider) : PreTransp
                 startTime = state.startTime,
                 endTime = state.endTime
             )
-        } catch (e: Exception) {
+        } catch (e: Throwable) {
```
dsvoronin, line 22: This catch didn't help because `KotlinReflectionInternalError` was thrown

subprojects/android-test/test-report/src/main/kotlin/com/avito/android/test/report/transport/PreTransportMappers.kt
```diff
@@ -20,12 +22,15 @@ internal interface PreTransportMappers {
...
         return javaClass.declaredFields
             .filter { it.name != "number" }
-            .map { field ->
-                field.name to javaClass.methods
-                    .find { it.name == "get${field.name.capitalize()}" }
-                    ?.invoke(this)
-                    .toString()
-            }
+            .filter { !it.type.isLambda() }
+            .map { field -> field.name to getFieldStringValue(field) }
```
Kutashov, line 26: Not sure, but can't we use kotlin property https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.reflect.jvm/java.lang.reflect.-field/ instead, can we?

Kutashov (approved):

subprojects/android-test/test-report/src/main/kotlin/com/avito/android/test/report/transport/PreTransportMappers.kt
dsvoronin: I think we could, not sure if java supported at all now

subprojects/android-test/test-report/src/test/kotlin/com/avito/android/test/report/transport/DataSetSerializerKtTest.kt
```diff
@@ -21,4 +19,11 @@ class DataSetSerializerKtTest : PreTransportMappers {
         assertThat(SinglePropertyDataSet("myValue").serialize())
             .isEqualTo(mapOf("myProperty" to "myValue"))
     }
+
+    @Test
+    fun `serialize() - not serializing lambda`() {
+        class LambdaPropertyDataSet(@Suppress("unused") val myLambda: () -> String) : DataSet
+
+        assertThat(LambdaPropertyDataSet { "Hello" }.serialize()).isEqualTo(emptyMap<String, Any>())
```
eugene-krivobokov, line 27: It'd be helpful to see in a report a self-described value, e.g. `Unsupported type: lambda value <issue>`

PR #795, opened 2021-02-20
Fix KubernetesReservationClient.kt channel races

sergioboi:
The most nicely that tests really find bugs

### DIFF
--- a/subprojects/gradle/runner/device-provider/src/main/kotlin/com/avito/android/runner/devices/internal/kubernetes/KubernetesReservationClient.kt
+++ b/subprojects/gradle/runner/device-provider/src/main/kotlin/com/avito/android/runner/devices/internal/kubernetes/KubernetesReservationClient.kt
@@ -21,9 +21,12 @@ import kotlinx.coroutines.channels.ReceiveChannel
 import kotlinx.coroutines.channels.consumeEach
 import kotlinx.coroutines.channels.distinctBy
 import kotlinx.coroutines.channels.filter
+import kotlinx.coroutines.channels.map
 import kotlinx.coroutines.channels.toList
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
 import kotlinx.coroutines.withContext
 
 @OptIn(ExperimentalCoroutinesApi::class)
@@ -41,6 +44,8 @@ internal class KubernetesReservationClient(
 
     private var state: State = State.Idling
 
+    private val lock = Mutex()
+
     override suspend fun claim(
         reservations: Collection<ReservationData>,
         scope: CoroutineScope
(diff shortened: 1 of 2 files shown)


subprojects/gradle/runner/device-provider/src/main/kotlin/com/avito/android/runner/devices/internal/kubernetes/KubernetesReservationClient.kt
```diff
@@ -41,38 +44,42 @@ internal class KubernetesReservationClient(
 
     private var state: State = State.Idling
 
+    private val lock = Mutex()
```
sergioboi, line 47: Hello, `coroutines` without races. LOL
So we concurrently close channels when other coroutines try to read and write with them
And there is no way except mutex to make this safer

subprojects/gradle/runner/device-provider/src/main/kotlin/com/avito/android/runner/devices/internal/kubernetes/KubernetesReservationClient.kt
```diff
@@ -103,16 +113,18 @@ internal class KubernetesReservationClient(
         launch(CoroutineName("waiting-pods")) {
             @Suppress("DEPRECATION")
             podsChannel
+                .map {
+                    if (serialsChannel.isClosedForSend) {
```
sergioboi, line 117: because of `distinct` operation on channels, we should check serialsChannel on the upper level

subprojects/gradle/runner/device-provider/src/main/kotlin/com/avito/android/runner/devices/internal/kubernetes/KubernetesReservationClient.kt
```diff
@@ -231,14 +245,20 @@ internal class KubernetesReservationClient(
         var pods = kubernetesApi.getPods(deploymentName)
 
         @Suppress("EXPERIMENTAL_API_USAGE")
-        while (!podsChannel.isClosedForSend && pods.isNotEmpty()) {
-            pods.forEach { pod ->
-                podsChannel.send(pod)
-            }
+        while (pods.isNotEmpty()) {
```
sergioboi, line 248: For example, we check from coroutine that `podsChannel.isClosed` and then send but another coroutine could concurrently close the channel
So we need to get a mutex and make those bunch of operations atomicaly

subprojects/gradle/runner/device-provider/src/test/kotlin/com/avito/android/runner/devices/internal/kubernetes/KubernetesReservationClientTest.kt
```diff
@@ -100,6 +102,7 @@ internal class KubernetesReservationClientTest {
     }
 
     @Test
+    @Timeout(1, unit = TimeUnit.SECONDS)
```
sergioboi, line 105: I don't want my tests to lead to default 600s timeout

subprojects/gradle/runner/device-provider/src/main/kotlin/com/avito/android/runner/devices/internal/kubernetes/KubernetesReservationClient.kt
```diff
@@ -85,9 +92,12 @@ internal class KubernetesReservationClient(
             namespace = kubernetesApi.namespace,
             reservation = reservation
         )
-        val deploymentName = deployment.metadata.name
-        deploymentsChannel.send(deploymentName)
-        kubernetesApi.createDeployment(deployment)
+        val deploymentName = lock.withLock {
+            val deploymentName = deployment.metadata.name
+            deploymentsChannel.send(deploymentName)
```
eugene-krivobokov, line 97: Is it ok that we send a deployment before creation?

eugene-krivobokov (approved):

subprojects/gradle/runner/device-provider/src/main/kotlin/com/avito/android/runner/devices/internal/kubernetes/KubernetesReservationClient.kt
sergioboi: I don't know I'm scared to change that)))

PR #798, opened 2021-02-20
MBS-10652 Add mkdocs sctrict check to PR

eugene-krivobokov:


ci/build.sh
```diff
@@ -7,3 +7,5 @@ set -e
 source $(dirname $0)/_main.sh
 
 runInBuilder "set -e; ./gradlew build ${GRADLE_ARGS}"
+
+bash $(dirname $0)/documentation/lint.sh
```
eugene-krivobokov, line 11: I've decided to check documentation after the build.
In case of error, it can be check locally and fixed fast. We don't need fast feedback loop for such rare cases.

ci/documentation/preview.sh
```diff
@@ -22,6 +22,7 @@ function openBrowser() {
 
 openBrowser
 
+# NB: Can't use strict mode with serving. Use lint.sh for checking.
```
eugene-krivobokov, line 25: Strict mode is used for publishing to production. It detects broken links, menus, and so on.

Don't know a way to enable it in serving. 
`dev-addr` usage is a warning, and strict mode forbids all warnings.

dsvoronin (approved):

PR #801, opened 2021-02-22
Mbs 10577 Extract Android application specific build checks

eugene-krivobokov:
Extract Android application specific build checks.
They are scoped to an application in all senses and should be configured separately.

Before:

```kts
root:

plugins {
    id("com.avito.android.build-checks")
}
buildChecks {
    uniqueRClasses {
        allowedNonUniquePackageNames.add("lib.package") <--- global config for all apps
    }
}
```

After:

```kts
application:

plugins {
    id("com.avito.android.build-checks")
}
buildChecks {
    uniqueRClasses {
        allowedNonUniquePackageNames.add("lib.package") <--- app specific config
    }
}
```


subprojects/gradle/build-checks/src/main/kotlin/com/avito/android/build_checks/BuildParamCheckPlugin.kt
```diff
@@ -162,129 +182,7 @@ public open class BuildParamCheckPlugin : Plugin<Project> {
             }
         }
     }
-
```
eugene-krivobokov, line 165: Extracted logic to "checkers". For now, with no changes in logic.

dsvoronin (approved):

dsvoronin:
And it will be backward compatible, right?

eugene-krivobokov:
> And it will be backward compatible, right?

Only binary compatibility but behavior is changed.
In this release, it won't work in runtime but warn about required migration.

sergioboi:
But what about copy-paste when I want to configure each app identically

eugene-krivobokov:
> But what about copy-paste when I want to configure each app identically

In this case, you can apply this logic by code in subprojects block or by convention plugins.
It's a matter of reusing common settings. E.g. like for AGP plugins.

PR #803, opened 2021-02-22
fix sync without optional parameters

dsvoronin:


subprojects/gradle/slack/build.gradle.kts
```diff
@@ -30,5 +30,7 @@ tasks.named<Test>("integrationTest").configure {
 }
 
 fun Test.applyOptionalSystemProperty(name: String) {
-    project.property(name)?.toString()?.let { value -> systemProperty(name, value) }
+    if(project.hasProperty(name)) {
+        project.property(name)?.toString()?.let { value -> systemProperty(name, value) }
+    }
 }
```
eugene-krivobokov: Can we reuse the function between modules?

eugene-krivobokov (approved):

subprojects/gradle/slack/build.gradle.kts
dsvoronin: I don't know how yet

dsvoronin: just extracted this function to convention plugin directory and it seems working

PR #805, opened 2021-02-24
Fix UnitTestCheck.kt

sergioboi:
We don't need to execute unit-test when android-test configuration change

### DIFF
--- a/subprojects/gradle/cd/src/main/kotlin/com/avito/ci/steps/UnitTestCheck.kt
+++ b/subprojects/gradle/cd/src/main/kotlin/com/avito/ci/steps/UnitTestCheck.kt
@@ -17,7 +17,6 @@ open class UnitTestCheck(context: String, name: String) : BuildStep(context, nam
 
             val configurations = project.internalModule.let {
                 listOf(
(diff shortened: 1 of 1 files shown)


subprojects/gradle/cd/src/main/kotlin/com/avito/ci/steps/UnitTestCheck.kt
```diff
@@ -17,7 +17,6 @@ open class UnitTestCheck(context: String, name: String) : BuildStep(context, nam
 
             val configurations = project.internalModule.let {
                 listOf(
-                    it.androidTestConfiguration,
                     it.testConfiguration,
                     it.implementationConfiguration
                 )
```
eugene-krivobokov, line 22: Will you add integration tests?

eugene-krivobokov (approved):

subprojects/gradle/cd/src/main/kotlin/com/avito/ci/steps/UnitTestCheck.kt
sergioboi: I will try

PR #809, opened 2021-02-24
MBS-10559 working maven central publication

dsvoronin:
## workaround for implicit staging repository creation

**problem**: parallel publication from multiple modules creates multiple implicit staging repositories on nexus; its hard to understand how to push it to central, which one is needed which is not
**solution**: create staging repo via api in separate task, publish all to this repo

## validation for central pom fixes

Maven central requirements: https://central.sonatype.org/pages/requirements.html

- javadocs was missing
- constraints publishing not working as is with pom validation error
 
## add version and teamcity build id to staging repo

to distinguish staging repositories


eugene-krivobokov (approved):

build-logic/publication/src/main/kotlin/convention.sonatype.gradle.kts
```diff
@@ -1,32 +1,74 @@
...
+}
+
+tasks.withType<PublishToMavenRepository>().configureEach {
+
+    // https://docs.gradle.org/current/userguide/publishing_customization.html#sec:configuring_publishing_tasks
+    if (name.contains(sonatypeRepoName)) {
+        doFirst {
+
+            // no direct task access, because "cannot be cast to class CreateStagingRepositoryTask" for some reason
+            val repositoryUrl = repositoryUrlOutputFilePath.get().asFile.readText()
```
eugene-krivobokov, line 51: It may be better to do IO in a dedicated task or create a provider.
I don't know a way to implement in `doFirst` / `doLast` a  full contract for a task: caching (inputs/outputs) and so on. 
See `TaskActionWrapper` inside `doFirst` implementation.

build-logic/publication/src/main/kotlin/com/avito/android/publish/CreateStagingRepositoryTask.kt
```diff
@@ -0,0 +1,105 @@
...
+import org.gradle.api.DefaultTask
+import org.gradle.api.file.RegularFile
+import org.gradle.api.provider.Property
+import org.gradle.api.tasks.Input
+import org.gradle.api.tasks.OutputFile
+import org.gradle.api.tasks.TaskAction
+import java.io.IOException
+import java.util.concurrent.TimeUnit
+
+abstract class CreateStagingRepositoryTask : DefaultTask() {
```
eugene-krivobokov, line 21: How can we test and debug this logic?

dsvoronin: it can be tested right in `publication/src/main/test`

PR #813, opened 2021-02-27
Include subprojects build; refactor Makefile

dsvoronin:
Maybe it's better to divide this proposals to separate PR's, if its uncomfortable to review

Proposals:

## Drop separate bash scripts in favor of makefile for both local and ci entrypoints

There is some duplication between bash and makefile
And Makefile rule can be run with `docker=true` now

## Move subprojects build inside its directory

It's unclear where root `build.gradle.kts` and `setting.gradle.kts` belongs.
Moved to `subprojects/` for clarity and future re-introduction of `samples/`


ci/docker/android-builder/entrypoint.sh
```diff
@@ -23,6 +23,10 @@ else
     find "${GRADLE_USER_HOME}" -maxdepth 3 -type d -not -user ${BUILD_USER} -print -execdir chown ${BUILD_USER} {} \+
 fi
 
+sudo --set-home --preserve-env "PATH=$PATH" -u ${BUILD_USER} bash -c "git config --global core.sshCommand 'ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no';
+git config --global user.name 'builder';
+git config --global user.email 'builder@avito.ru';"
```
eugene-krivobokov, line 28: What do we do as builder user in git?

Makefile
```diff
@@ -64,93 +93,113 @@ __check_defined = \
...
+# По-разным причинам работа контейнера при прошлой сборке может не завершиться
+# Здесь мы перестраховываемся и останавливаем все работающие контейнеры
+# Перед сборкой не должно быть других контейнеров в любом случае
+clear_docker_containers:
+	containers=$(docker container ls -aq)
+	if [[ ! -z "$(containers)" ]]; then \
+	docker container rm --force $(containers); \
+	fi
+
+init_git:
```
eugene-krivobokov: I'd like to hide such functions from autocompletion.
As for me, they are internal.

At least, we can add prefix to show this.

Makefile
```diff
@@ -12,6 +21,25 @@ useCompositeBuild=true
...
+	--volume "$(shell pwd)":/app \
+	--volume "$(GRADLE_CACHE_DIR)":/gradle/caches \
+	--volume "$(GRADLE_WRAPPER_DIR)":/gradle/wrapper \
+	--workdir /app \
+	--env TZ="Europe/Moscow" \
+	--env LOCAL_USER_ID="$(USER_ID)" \
+	--env GRADLE_USER_HOME=/gradle \
+	android-builder
+endef
+endif
```
eugene-krivobokov, line 46: About full migration to makefile:

I'd like to discuss such decisions in a more formal way. I see a change, but don't understand the whole picture.
I'm trying to figure out PRO/CON and compare (example: https://medium.com/redbubble/rewriting-our-deploy-tooling-from-makefile-to-bash-and-back-again-4edeab5b0afa). 
Will we do this in an internal repository?

settings.gradle.kts
```diff
@@ -1,330 +1,5 @@
...
-}
-
-fun booleanProperty(name: String, defaultValue: Boolean): Boolean {
-    return if (settings.extra.has(name)) {
-        settings.extra[name]?.toString()?.toBoolean() ?: defaultValue
-    } else {
-        defaultValue
-    }
-}
+includeBuild("subprojects")
```
eugene-krivobokov, line 5: What negative sides do you see?

For instance:
- Will any task rebuild all subprojects?

ci/docker/android-builder/entrypoint.sh
dsvoronin: It is required for gradle integration tests using git commands

Makefile
dsvoronin: agree

Makefile
dsvoronin: I like to see this repo as a platform for more experiments, successful ones could be used in internal, this one could too.
Do you like to see a discussion about makefile first? With more bold points?

settings.gradle.kts
dsvoronin: nope, it won't 
`Task 'assemble' not found in root project 'avito-android-infra'.`

Makefile
eugene-krivobokov: > Do you like to see a discussion about makefile first? With more bold points?

Yes, it would be helpful to write and share knowledge about such experiments. 
Ideally, I want to manage them as some artifacts besides the code. Maybe only in issues. For big changes - tech specs, ... 

When something is only in code, it's harder to see common points: 
- What experiments we have and why? 
- In what state?
- How can I help?

The change itself is good. I'm talking only about making the process more transparent. It's like the idea from Kanban - to visualize all unfinished work

sergioboi (approved):

ci/docker/android-builder/entrypoint.sh
eugene-krivobokov: Ok, just didn't find how it worked before

Makefile
dsvoronin: moved it to entrypoint

Makefile
dsvoronin: I will merge this one as it only adds value: local makefile commands in docker
And not changes bash entrypoints yet
Will start a discussion

PR #814, opened 2021-03-01
MBS-10556 add missing retries to ReportApi and FileStorage

dsvoronin:
All methods are POST, was retrying only GET's
discovered while browsing kibana logs

### DIFF
--- a/build-logic/testing-convention/src/main/kotlin/convention.unit-testing.gradle.kts
+++ b/build-logic/testing-convention/src/main/kotlin/convention.unit-testing.gradle.kts
@@ -19,6 +19,15 @@ tasks.withType<Test>().configureEach {
      * fix for multiple `WARNING: Illegal reflective access`
      */
     jvmArgs = listOf("--add-opens", "java.base/java.lang=ALL-UNNAMED")
+
+    /**
+     * IDEA adds an init script, using it to define if it is an IDE run
+     * used in `:test-project`
+     */
+    systemProperty(
+        "isInvokedFromIde",
+        gradle.startParameter.allInitScripts.find { it.name.contains("ijtestinit") } != null
+    )
 }
 
 plugins.withType<KotlinBasePluginWrapper>() {
--- a/subprojects/common/report-viewer/src/main/kotlin/com/avito/report/ReportsApi.kt
+++ b/subprojects/common/report-viewer/src/main/kotlin/com/avito/report/ReportsApi.kt
@@ -1,5 +1,6 @@
 package com.avito.report
 
+import com.avito.http.RetryInterceptor
 import com.avito.logger.LoggerFactory
 import com.avito.logger.create
 import com.avito.report.internal.JsonRpcRequestProvider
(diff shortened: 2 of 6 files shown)


subprojects/common/report-viewer/src/main/kotlin/com/avito/report/ReportsApi.kt
```diff
@@ -66,15 +67,21 @@ interface ReportsApi : ReportsAddApi, ReportsFetchApi {
...
                         verbose = verboseHttp,
-                        logger = loggerFactory.create<ReportsApi>(),
+                        logger = logger,
                         readTimeoutSec = readTimeout,
-                        writeTimeoutSec = writeTimeout
+                        writeTimeoutSec = writeTimeout,
+                        retryInterceptor = RetryInterceptor(
+                            logger = logger,
+                            allowedMethods = listOf("POST")
+                        )
```
eugene-krivobokov, line 84: What statuses did we receive from RV?
Do we need to add them to the interceptor?

eugene-krivobokov (approved):

subprojects/common/report-viewer/src/main/kotlin/com/avito/report/ReportsApi.kt
dsvoronin: https://github.com/avito-tech/avito-android/blob/0bada247948b9db5ff93a89b33251b730ab65bcb/subprojects/common/okhttp/src/main/kotlin/com/avito/http/RetryInterceptor.kt#L18-L22

These defaults are ok, it's 502 mainly

build-logic/testing-convention/src/main/kotlin/convention.unit-testing.gradle.kts
```diff
@@ -19,6 +19,15 @@ tasks.withType<Test>().configureEach {
...
     jvmArgs = listOf("--add-opens", "java.base/java.lang=ALL-UNNAMED")
+
+    /**
+     * IDEA adds an init script, using it to define if it is an IDE run
+     * used in `:test-project`
+     */
+    systemProperty(
+        "isInvokedFromIde",
+        gradle.startParameter.allInitScripts.find { it.name.contains("ijtestinit") } != null
+    )
```
eugene-krivobokov, line 30: What about Gradle tests?

subprojects/common/report-viewer/src/test/kotlin/com/avito/report/ReportsApiTest.kt
```diff
@@ -55,7 +62,9 @@ internal class ReportsApiTest {
 
     @Test
     fun `getReport - returns Error - when throws exception with no data`() {
-        mockWebServer.enqueue(MockResponse().setResponseCode(500))
+        repeat(retryCount) {
```
eugene-krivobokov, line 65: Is it an imitation of "unrecoverable" case?
In this case we duplicate implementation here. We can show the intention by Dispatcher and don't rely on count.

PR #815, opened 2021-03-01
MBS-10556 promote incubating report hosts configuration

dsvoronin:
Supported in Avito project

### DIFF
--- a/subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/configuration/InstrumentationTestReportExtension.kt
+++ b/subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/configuration/InstrumentationTestReportExtension.kt
@@ -1,5 +1,6 @@
 package com.avito.instrumentation.configuration
 
+import groovy.lang.Closure
 import org.gradle.api.Action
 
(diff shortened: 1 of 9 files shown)


subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/configuration/InstrumentationTestReportExtension.kt
```diff
@@ -24,4 +25,12 @@ public class InstrumentationTestReportExtension {
         reportViewer.validate()
         this.reportViewer = reportViewer
     }
+
+    public fun reportViewer(closure: Closure<ReportViewer>) {
```
sergioboi, line 29: Action must be enough for both groovy and kotlin

sergioboi (approved):

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/configuration/InstrumentationTestReportExtension.kt
dsvoronin: But it's not working in nested one's; probably we doing something wrong

PR #818, opened 2021-03-01
MBS-10666 New experimental TestRunService

dsvoronin:
This is just a delegate to service with no changes;
preparation for further work

### DIFF
--- a/subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/service/TestRunnerService.kt
+++ b/subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/service/TestRunnerService.kt
@@ -1,0 +1,20 @@
+package com.avito.instrumentation.service
+
+import com.avito.instrumentation.internal.InstrumentationTestsAction
+import org.gradle.api.services.BuildService
+import org.gradle.api.services.BuildServiceParameters
+
+@Suppress("UnstableApiUsage")
+public abstract class TestRunnerService : BuildService<TestRunnerService.Params>, AutoCloseable {
+
+    public interface Params : BuildServiceParameters {
+        // todo
+    }
+
+    internal fun runTests(params: InstrumentationTestsAction.Params) {
+        InstrumentationTestsAction(params).run()
+    }
+
+    override fun close() {
+    }
+}
--- a/subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/service/TestRunnerWorkAction.kt
+++ b/subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/service/TestRunnerWorkAction.kt
@@ -1,0 +1,21 @@
+package com.avito.instrumentation.service
+
+import com.avito.instrumentation.internal.InstrumentationTestsAction
+import org.gradle.api.provider.Property
+import org.gradle.workers.WorkAction
+import org.gradle.workers.WorkParameters
+
+@Suppress("UnstableApiUsage")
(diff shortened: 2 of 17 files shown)


eugene-krivobokov (approved):

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/service/TestRunnerService.kt
```diff
@@ -0,0 +1,20 @@
...
+
+@Suppress("UnstableApiUsage")
+public abstract class TestRunnerService : BuildService<TestRunnerService.Params>, AutoCloseable {
+
+    public interface Params : BuildServiceParameters {
+        // todo
+    }
+
+    internal fun runTests(params: InstrumentationTestsAction.Params) {
+        InstrumentationTestsAction(params).run()
```
sergioboi, line 15: As I understand it runs action blocking compare to `WorkerApi`

sergioboi: Should we wrap this to `Work`?

sergioboi: I see the next file, forgot it

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/service/TestRunnerService.kt
```diff
@@ -0,0 +1,20 @@
+package com.avito.instrumentation.service
+
+import com.avito.instrumentation.internal.InstrumentationTestsAction
+import org.gradle.api.services.BuildService
+import org.gradle.api.services.BuildServiceParameters
+
+@Suppress("UnstableApiUsage")
+public abstract class TestRunnerService : BuildService<TestRunnerService.Params>, AutoCloseable {
```
sergioboi, line 8: Must it be serializable?

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/service/TestRunnerWorkAction.kt
```diff
@@ -0,0 +1,21 @@
+package com.avito.instrumentation.service
+
+import com.avito.instrumentation.internal.InstrumentationTestsAction
+import org.gradle.api.provider.Property
+import org.gradle.workers.WorkAction
+import org.gradle.workers.WorkParameters
+
+@Suppress("UnstableApiUsage")
+public abstract class TestRunnerWorkAction : WorkAction<TestRunnerWorkAction.Params> {
```
sergioboi, line 9: We could avoid serialization with `in memory work`

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/service/TestRunnerService.kt
dsvoronin: nope

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/service/TestRunnerWorkAction.kt
dsvoronin: yes, but why we need it? It should be serializable by gradle contract and i see no problem with it.
I understand that in case of inprocess strategy it is not necessary

PR #820, opened 2021-03-02
Add ability to add text entries to test reports

sergioboi:


subprojects/android-test/test-report/src/main/kotlin/com/avito/android/test/report/ReportImplementation.kt
```diff
@@ -254,9 +254,8 @@ class ReportImplementation(
     override fun addHtml(label: String, content: String, wrapHtml: Boolean) {
```
eugene-krivobokov, line 254: Do we still need `addHtml` with `wrapHtml = false`?

sergioboi: I don't know It hard for me to decide in that PR
Let's look further and then remove this parameter

eugene-krivobokov (approved):

PR #830, opened 2021-03-05
make for android-builder image publishing

dsvoronin:
+ publish image from #813


Makefile
```diff
@@ -161,6 +162,22 @@ build_android_image:
 	cd ./ci/docker/android-builder && \
 	docker build -t android-builder .
 
+# new ANDROID_BUILDER_TAG will be printed after successful publishing
+publish_android_builder_internal:
```
eugene-krivobokov: Let's use a prefix for visibility:

`internal_publish_android_builder`

Otherwise, we will get it in autocompletion after `publish_android_builder...`

eugene-krivobokov (approved):

PR #832, opened 2021-03-09
Fix sonatype publication

sergioboi:


build-logic/publication/src/main/kotlin/convention.sonatype.gradle.kts
```diff
@@ -32,9 +32,9 @@ val createStagingRepositoryTask: TaskProvider<CreateStagingRepositoryTask> = wit
         register<CreateStagingRepositoryTask>(createStagingTaskName) {
             group = "publication"
 
-            stagingProfileId.set(ossrhStagingProfileId)
-            user.set(ossrhUsername)
-            password.set(ossrhPassword)
+            stagingProfileId.set(ossrhStagingProfileId.get())
```
sergioboi: Force to fail when read those pros because It was not clear why and when we loss them

ci/publish.sh
```diff
@@ -1,7 +1,12 @@
 #!/usr/bin/env bash
 
 # This is an entrypoint for CI build step, don't change it's relative path(name)
+# force Envs be set up
+set -u
```
sergioboi, line 5: Force fails when we read `unset` env

ci/publish.sh
```diff
@@ -1,7 +1,12 @@
...
 
 source $(dirname $0)/_main.sh
-
-runInBuilder "./gradlew -p subprojects publishRelease ${GRADLE_ARGS} --no-parallel --stacktrace"
+readonly STAGING_PROFILE_ID="-Pavito.ossrh.stagingProfileId=${$OSSRH_STAGING_ID}"
+readonly USER="-Pavito.ossrh.user=$OSSRH_USER"
+readonly PASSWORD="-Pavito.ossrh.password=$OSSRH_PASSWORD"
+readonly CREDENTIALS = "${STAGING_PROFILE_ID} ${USER} ${PASSWORD}"
```
sergioboi: Move from TeamCity configuration to code. Because TeamCity conf was broken after moving `gradle.properties` file

build-logic/publication/src/main/kotlin/convention.sonatype.gradle.kts
dsvoronin: what if i don't have this properties locally? did you check that project syncing in IDE without it?

ci/publish.sh
dsvoronin: you moved only 3 out of 6 properties from teamcity; why so?

build-logic/publication/src/main/kotlin/convention.sonatype.gradle.kts
sergioboi: I get, turn this file back
I see that only one option to force validation and be safe it's an additional flag. I don't want to create that flag now

ci/publish.sh
sergioboi: Because of the evening)))

eugene-krivobokov (approved):

PR #836, opened 2021-03-12
Mbs 10762 Don't track build agent name in build metrics from CI

eugene-krivobokov:


subprojects/gradle/build-environment/src/main/kotlin/com/avito/utils/gradle/EnvironmentInfo.kt
```diff
@@ -0,0 +1,84 @@
...
+import com.avito.logger.handler.DefaultLoggingHandler
+import com.avito.utils.gradle.Environment
+import com.avito.utils.gradle.internal.EnvironmentInfoImpl
+import org.gradle.api.Project
+import org.gradle.api.provider.Provider
+
+/**
+ * Use [Project.environmentInfo] to gain instance
+ */
+interface EnvironmentInfo { // TODO: merge with BuildEnvironment and EnvArgs
```
eugene-krivobokov, line 19: Just moved temporarily this class from `sentry-config` module.
It is used mostly in metrics.

I want to make a better abstraction later. Maybe merge some attributes to similar abstractions: BuildEnvironment and EnvArgs

subprojects/gradle/build-environment/src/main/kotlin/com/avito/utils/gradle/internal/EnvironmentInfoImpl.kt
```diff
@@ -61,6 +63,4 @@ internal class EnvironmentInfoImpl(
     override fun teamcityBuildId(): String? {
         return project.rootProject.getOptionalStringProperty("teamcityBuildId")
     }
-
-    override fun isInvokedFromIde() = project.hasProperty("android.injected.invoked.from.ide")
```
eugene-krivobokov, line 65: Duplicates `BuildEnvironment`

subprojects/gradle/git/build.gradle.kts
```diff
@@ -5,9 +5,6 @@ plugins {
 }
 
 dependencies {
-    api(project(":gradle:build-environment")) {
-        because("project.buildEnvironment only")
-    }
```
eugene-krivobokov, line 10: It was not clear to consume build environment through git.

subprojects/gradle/build-metrics/src/main/kotlin/com/avito/android/plugin/build_metrics/BuildMetricTracker.kt
```diff
@@ -12,7 +13,11 @@ public class BuildMetricTracker(
 ) {
 
     private val node by lazy {
-        env.get().node?.take(32) ?: "unknown"
+        val env = env.get()
+        when (env.environment) {
+            is Environment.CI -> "_" // Don't need details from CI. Empty value for backward compatibility in series name
+            else -> env.node ?: "unknown"
+        }
```
eugene-krivobokov, line 21: The key change in behavior

dsvoronin (approved):

subprojects/gradle/build-metrics/src/main/kotlin/com/avito/android/plugin/build_metrics/BuildMetricTracker.kt
sergioboi: Explain please what scenarios are affected
And why we need this change

subprojects/gradle/build-environment/src/main/kotlin/com/avito/utils/gradle/EnvironmentInfo.kt
```diff
@@ -0,0 +1,84 @@
...
+    val node: String?
```
sergioboi, line 20: What is the node property?

eugene-krivobokov: It's used to describe a current host or machine in CI.

```kotlin
when (environment) {
            is Environment.Local -> gitUserEmail ?: userName()
            is Environment.Mirakle -> userName()
            is Environment.CI -> teamcityAgentName()
            is Environment.Unknown -> "unknown"
        }
```

It's a workaround for metric because we don't have a unique build id like in build scan.

subprojects/gradle/build-metrics/src/main/kotlin/com/avito/android/plugin/build_metrics/BuildMetricTracker.kt
eugene-krivobokov: This is only a performance issue - excessive metrics.

We add info to a metric for debug purposes:

`...<environment>.<node>...`

Environment: `ci` | `local` | `mirakle` | `_` (unknown)
Node: username for local builds, agent name for CI builds

In CI builds any additional information about a build agent is useless. It only produces unneeded unique metrics.

PR #837, opened 2021-03-14
Fix sonatype publication

sergioboi:


ci/publish.sh
```diff
@@ -1,16 +1,29 @@
...
-local readonly password="-Pavito.ossrh.password=${OSSRH_PASSWORD}"
 
-local readonly CREDENTIALS = "${profileId} ${user} ${password}"
+# force Envs be set up
+set -uex
+function publish() {
+    # PGP_KEY is a large 4kb key with unescaped symbols
+    local readonly gradle_properties_path="subprojects/gradle.properties"
+    if [[ -f ${gradle_properties_path} ]]; then
+        echo "avito.pgp.key=$PGP_KEY" >> ${gradle_properties_path}
```
sergioboi, line 13: So, I couldn't use `docker env-list`

eugene-krivobokov: Why?

eugene-krivobokov (approved):

ci/publish.sh
sergioboi: It didn't work as expected, so I decided to solve a problem how I can

PR #841, opened 2021-03-15
MBS-10798 describe jsonrpc method in RetryInterceptor

dsvoronin:


subprojects/common/report-viewer/src/main/kotlin/com/avito/report/internal/JsonRpcRequestProvider.kt
```diff
@@ -20,21 +20,44 @@ internal class JsonRpcRequestProvider(
...
-    private inline fun <reified Response : Any> internalRequest(request: Any): Response {
-        val response = httpClient.newCall(
-            Request.Builder()
-                .url(host.removeSuffix("/"))
-                .header("Accept", jsonMime)
-                .header("Content-Type", jsonMime)
-                .post(gson.toJson(request).toRequestBody(jsonMime.toMediaType()))
-                .build()
-        ).execute()
+    private inline fun <reified Response : Any> internalRequest(jsonRpcRequest: Any): Response {
```
sergioboi, line 23: Propose to replace `Any` with the concrete type

sergioboi (approved):

subprojects/common/report-viewer/src/main/kotlin/com/avito/report/internal/JsonRpcRequestProvider.kt
dsvoronin: `Any` was chosen, because `Any` is a parameter of `gson.toJson()`, and it was only consumer

subprojects/common/report-viewer/src/main/kotlin/com/avito/report/internal/JsonRpcRequestProvider.kt
```diff
@@ -20,21 +20,44 @@ internal class JsonRpcRequestProvider(
...
+            }
+        } else {
+            throw Exception("Request failed: ${response.message} ${responseBody}\n$jsonRpcRequest")
+        }
+    }
+
+    private fun getMethod(request: Any): String? = if (request is RfcRpcRequest) {
+        request.method
+    } else if (request is List<*>) {
+        val firstRequest = request.firstOrNull()
```
eugene-krivobokov, line 54: Why do we pick only the first from a batch?
We want to use these metrics to estimate an error rate.
Picking only the first method can create a false picture in case of different requests in the batch.

My proposal is to track all unique request methods and\or add info about the batch.

subprojects/common/report-viewer/src/main/kotlin/com/avito/report/internal/JsonRpcRequestProvider.kt
sergioboi: I don't understand. Could you explain this decision wider?
What I see, we have a concrete type for jsonRpcRequest but decide to cast it to `Any` because of JSON
I don't understand what problems JSON has with concrete type

PR #842, opened 2021-03-15
Mbs 10798 artifactory proxy

dsvoronin:
add missing artifactory proxies for convention plugins and gradle test kit

### DIFF
--- a/build-logic/settings.gradle.kts
+++ b/build-logic/settings.gradle.kts
@@ -4,3 +4,54 @@ include("kotlin-convention")
 include("android-convention")
 include("testing-convention")
 include("publication")
+
+val artifactoryUrl: String? by settings
+
+fun MavenArtifactRepository.artifactoryUrl(repositoryName: String) {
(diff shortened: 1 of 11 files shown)


build-logic/settings.gradle.kts
```diff
@@ -4,3 +4,54 @@ include("kotlin-convention")
...
+    if (artifactoryUrl.isNullOrBlank()) {
+        name = repositoryName
+        setUrl(originalRepo)
+    } else {
+        name = "Proxy for $repositoryName: $originalRepo"
+        artifactoryUrl(repositoryName)
+    }
+}
+
+dependencyResolutionManagement {
```
eugene-krivobokov, line 25: Don't we need the same proxies in `pluginManagement.repositories`?

eugene-krivobokov (approved):

build-logic/settings.gradle.kts
dsvoronin: yes, you're right

PR #843, opened 2021-03-15
MBS-10798 RemoteStorage http client 2min timeout and no retries

dsvoronin:


subprojects/common/file-storage/src/main/kotlin/com/avito/filestorage/RemoteStorageFactory.kt
```diff
@@ -0,0 +1,55 @@
...
+        loggerFactory = loggerFactory,
+        timeProvider = timeProvider
+    )
+
+    private fun getHttpClient(
+        verbose: Boolean = false,
+        logger: Logger,
+        retryInterceptor: RetryInterceptor?
+    ): OkHttpClient {
+        return OkHttpClient.Builder()
```
sergioboi, line 39: We should reuse OkHttpBuilder/Client
Every new client will produce its own Thread Pool

subprojects/common/file-storage/src/main/kotlin/com/avito/filestorage/RemoteStorageFactory.kt
```diff
@@ -0,0 +1,55 @@
...
+import com.avito.logger.create
+import com.avito.time.TimeProvider
+import okhttp3.HttpUrl.Companion.toHttpUrlOrNull
+import okhttp3.OkHttpClient
+import okhttp3.logging.HttpLoggingInterceptor
+import java.util.concurrent.TimeUnit
+
+object RemoteStorageFactory {
+
+    private const val TIMEOUT_SEC = 120L
```
sergioboi, line 16: Make a task for monitoring the time of `file-storage` requests because 120s is a random constant

subprojects/common/report-viewer/src/main/kotlin/com/avito/report/ReportsApiFactory.kt
```diff
@@ -55,4 +58,29 @@ object ReportsApiFactory {
...
+
+    private fun getHttpClient(
+        verbose: Boolean = false,
+        logger: Logger,
+        readTimeoutSec: Long,
+        writeTimeoutSec: Long,
+        retryInterceptor: RetryInterceptor?
+    ): OkHttpClient {
+
+        return OkHttpClient.Builder()
```
sergioboi, line 60: The same about reusing OkHttp

sergioboi (approved):

subprojects/common/file-storage/src/main/kotlin/com/avito/filestorage/RemoteStorageFactory.kt
eugene-krivobokov: We can do it in MBS-10704

PR #850, opened 2021-03-17
Add main looper dump to the report

sergioboi:


eugene-krivobokov (approved):

subprojects/android-test/test-report/src/main/kotlin/com/avito/android/test/report/ReportImplementation.kt
```diff
@@ -113,39 +113,45 @@ class ReportImplementation(
...
+            ResourceIncidentPresenter(),
+            ResourceManagerIncidentPresenter(),
+            AppCrashIncidentPresenter()
+        ),
+        fallbackPresenter = FallbackIncidentPresenter()
+    )
+
+    private fun addTroubleshootingEntries() {
+        addText("Threads dump", ThreadDumper.getThreadDump())
+        addText("Main looper dump", MainLooperDumper.getDump())
```
eugene-krivobokov: Do you have an example of how it looks?

sergioboi: Now only, I can't test it fast now. I look sources and decide that is enough

eugene-krivobokov: Ok, I'm just curious.

PR #859, opened 2021-03-21
Remove ambiguous try-catch-throw

sergioboi:
we will see an error in release builds if occurs


eugene-krivobokov (approved):

subprojects/gradle/upload-to-googleplay/src/main/kotlin/com/avito/upload_to_googleplay/GooglePlayDeployer.kt
```diff
@@ -47,21 +47,16 @@ interface GooglePlayDeployer {
...
-                        val editId = insertEdit(deploy.applicationId)
-                        val versionCode = uploadBinary(deploy, editId)
-                        setTrack(deploy, editId, versionCode)
-                        validate(deploy, editId)
-                        editId to deploy
-                    }
-                commitAll(editIdWithDeploy)
-                logger.info("SUCCESS: Deploy $deploys")
-            } catch (e: Exception) {
-                logger.critical("FAILED: Deploy", e)
```
dsvoronin, line 62: We won't find it in kibana now, is it ok?

PR #861, opened 2021-03-22
Report "invalid state" message 

dsvoronin:
will now group correctly in sentry
detailed cause moved to cause

image

^^ will be 2 errors

### DIFF
--- a/subprojects/android-test/test-report/src/main/kotlin/com/avito/android/test/report/ReportImplementation.kt
+++ b/subprojects/android-test/test-report/src/main/kotlin/com/avito/android/test/report/ReportImplementation.kt
@@ -113,9 +113,7 @@ class ReportImplementation(
         exception: Throwable,
         screenshot: FutureValue<RemoteStorage.Result>?,
     ) = methodExecutionTracing("registerIncident") {
-        val currentState = getCastedState<ReportState.Initialized> {
-            IllegalStateException(it, exception)
-        }
+        val currentState = getCastedState<ReportState.Initialized>(cause = exception)
 
         if (currentState.incident == null) {
             if (screenshot != null) {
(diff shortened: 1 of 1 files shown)


eugene-krivobokov (approved):

subprojects/android-test/test-report/src/main/kotlin/com/avito/android/test/report/ReportImplementation.kt
```diff
@@ -113,9 +113,7 @@ class ReportImplementation(
         exception: Throwable,
         screenshot: FutureValue<RemoteStorage.Result>?,
     ) = methodExecutionTracing("registerIncident") {
-        val currentState = getCastedState<ReportState.Initialized> {
```
sergioboi, line 116: I have already fixed not enough reporting in these lines

subprojects/android-test/test-report/src/main/kotlin/com/avito/android/test/report/ReportImplementation.kt
```diff
@@ -335,11 +333,29 @@ class ReportImplementation(
...
+        val castedClass = getCastedStateOrNull<T>()
+
+        if (castedClass == null) {
+            val _state = state
+
+            val detailedCause = cause ?: when (_state) {
+                is ReportState.Initialized.NotStarted ->
+                    IllegalStateException("Test not started, incident = ${_state.incident}")
+
+                is ReportState.Initialized.Started -> null
```
sergioboi, line 346: I disagree that we need different reactions when the expected state different from the actual.
Could you explain your motivation?

dsvoronin: Motivation is in PR descriptions, we reported doubles, because error message contained too much: object hashes and full toString content

i changed it to exception with `class.simpleName`s and additional `cause` which could be available or not, depending of state

sergioboi: I understand the part about `state::class.java.simpleName`
But why do you use  `_state.incident` only with `ReportState.Initialized.NotStarted`?
All `ReportState.Initialized` may have `_state.incident`

dsvoronin: ah, that was my mistake, let me fix it real quick

PR #863, opened 2021-03-22
Change retry logging level to DEBUG

sergioboi:
Current warnings amount is 22k only for half of the day
Screenshot 2021-03-22 at 14 45 38


subprojects/common/okhttp/src/main/kotlin/com/avito/http/RetryInterceptor.kt
```diff
@@ -44,9 +44,7 @@ public class RetryInterceptor constructor(
             prepareForRetry(response)
 
             val request = if (tryCount > 1) {
-                val modifiedRequest = modifyRetryRequest(originalRequest)
-                logger.debug("Retrying request ($tryCount / $retries) to ${describeRequest(modifiedRequest)}")
```
sergioboi, line 48: This covers by failed requests

dsvoronin (approved):

PR #864, opened 2021-03-22
Add test when second deployment creation fails then exception will be thrown

sergioboi:


subprojects/gradle/runner/device-provider/src/main/kotlin/com/avito/android/runner/devices/internal/kubernetes/KubernetesReservationClient.kt
```diff
@@ -32,6 +32,12 @@ import kotlinx.coroutines.sync.withLock
 import kotlinx.coroutines.withContext
 import kotlin.coroutines.cancellation.CancellationException
 
+/**
```
sergioboi, line 35:
```kotlin
@Test
    fun t1() {
        runBlocking {
            val channel = Channel<Int>(Channel.UNLIMITED)
            channel.invokeOnClose {
                println("Closed $it")
            }
            launch {
                channel.consumeEach { }
            }
            launch {
                throw RuntimeException()
            }
        }
    }

    @Test
    fun t2() {
        val channel = Channel<Int>(Channel.UNLIMITED)
        channel.invokeOnClose { println("Closed") }
        runBlocking {
            GlobalScope.launch {
                launch {
                    channel.consumeEach { }
                }
            }.cancelAndJoin()
        }
    }
```

eugene-krivobokov: Write results, please

sergioboi: "Close" will be printed in both cases

eugene-krivobokov (approved):

PR #868, opened 2021-03-22
Fix log

sergioboi:


subprojects/gradle/runner/device-provider/src/main/kotlin/com/avito/android/runner/devices/internal/kubernetes/KubernetesReservationClient.kt
```diff
@@ -280,7 +280,7 @@ internal class KubernetesReservationClient(
                 }
             }
         }
-        logger.debug("listenPodsFromDeployment finished")
+        logger.debug("listenPodsFromDeployment finished, [deploymentName=$deploymentName]")
```
sergioboi, line 283: @eugene-krivobokov I left `listenPodsFromDeployment` part because its function name
I think it will help to understand logs without navigating to sources but maybe I'm wrong

dsvoronin (approved):

subprojects/gradle/runner/device-provider/src/main/kotlin/com/avito/android/runner/devices/internal/kubernetes/KubernetesReservationClient.kt
eugene-krivobokov: It looks that we have two cases.

1. The lifecycle of the test runner. It expresses stages and it could not repeat used classes and function as it is.
I thought about this case.

2. Low-level detailed information for troubleshooting.
This is also ok, but we must be consistent. See the first message in the 258 line.

eugene-krivobokov (approved):

PR #870, opened 2021-03-22
wrapper now also copied to subprojects to handle IDE run problem

dsvoronin:


build.gradle.kts
```diff
@@ -0,0 +1,15 @@
+/**
+ * Tests run from IDE in subprojects module can't recognize root wrapper
+ */
+val subprojectsWrapper by tasks.registering(Copy::class) {
+    from("${rootDir}/gradle/wrapper")
+    into("$rootDir/subprojects/gradle/wrapper")
```
eugene-krivobokov, line 6:
```suggestion
    into("${rootDir}/subprojects/gradle/wrapper")
```
?

eugene-krivobokov (approved):

orcchg:
hello! Could you please explain why this changes and what kind of IDE problems? I suspect similar on my side, so I'd like to clarify. Thank you!

sergioboi:
> hello! Could you please explain why this changes and what kind of IDE problems? I suspect similar on my side, so I'd like to clarify. Thank you!

@orcchg We faced that when you run tests in a project with that kind of project structure then IDE implicitly uses different Gradle. We don't how IDE makes a decision on what Gradle to use but it doesn't use Gradle we defined in the root project

dsvoronin:
> hello! Could you please explain why this changes and what kind of IDE problems? I suspect similar on my side, so I'd like to clarify. Thank you!

I thinks it goes that way:

Current project structure is:

- root project (has wrapper folder, includeBuild subprojects)
- - subrojects

While clicking on run test button, IDE creates run configuration with subprjects` as root project 
Can't find gradle wrapper and fallbacks to default system one
I think it comes prebuild with AS

PR #873, opened 2021-03-23
Mbs 10763 Fixes in dependencies graph builder

eugene-krivobokov:
- Add missing cases for pseudo-cycles in test fixtures 
- Replace excessive Gradle Test Kit cases with printing by unit tests
- Reduce complexity in finding all dependencies.

See changes by commits


subprojects/gradle/module-dependencies-graph/src/gradleTest/kotlin/com/avito/module/dependencies/AppsDependenciesGraphTest.kt
```diff
@@ -1,163 +0,0 @@
...
-
-import com.avito.test.gradle.gradlew
-import org.junit.jupiter.api.BeforeEach
-import org.junit.jupiter.api.Test
-import org.junit.jupiter.api.io.TempDir
-import java.io.File
-
-// TODO: try to get rid of the printing task. It's used for testing only and fragile to changed output
-//  We can cover all cases by unit tests easier (see DependenciesGraphTest)
-internal class AppsDependenciesGraphTest {
```
eugene-krivobokov, line 11: No need to support a task only for testing. 

The cases are covered by other tests:
- AndroidAppsGraphBuilderTest - unit tests for graph
- FindAndroidApp... - integration tests with Gradle Test Kit. They are real user scenarious.

subprojects/gradle/module-dependencies-graph/src/gradleTest/kotlin/com/avito/module/dependencies/DependenciesGraphTestProject.kt
```diff
@@ -57,15 +57,15 @@ internal object DependenciesGraphTestProject {
                         ),
                     )
                 ),
-                AndroidLibModule(name = "LeafC", dependencies = setOf(GradleDependency.Safe.project(":LeafD"))),
-                AndroidLibModule(name = "LeafG", dependencies = setOf(GradleDependency.Safe.project(":LeafE"))),
-                KotlinModule(name = "LeafD", dependencies = setOf(GradleDependency.Safe.project(":LeafE"))),
```
eugene-krivobokov, line 62: In graph theory "leaf" node has no dependencies. 
This distinction also helps to understand the structure easier in tests.

subprojects/gradle/module-dependencies-graph/src/main/kotlin/com/avito/module/internal/dependencies/ModuleProjectConfigurationDependenciesNode.kt
```diff
@@ -17,15 +17,16 @@ internal class ModuleProjectConfigurationDependenciesNode(
     }
 
     fun allDependencies(): Set<Project> {
-        return dependencies.mapValues { (_, nodeSet) ->
-            nodeSet.flatMap { node ->
-                mutableSetOf(node.project).also {
-                    it.addAll(
-                        node.allDependencies()
```
eugene-krivobokov, line 24: Recursive `allDependencies` has terrible complexity. 
It repeatedly collects all dependencies in each node.

We had OOM in avito apps scale. Millions of visits\objects for hundreds of modules.

dsvoronin (approved):

sergioboi:
Prefer in future see a couple of PR comparing to that one

subprojects/gradle/module-dependencies-graph/src/main/kotlin/com/avito/module/internal/dependencies/DependenciesGraphBuilder.kt
```diff
@@ -60,7 +60,9 @@ internal class DependenciesGraphBuilder(
                                 dependenciesOnProjects(projectDependency.dependencyProject, ConfigurationType.Main)
                             }
                             .toSet()
-                    }.toMap()
+                    }
+                    .filter { (_, nodes) -> nodes.isNotEmpty() }
```
sergioboi: Filter break correctness of the graph, and I don't understand what this filter gives

subprojects/gradle/module-dependencies-graph/src/main/kotlin/com/avito/module/internal/dependencies/ModuleProjectConfigurationDependenciesNode.kt
```diff
@@ -17,15 +17,26 @@ internal class ModuleProjectConfigurationDependenciesNode(
...
-                    )
-                }
+        val dependencies = mutableSetOf<Project>()
```
sergioboi: It helps to avoid recursive visits only for one call `allDependencies`
But we call it for each root app, so we could reduce complexity more if we will use global cache where the key is node and value is all it dependencies

subprojects/gradle/module-dependencies-graph/src/main/kotlin/com/avito/module/internal/dependencies/DependenciesGraphBuilder.kt
eugene-krivobokov: Reverted this change

subprojects/gradle/module-dependencies-graph/src/main/kotlin/com/avito/module/internal/dependencies/AndroidAppsGraphBuilder.kt
```diff
@@ -6,17 +6,10 @@ internal class AndroidAppsGraphBuilder(
...
 
-    fun buildDependenciesGraph(type: ConfigurationType): Set<ModuleProjectConfigurationDependenciesNode> {
+    fun buildDependenciesGraph(type: ConfigurationType): Set<ProjectConfigurationNode> {
         return graphBuilder.buildDependenciesGraph(type)
             .filter { node ->
                 node.project.plugins.hasPlugin("com.android.application")
             }.toSet()
     }
-
-    fun buildDependenciesGraphFlatten(type: ConfigurationType): List<ProjectWithDeps> {
```
eugene-krivobokov, line 16: Remove from Android graph. Here is nothing about Android, only collecting data from nodes.

subprojects/gradle/module-dependencies-graph/src/main/kotlin/com/avito/module/internal/dependencies/ModuleProjectConfigurationDependenciesNode.kt
eugene-krivobokov: I've decided to keep it as it is.

Caching results between different apps is a controversial solution. 
It requires storing all transitive dependencies for each module. This is a lot of duplicated data that requires much more memory.

Traversing on direct dependencies took 2-4ms for an app. This is definitely not a bottleneck.

PR #874, opened 2021-03-23
Add detailed no suitable apps advice

sergioboi:


subprojects/gradle/module-dependencies-graph/src/gradleTest/kotlin/com/avito/module/dependencies/FindAndroidAppTaskAdvisorTest.kt
```diff
@@ -0,0 +1,166 @@
...
+            .isEqualTo(
+                """
+                |There are multiple suitable apps [:app, :appSecond]
+                |You should prefer :app because it has the least dependencies in graph size=1
+            """.trimMargin()
+            )
+    }
+
+    @Test
+    public fun `no suitable apps have equal deps`() {
```
sergioboi, line 74: From there tests for a new advice

subprojects/gradle/module-dependencies-graph/src/main/kotlin/com/avito/module/internal/dependencies/FindAndroidAppTaskAdvisor.kt
```diff
@@ -10,10 +10,31 @@ internal class FindAndroidAppTaskAdvisor {
...
+            .groupBy { it.missedModules.size }
+            .mapValues { (_, results) ->
+                results.sortedBy { it.dependencies.size }
+            }
+            .map { it.key to it.value }
+            .sortedBy { it.first }
+
+        return buildString {
+            appendLine("There are no suitable Android apps")
+            appendLine("Apps sorted by missing modules:")
```
eugene-krivobokov:
```suggestion
            appendLine("Apps are sorted by missing modules:")
```

subprojects/gradle/module-dependencies-graph/src/gradleTest/kotlin/com/avito/module/dependencies/FindAndroidAppTaskActionTest.kt
```diff
@@ -0,0 +1,63 @@
...
+
+    private val root = rootProject()
+    private val graphBuilder = AndroidAppsGraphBuilder(DependenciesGraphBuilder(root, StubLoggerFactory))
+    private val action = FindAndroidAppTaskAction(graphBuilder)
+
+    @Test
+    public fun `NoSuitableApps - Result contains all no suitable apps`() {
+        val rootApp = androidApp("rootApp", root)
+        val rootAppSecond = androidApp("rootAppSecond", root)
+        val a = javaLib("a", rootApp)
```
eugene-krivobokov, line 23: Can we use an Android application as a parent Gradle project?
I have doubts that it will work properly in a real project.

eugene-krivobokov (approved):

subprojects/gradle/module-dependencies-graph/src/gradleTest/kotlin/com/avito/module/dependencies/FindAndroidAppTaskActionTest.kt
sergioboi: In that case, this is not necessary because we don't build a graph. We need only project objects
Could you describe your doubts and maybe we need a special integration test

eugene-krivobokov: Screenshot from 2021-03-24 12-33-48

In this case, the library is inside the application. It works but looks weird. 

I suspect problems with impact analysis and build cache.
Anyway, I don't see reasons to support and test such cases.
This is why I want to avoid them everywhere.

PR #875, opened 2021-03-24
Mbs 10704 service metrics

dsvoronin:
see `Networking.md` for docs


subprojects/common/http-client/src/main/kotlin/com/avito/http/HttpClientProvider.kt
```diff
@@ -0,0 +1,72 @@
...
+        timeoutMs: Long? = null,
+        retryPolicy: RetryPolicy? = null,
+        builderTransform: OkHttpClient.Builder.() -> OkHttpClient.Builder = { this }
+    ): OkHttpClient {
+
+        val seriesName = SeriesName.create("service", serviceName)
+
+        return builder
+            .also { this.builderTransform(it) }
+            .also { builderTransform(it) }
```
eugene-krivobokov: What client is using both transformations? Is one is not enough?

subprojects/common/http-client/src/main/kotlin/com/avito/http/internal/ServiceMetricsInterceptor.kt
```diff
@@ -0,0 +1,28 @@
...
+
+        val latency = response.receivedResponseAtMillis - response.sentRequestAtMillis
+
+        listener.onResponse(
+            code = response.code,
+            latencyMs = latency
+        )
+        response
+    } catch (e: IOException) {
+        if (e is SocketTimeoutException) {
```
eugene-krivobokov: I think that this distinction is a client's responsibility.
Otherwise, the network layer knows about the usage of different errors. We duplicate decisions.

I propose to have only one method in interface: `onError(e: IOException)`

subprojects/common/http-client/src/main/kotlin/com/avito/http/internal/StatsdServiceEventsListener.kt
```diff
@@ -0,0 +1,27 @@
...
+
+internal class StatsdServiceEventsListener(
+    private val statsDSender: StatsDSender,
+    private val prefix: SeriesName
+) : ServiceEventsListener {
+
+    override fun onResponse(code: Int, latencyMs: Long) {
+        statsDSender.send(CountMetric(prefix.append("response", "$code")))
+        statsDSender.send(TimeMetric(prefix.append("latency"), latencyMs))
+    }
```
eugene-krivobokov: It looks that we don't need 2 metrics - response and latency. They are the same:

What is a response? It's a time metric with different results.

time: `<endpoint>.<result>`

- `api.method.200 : 25 ms`
- `api.method.502 : 10 ms`
- `api.method.timeout : 30_000 ms`

Time metric will give information about latency and count of different events.

subprojects/common/http-client/src/main/kotlin/com/avito/http/internal/ServiceMetricsInterceptor.kt
```diff
@@ -0,0 +1,28 @@
...
+import java.net.SocketTimeoutException
+
+internal class ServiceMetricsInterceptor(private val listener: ServiceEventsListener) : Interceptor {
+
+    override fun intercept(chain: Interceptor.Chain): Response = try {
+        val response = chain.proceed(chain.request())
+
+        val latency = response.receivedResponseAtMillis - response.sentRequestAtMillis
+
+        listener.onResponse(
```
eugene-krivobokov: We must distinguish requests to different methods.
It's important for troubleshooting.


- `api/upload` - 1k RPM, error rate 50%
- `api/fetch` - 10k RPM, error rate 0.1%

Without distinguishing we will see 11k RPM with error rate 0.4%. We would think that everything is ok.

subprojects/gradle/qapps/src/main/kotlin/com/avito/plugin/QAppsUploadAction.kt
```diff
@@ -34,10 +34,16 @@ internal class QAppsUploadAction(
         Retrofit.Builder()
             .baseUrl(host)
             .addConverterFactory(GsonConverterFactory.create(GsonBuilder().create()))
-            .client(httpClient())
+            .client(
+                httpClientProvider.provide(
+                    "qapps",
+                    TimeUnit.SECONDS.toMillis(60),
```
eugene-krivobokov: Let's use `java.time.Duration` instead of long

- It's safer, no need to keep a dimension in the name - `...Ms` and so on.
- It looks better, not inversed

```kotlin
TimeUnit.SECONDS.toMillis(60)

Duration.ofSeconds(60)
```

subprojects/common/http-client/src/main/kotlin/com/avito/http/HttpClientProvider.kt
```diff
@@ -0,0 +1,72 @@
...
+public class HttpClientProvider(
+    private val statsDSender: StatsDSender,
+    private val loggerFactory: LoggerFactory,
+    private val builderTransform: OkHttpClient.Builder.() -> OkHttpClient.Builder = { this }
+) {
+
+    private val builder = OkHttpClient.Builder()
+
+    public fun provide(
+        serviceName: String,
```
eugene-krivobokov: Why don't we take it from the request? 
A client can lie (make a mistake), but a request can't.

subprojects/gradle/qapps/src/main/kotlin/com/avito/plugin/QAppsUploadAction.kt
dsvoronin: it is , but we still not supported java8 api's

subprojects/common/http-client/src/main/kotlin/com/avito/http/HttpClientProvider.kt
dsvoronin: Did you think how to transform request to meaningful series already?

subprojects/common/http-client/src/main/kotlin/com/avito/http/internal/ServiceMetricsInterceptor.kt
dsvoronin: agreed. moved this if to `StatsdServiceEventsListener`

subprojects/gradle/qapps/src/main/kotlin/com/avito/plugin/QAppsUploadAction.kt
eugene-krivobokov: Can't we use it inside Gradle plugins?

subprojects/common/http-client/src/main/kotlin/com/avito/http/internal/StatsdServiceEventsListener.kt
dsvoronin: You're right, will do along with per-method metrics

subprojects/common/http-client/src/main/kotlin/com/avito/http/HttpClientProvider.kt
dsvoronin: One is enough from API perspective, but constructor one needed for testing purposes;
i'll try to get rid of it

subprojects/common/http-client/src/main/kotlin/com/avito/http/HttpClientProvider.kt
eugene-krivobokov: I thought about simple mapping for REST:

`http://integration.api/storage/files/upload` -> `integration_api.storage_files_upload`

And the ability to override method description for JSON RPC:

`http://integration.api/action` -> `integration_api.<request descriptor>`

"request descriptor" is calculated in advance and passed by tag: https://github.com/avito-tech/avito-android/pull/841#discussion_r594924626

Some additional considerations.

Why do we need "serviceName" at all? I see only one reason, to make it stable against changes in the host.
`http://integration.api` -->`http://proxy-integration.api`
We can live without it. 
A new endpoint means a different route. I'd like to distinguish metrics.

Generated metric names can be easily grouped by the host in Grafana.

subprojects/gradle/qapps/src/main/kotlin/com/avito/plugin/QAppsUploadAction.kt
dsvoronin: HttpClientProvider also provide for android runtime

subprojects/common/http-client/src/main/kotlin/com/avito/http/HttpClientProvider.kt
dsvoronin: we have api's with schemes like: `http://prod.k/service-name/api/methods` and `http://service.name/api/methods`
i don't know how to group this properly

subprojects/common/http-client/src/main/kotlin/com/avito/http/HttpClientProvider.kt
```diff
@@ -0,0 +1,25 @@
...
+public class HttpClientProvider(
+    private val statsDSender: StatsDSender,
+    private val timeProvider: TimeProvider
+) {
+
+    private val builder = OkHttpClient.Builder()
+
+    public fun provide(): OkHttpClient.Builder {
+        return builder.eventListenerFactory {
+            StatsHttpEventListener(
```
eugene-krivobokov, line 19: We propagate all dependencies for StatsHttpEventListener in HttpClientProvider.
Let's use DI principle and pass EventListener to HttpClientProvider.

subprojects/common/http-client/src/main/kotlin/com/avito/http/internal/StatsHttpEventListener.kt
```diff
@@ -0,0 +1,59 @@
...
+import okhttp3.Response
+import java.io.IOException
+import java.net.SocketTimeoutException
+
+internal class StatsHttpEventListener(
+    private val statsDSender: StatsDSender,
+    private val timeProvider: TimeProvider
+) : EventListener() {
+
+    var callStarted = 0L
```
eugene-krivobokov, line 24: Shouldn't we use Factory to make it thread-safe (unique for each Call)?

https://square.github.io/okhttp/events/#eventlistenerfactory

docs/content/contributing/Networking.md
```diff
@@ -0,0 +1,60 @@
+# Networking
+
+We using [Okhttp](https://square.github.io/okhttp/) sometimes with [Retrofit](https://square.github.io/retrofit/) for
+networking in gradle plugins and android.
```
eugene-krivobokov:
```suggestion
networking in Gradle plugins and Android libraries.
```

docs/content/contributing/Networking.md
```diff
@@ -0,0 +1,60 @@
+# Networking
+
+We using [Okhttp](https://square.github.io/okhttp/) sometimes with [Retrofit](https://square.github.io/retrofit/) for
```
eugene-krivobokov:
```suggestion
We use [Okhttp](https://square.github.io/okhttp/) sometimes with [Retrofit](https://square.github.io/retrofit/) for
```

docs/content/contributing/Networking.md
```diff
@@ -0,0 +1,60 @@
...
+}
+```
+
+Create an instance of `HttpClientProvider`
+
+```kotlin
+val httpClientProvider = HttpClientProvider(statsdSender, timeProvider)
+```
+
+Get client
```
eugene-krivobokov:
```suggestion
Get the client:
```

docs/content/contributing/Networking.md
```diff
@@ -0,0 +1,60 @@
+# Networking
+
+We using [Okhttp](https://square.github.io/okhttp/) sometimes with [Retrofit](https://square.github.io/retrofit/) for
+networking in gradle plugins and android.
+
+## Obtain an OkHttpClient
```
eugene-krivobokov:
```suggestion
## Obtaining an OkHttpClient
```
Because it's a scenario but not a request or command

docs/content/contributing/Networking.md
```diff
@@ -0,0 +1,60 @@
...
+
+Get client
+
+```kotlin
+val httpClient = httpClientProvider.provide().build()
+```
+
+Method provide() returns `OkHttpClient.Builder` so you can configure it further
+
+## Required tag
```
eugene-krivobokov, line 30: Is it on the same level in menu as "Obtain an OkHttpClient" scenario?
It looks like a sub scenario in creation.

docs/content/contributing/Networking.md
```diff
@@ -0,0 +1,60 @@
...
+
+Obtained client needs some additional information to be able to gather required statistics.
+
+Every request should contain a tag of type `RequestMetadata`
+
+???+ warning 
+    Missing tag will lead to missing metrics for this api method / service 
+    Warning message could be found in logs
```
eugene-krivobokov, line 38: No one reads logs. We need a way to not losing metrics. It's easy to forget.
Can't we make it fail-fast?

subprojects/common/http-client/src/main/kotlin/com/avito/http/HttpClientProvider.kt
eugene-krivobokov: Fixed

docs/content/contributing/Networking.md
eugene-krivobokov: The alternative is an opt-out if the client doesn't need metrics for a method.

sergioboi: I propose to send metric with `no-tag` it will help us to see it on dashboard

subprojects/common/http-client/src/main/kotlin/com/avito/http/internal/StatsHttpEventListener.kt
dsvoronin: i am using factory

subprojects/common/file-storage/src/main/kotlin/com/avito/filestorage/RemoteStorageFactory.kt
```diff
@@ -17,40 +13,19 @@ object RemoteStorageFactory {
...
         endpoint: String,
+        httpClientProvider: HttpClientProvider,
         loggerFactory: LoggerFactory,
         timeProvider: TimeProvider,
-        httpClient: OkHttpClient = getHttpClient(
-            logger = loggerFactory.create<RemoteStorage>(),
-            retryInterceptor = null,
-            verbose = false // do not enable for production, generates a ton of logs
-        )
+        testInterceptor: Interceptor? = null,
```
sergioboi: This is an abstraction leak. You could wrap the `httpClientProvider` to return the client with applied testInterceptor

subprojects/common/http-client/src/main/kotlin/com/avito/http/internal/StatsHttpEventListener.kt
```diff
@@ -0,0 +1,71 @@
...
+                statsDSender.send(TimeMetric(prefix.append("unknown"), latencyMs))
+            }
+        }
+    }
+
+    private fun serviceMetric(request: Request): SeriesName? {
+        val metadata = request.tag(RequestMetadata::class.java)
+
+        return if (metadata == null) {
+            logger.warn(
```
sergioboi, line 57: It won't help us to find what request missed the `tag`

docs/content/contributing/Networking.md
dsvoronin: > Can't we make it fail-fast?

fail fast is prohibited in event listener

https://square.github.io/okhttp/4.x/okhttp/okhttp3/-event-listener/
```
All event methods must execute fast, without external locking, cannot throw exceptions, attempt to mutate the event parameters, or be re-entrant back into the client. Any IO - writing to files or network should be done asynchronously.
```

> I propose to send metric with no-tag it will help us to see it on dashboard

i like it, will do

subprojects/common/http-client/src/test/kotlin/com/avito/http/HttpClientProviderMetricsTest.kt
```diff
@@ -0,0 +1,130 @@
...
+                actual.name == expected.name
+            } else false
+        },
+        "metric names are equal"
+    )
+
+    private val doesNotMatter = 111L
+
+    @Test
+    fun `timeout metric`() {
```
sergioboi: Let's test a few parallel different request to see that Event listener handle that

dsvoronin: will do

subprojects/common/file-storage/src/main/kotlin/com/avito/filestorage/RemoteStorageFactory.kt
dsvoronin: I did that in first place, but it's also overcomplicated

subprojects/common/http-client/src/main/kotlin/com/avito/http/internal/StatsHttpEventListener.kt
dsvoronin: i agree, will add more info

subprojects/common/statsd/src/main/kotlin/com/avito/android/stats/StatsMetric.kt
```diff
@@ -12,6 +12,28 @@ class TimeMetric(
 ) : StatsMetric() {
     override val value: Long = timeInMs
     override val type = "time"
+
+    override fun equals(other: Any?): Boolean {
```
sergioboi: Have you tried a `data class`?

dsvoronin: data class should work

subprojects/common/http-client/src/main/kotlin/com/avito/http/HttpClientProvider.kt
dsvoronin: on second though it doesn't make much sense 
`HttpClientProvider` is exactly the place we want to hide concrete listener creation

subprojects/common/okhttp/src/main/kotlin/com/avito/http/RetryInterceptor.kt
```diff
@@ -23,9 +23,8 @@ public class RetryInterceptor constructor(
     ),
     private val delayMs: Long = TimeUnit.SECONDS.toMillis(1),
     private val useIncreasingDelay: Boolean = true,
-    private val logger: Logger,
-    private val modifyRetryRequest: (Request) -> Request = { it },
-    private val describeRequest: (Request) -> String = { it.url.redact() }
+    private val logger: Logger, // todo remove after release
+    private val modifyRetryRequest: (Request) -> Request = { it }
```
sergioboi, line 27: Maybe OkhttpClient has some API for modifications. Because we use it only with Elastic

sergioboi (approved):

subprojects/common/file-storage/src/main/kotlin/com/avito/filestorage/RemoteStorageFactory.kt
dsvoronin: removed it

subprojects/common/okhttp/src/main/kotlin/com/avito/http/RetryInterceptor.kt
dsvoronin: it should be applied only to 2+ try inside retry interceptor

eugene-krivobokov (approved):

subprojects/common/http-client/src/main/kotlin/com/avito/http/HttpClientProvider.kt
sergioboi: I don't think so, but I don't want to spend time arguing so do what you should

subprojects/common/okhttp/src/main/kotlin/com/avito/http/RetryInterceptor.kt
sergioboi: I want you to notice that `retry` and `modify` are two different intentions, so I don't like this coupling
But feel free to ignore my opinion

subprojects/common/http-client/src/test/kotlin/com/avito/http/HttpClientProviderMetricsTest.kt
```diff
@@ -0,0 +1,185 @@
...
+            httpClient.blockingCall()
+        }
+
+        assertThat(statsDSender.getSentMetrics()).comparingElementsUsing(metricNamesCorrespondence).containsExactly(
+            TimeMetric(SeriesName.create("service", "some-service", "some-method", "timeout"), doesNotMatter)
+        )
+    }
+
+    @Test
+    fun `all metrics send - multiple parallel requests`() {
```
eugene-krivobokov, line 68: This test is flaky.
Please, take a look at Teamcity builds for #939 PR.

dsvoronin: Created MBS-11108 for this

eugene-krivobokov: Thanks!

subprojects/common/okhttp/src/main/kotlin/com/avito/http/RetryInterceptor.kt
dsvoronin: Wrapper around `RetryInterceptor`, that will add request modification capability should be ok?
I don't know how to modify request after first try, because knowledge about tries encapsulated inside this interceptor

Maybe it could be done using tags. What do you think?

Or you don't understand why anyone could want to modify request on different tries at all?

PR #876, opened 2021-03-25
Collect data about common dependencies between apps

eugene-krivobokov:
Collecting data about modules dependencies to find duplications.
These are not final metrics for tracking, only extra information for further researches.

- Total dependencies in applications
- Proportion of common dependencies between applications. This shows duplicates.

The next steps are:
- Find root problems in heavy and duplicated applications
- Decide what can we do to track and fix them

### DIFF
--- a/subprojects/gradle/module-dependencies-graph/src/main/kotlin/com/avito/module/metrics/metrics/AppsHealthData.kt
+++ b/subprojects/gradle/module-dependencies-graph/src/main/kotlin/com/avito/module/metrics/metrics/AppsHealthData.kt
@@ -1,0 +1,39 @@
+package com.avito.module.metrics.metrics
+
+import com.avito.math.Percent
+import com.avito.math.percentOf
+import org.gradle.util.Path
+
+internal data class AppsHealthData(
+    val absolute: Map<Path, AbsoluteMetrics>,
+    val relative: Matrix<Path, RelativeMetrics>
+)
+
+internal class AbsoluteMetrics(
+    val allDependencies: Int
+)
+
+internal class RelativeMetrics(
+    val baselineDependencies: Set<Path>,
+    val comparedDependencies: Set<Path>,
+) {
+
+    val commonDependencies: Set<Path>
+        get() = baselineDependencies.intersect(comparedDependencies)
+
+    val uniqueDependencies: Set<Path>
+        get() = baselineDependencies.minus(commonDependencies)
+
+    /**
+     * Proportion of compared dependencies included to baseline dependencies
+     *
+     * Example:
+     * application X has modules A, B
+     * application Y has modules B
+     *
+     * X -> Y: 50% (includes 1 out of 2 dependencies)
+     * Y -> X: 100% (includes 1 out of 1)
+     */
+    val commonDependenciesRatio: Percent
+        get() = commonDependencies.size.percentOf(baselineDependencies.size)
+}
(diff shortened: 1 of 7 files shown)


subprojects/gradle/module-dependencies-graph/src/main/kotlin/com/avito/module/metrics/metrics/AppsHealthData.kt
```diff
@@ -0,0 +1,39 @@
...
+     *
+     * Example:
+     * application X has modules A, B
+     * application Y has modules B
+     *
+     * X -> Y: 50% (includes 1 out of 2 dependencies)
+     * Y -> X: 100% (includes 1 out of 1)
+     */
+    val commonDependenciesRatio: Percent
+        get() = commonDependencies.size.percentOf(baselineDependencies.size)
```
eugene-krivobokov, line 38: The key information - the proportion of common modules between apps.

subprojects/gradle/module-dependencies-graph/src/main/kotlin/com/avito/module/metrics/metrics/Matrix.kt
```diff
@@ -0,0 +1,25 @@
+package com.avito.module.metrics.metrics
+
+internal class Matrix<KEY, DATA> {
```
sergioboi: I propose to rename KEY to COORDINATE
because it's a Matrix. Matrix should have two coordinate and value

subprojects/gradle/module-dependencies-graph/src/main/kotlin/com/avito/module/metrics/metrics/Matrix.kt
```diff
@@ -0,0 +1,25 @@
...
+
+    private val data = mutableMapOf<KEY, MutableMap<KEY, DATA>>()
+
+    fun keys(): Set<KEY> {
```
sergioboi: It's not clear what will return a fun `keys` on a Matrix

subprojects/gradle/module-dependencies-graph/src/main/kotlin/com/avito/module/metrics/metrics/AppsHealthData.kt
```diff
@@ -0,0 +1,39 @@
+package com.avito.module.metrics.metrics
+
+import com.avito.math.Percent
+import com.avito.math.percentOf
+import org.gradle.util.Path
+
+internal data class AppsHealthData(
```
sboishtyan, line 7: Let's add unit tests for those calculations

sergioboi (approved):

subprojects/gradle/module-dependencies-graph/src/main/kotlin/com/avito/module/metrics/metrics/Matrix.kt
eugene-krivobokov: Renamed

subprojects/gradle/module-dependencies-graph/src/main/kotlin/com/avito/module/metrics/metrics/Matrix.kt
eugene-krivobokov: The idea was to get all coordinates.

Removed the function. Split to rows and keys coordinates.

sergioboi: I understand what this function does. I try to image like a user: if I see a function invocation will I understand what it returns?

PR #881, opened 2021-03-26
MBS-10204 test maxparallelworkers=1 for gradle test kit

dsvoronin:


build-logic/testing-convention/src/main/kotlin/convention.gradle-testing.gradle.kts
```diff
@@ -32,8 +32,7 @@ val gradleTestTask = tasks.register<Test>("gradleTest") {
 
     useJUnitPlatform()
 
-    @Suppress("MagicNumber")
-    maxParallelForks = 8
+    maxParallelForks = 1
```
eugene-krivobokov, line 35: Did it help?

dsvoronin: will do profiler runs

PR #891, opened 2021-03-29
Metrics for affected applications in impact analysis

eugene-krivobokov:
## The context

We want to understand the efficiency of impact analysis.
Now we collect affected modules in all configuration types.

These are real values after changes in SERP module:
```
<prefix>.impact.implementation.all:435
<prefix>.impact.implementation.modified:21
<prefix>.impact.unittests.all:435
<prefix>.impact.unittests.modified:21
<prefix>.impact.androidtests.all:435
<prefix>.impact.androidtests.modified:25
<prefix>.impact.lint.all:435
<prefix>.impact.lint.modified:25
```

## The problem

The count of modules is impractical when we have hundreds of them. 
The value is volatile and has no strong correlation with heavy tasks in applications.

For the "main" configuration:
Screenshot from 2021-03-30 10-45-39

## Changes

- Replaced absolute values by percentages. 
Absolute values are useless without a full-detailed report for impact analysis. 
Example: 75 modules changed: is it good? why 75?
I propose to use these metrics only for long-term trends and nothing more.

Sample values after the same changes in SERP module:
```
<prefix>.impact.modules.implementation.modified:4 <--- %
<prefix>.impact.modules.unittests.modified:4
<prefix>.impact.modules.androidtests.modified:5
<prefix>.impact.modules.lint.modified:5
```

- Added metrics for applications. 
They are more important. They contain heavy tasks in a critical path.

Sample values after the same changes in SERP module:
```
<prefix>.impact.apps.implementation.modified:42
<prefix>.impact.apps.unittests.modified:42
<prefix>.impact.apps.androidtests.modified:61
<prefix>.impact.apps.lint.modified:61
```

Compare the values:

```
<prefix>.impact.modules.implementation.modified:4 %
<prefix>.impact.apps.implementation.modified:42 %
```

I've changed one module. 
It affects 4% of all modules. It looks ok. 
But these modules affect 42% of all applications. It highlights the impact.


subprojects/gradle/impact/src/main/kotlin/com/avito/impact/plugin/ImpactMetricsSender.kt
```diff
@@ -1,59 +0,0 @@
...
-import com.avito.android.sentry.EnvironmentInfo
-import com.avito.android.stats.GaugeLongMetric
-import com.avito.android.stats.SeriesName
-import com.avito.android.stats.StatsDSender
-import com.avito.impact.ModifiedProjectsFinder
-import com.avito.impact.configuration.internalModule
-import com.avito.module.configurations.ConfigurationType
-import com.avito.utils.gradle.Environment
-
-class ImpactMetricsSender(
```
eugene-krivobokov, line 12: Moved to internal package

subprojects/gradle/impact/src/main/kotlin/com/avito/impact/plugin/internal/ImpactMetricsSender.kt
```diff
@@ -0,0 +1,90 @@
...
+
+    private fun sendModulesMetrics(
+        configurationType: ConfigurationType,
+        projects: Set<Project>,
+        modified: Set<ModifiedProject>
+    ) {
+        sendMetric(
+            metric = SeriesName.create("modules", configurationType.name.toLowerCase(Locale.US), "modified"),
+            value = modified.size.percentOf(projects.size).toLong()
+        )
```
eugene-krivobokov, line 68: Breaking change in old metrics for affected modules: 
- Added `modules` prefix
- Converted to percentages

Don't preserve old metrics because they are volatile and we didn't use them for making decisions.

subprojects/gradle/impact/src/main/kotlin/com/avito/impact/plugin/internal/ImpactMetricsSender.kt
```diff
@@ -0,0 +1,90 @@
...
+        configurationType: ConfigurationType,
+        projects: Set<Project>,
+        modified: Set<ModifiedProject>
+    ) {
+        val apps = projects.count { it.isAndroidApp() }
+        val modifiedApps = modified.count { it.project.isAndroidApp() }
+        sendMetric(
+            metric = SeriesName.create("apps", configurationType.name.toLowerCase(Locale.US), "modified"),
+            value = modifiedApps.percentOf(apps).toLong()
+        )
```
eugene-krivobokov, line 81: New metrics for affected Android applications.

subprojects/gradle/test-project/src/main/kotlin/com/avito/test/gradle/InMemoryProjectGenerator.kt
```diff
@@ -1,15 +1,15 @@
-package com.avito.module.dependencies
+package com.avito.test.gradle
 
 import org.gradle.api.Project
 import org.gradle.api.plugins.JavaLibraryPlugin
 import org.gradle.testfixtures.ProjectBuilder
```
eugene-krivobokov, line 6: Moved test fixtures for `ProjectBuilder` to `test-project` module.
There are lots of needed improvements in API to make them more reusable, but it's out of scope.

subprojects/gradle/impact/src/gradleTest/kotlin/com/avito/impact/plugin/internal/ImpactMetricsSenderTest.kt
```diff
@@ -0,0 +1,84 @@
...
+            it.name.toString().contains("impact.apps.implementation.modified")
+        }.also { metric ->
+            verifyGaugeValue(metric, 0)
+        }
+
+        metrics.find {
+            it.name.toString().contains("impact.apps.androidtests.modified")
+        }.also { metric ->
+            verifyGaugeValue(metric, 50)
+        }
```
eugene-krivobokov, line 72: Test case for impact metrics.

Stub impact analysis results because we use its results as clients.

subprojects/gradle/impact-shared/src/main/java/com/avito/impact/ModifiedProjectsFinder.kt
```diff
@@ -27,46 +47,10 @@ class ModifiedProjectsFinder(
...
-                    it.internalModule.isModified(type)
-                }
-            }.toSet()
-    }
-
-    @Deprecated(
-        "Используется только для поиска по ReportType.ANDROID_TESTS. " +
-            "Оптимизация для UI тестов, явно игнорируем изменения в реализации, чтобы не сваливаться всегда в fallback"
-    )
-    fun findModifiedProjectsWithoutDependencyToAnotherConfigurations(
```
eugene-krivobokov, line 48: Removed as not used

subprojects/gradle/impact-shared/src/main/java/com/avito/impact/ModifiedProjectsFinder.kt
```diff
@@ -27,46 +47,10 @@ class ModifiedProjectsFinder(
...
-                configuration.dependencies.any { dependency -> dependency.isModified }
-                    || configuration.hasChangedFiles
-            }
-        }
-
-    private fun findProjects(
-        rootProject: Project,
-        configurationType: ConfigurationType,
-        predicate: (project: Project) -> Boolean
-    ): Set<ModifiedProject> {
```
eugene-krivobokov, line 62: Here was a mistake. The function could return all projects as modified.

```
20 line: findProjects(rootProject, type, predicate = { true })
```

This is why I've introduced an explicit API to get all analyzed projects: `allProjects()`.

dsvoronin (approved):

sergioboi:
The only problem that current values are useless because they calculated wrong.
Consider next step is refactoring that how we parse `ConfigurationType`

eugene-krivobokov:
> The only problem that current values are useless because they calculated wrong.
> Consider next step is refactoring that how we parse `ConfigurationType`

Not useless completely, but agree. Will file an issue.

PR #893, opened 2021-03-29
Mbs 10704 fix metrics from device

dsvoronin:
Fix BundleArgsProvider: int value was not parsed properly (all instrumentation args are strings, but `bundle.getInt()` was expected)

Fix 1sec timeout for signer service, not enough for upload download apk; correct value will be set based on metrics

### DIFF
--- a/subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/runner/InHouseInstrumentationTestRunner.kt
+++ b/subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/runner/InHouseInstrumentationTestRunner.kt
@@ -95,10 +95,7 @@ abstract class InHouseInstrumentationTestRunner :
      */
     val testRunEnvironment: TestRunEnvironment by lazy {
         if (isRealRun(instrumentationArguments)) {
-            logger.debug("Instrumentation arguments: $instrumentationArguments")
-            val result = createRunnerEnvironment(instrumentationArguments)
-            logger.debug("TestRunEnvironment: $result")
-            result
+            createRunnerEnvironment(instrumentationArguments)
         } else {
             TestRunEnvironment.OrchestratorFakeRunEnvironment
         }
@@ -216,8 +213,10 @@ abstract class InHouseInstrumentationTestRunner :
     }
 
     override fun beforeOnCreate(arguments: Bundle) {
-        injectTestMetadata(arguments)
+        injectTestMetadata(instrumentationArguments)
+        logger.debug("Instrumentation arguments: $instrumentationArguments")
         val environment = testRunEnvironment.asRunEnvironmentOrThrow()
+        logger.debug("TestRunEnvironment: $environment")
         initApplicationCrashHandling()
         addReportListener(arguments)
         initTestCase(environment)
--- a/subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/runner/InHouseInstrumentationTestRunner.kt
+++ b/subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/runner/InHouseInstrumentationTestRunner.kt
@@ -95,7 +95,10 @@ abstract class InHouseInstrumentationTestRunner :
      */
     val testRunEnvironment: TestRunEnvironment by lazy {
         if (isRealRun(instrumentationArguments)) {
-            createRunnerEnvironment(instrumentationArguments)
+            logger.debug("Instrumentation arguments: $instrumentationArguments")
+            val result = createRunnerEnvironment(instrumentationArguments)
+            logger.debug("TestRunEnvironment: $result")
+            result
         } else {
             TestRunEnvironment.OrchestratorFakeRunEnvironment
         }
@@ -214,8 +217,6 @@ abstract class InHouseInstrumentationTestRunner :
 
     override fun beforeOnCreate(arguments: Bundle) {
         injectTestMetadata(arguments)
-        logger.debug("Instrumentation arguments: $arguments")
-        logger.debug("TestRunEnvironment: $testRunEnvironment")
         val environment = testRunEnvironment.asRunEnvironmentOrThrow()
         initApplicationCrashHandling()
         addReportListener(arguments)
--- a/subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/runner/TestRunEnvironment.kt
+++ b/subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/runner/TestRunEnvironment.kt
@@ -183,19 +183,24 @@ private fun parseSentryConfig(argumentsProvider: ArgsProvider): SentryConfig {
     }
 }
 
-private fun parseStatsDConfig(argumentsProvider: ArgsProvider): StatsDConfig {
+internal fun parseStatsDConfig(argumentsProvider: ArgsProvider): StatsDConfig {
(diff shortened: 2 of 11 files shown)


subprojects/gradle/signer/src/main/kotlin/com/avito/plugin/SignArtifactTask.kt
```diff
@@ -48,9 +48,9 @@ abstract class SignArtifactTask @Inject constructor(objects: ObjectFactory) : De
             timeProvider = timeProvider,
             loggerFactory = loggerFactory
         ).provide()
-            .connectTimeout(TIMEOUT_SEC, TimeUnit.SECONDS)
-            .writeTimeout(TIMEOUT_SEC, TimeUnit.SECONDS)
-            .readTimeout(TIMEOUT_SEC, TimeUnit.SECONDS)
+            .connectTimeout(10L, TimeUnit.SECONDS)
+            .writeTimeout(40L, TimeUnit.SECONDS)
+            .readTimeout(40L, TimeUnit.SECONDS)
```
eugene-krivobokov, line 53: Why so big timeout for reading? Do we see the problem in metrics?

eugene-krivobokov: If I'm not mistaken read timeout defines a maximum time of inactivity between two data packets when waiting for the server's response. If so, it shouldn't depend much on uploaded file size.

eugene-krivobokov (approved):

sergioboi (approved):

subprojects/gradle/signer/src/main/kotlin/com/avito/plugin/SignArtifactTask.kt
dsvoronin: yes, i just added a big enough value to fix it, and then gather metrics to set correct value

subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/runner/TestRunEnvironment.kt
```diff
@@ -183,19 +183,24 @@ private fun parseSentryConfig(argumentsProvider: ArgsProvider): SentryConfig {
     }
 }
 
-private fun parseStatsDConfig(argumentsProvider: ArgsProvider): StatsDConfig {
+internal fun parseStatsDConfig(argumentsProvider: ArgsProvider): StatsDConfig {
     val host = argumentsProvider.getOptionalArgument("statsDHost")
-    val port = argumentsProvider.getOptionalIntArgument("statsDPort")
```
eugene-krivobokov, line 188: Why we didn't fail here?

subprojects/android-test/test-inhouse-runner/src/test/kotlin/com/avito/android/runner/TestRunEnvironmentKtTest.kt
```diff
@@ -0,0 +1,22 @@
...
+
+import com.avito.android.stats.StatsDConfig
+import com.avito.truth.isInstanceOf
+import com.google.common.truth.Truth.assertThat
+import org.junit.jupiter.api.Test
+
+internal class TestRunEnvironmentKtTest {
+
+    @Test
+    fun `statsdconfig enabled - if required params passed`() {
```
eugene-krivobokov, line 11: We need also a negative scenario

dsvoronin: yep, let me add it

subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/runner/TestRunEnvironment.kt
dsvoronin: The code was:

```
 override fun getOptionalIntArgument(name: String): Int? {
        val defaultValue = Int.MIN_VALUE
        val result = bundle.getInt(name, defaultValue)
        return if (result == defaultValue) {
            null
        } else {
            result
        }
    }
```

We just returned null and fell to `Disabled`

subprojects/android-test/test-inhouse-runner/src/test/kotlin/com/avito/android/runner/TestRunEnvironmentKtTest.kt
dsvoronin: done

subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/runner/TestRunEnvironment.kt
eugene-krivobokov: I meant that we should fail but not return defaults.
I propose to use defaults only for absent values.
We can check it by `containsKey(name)` to distinguish cases.

dsvoronin: you mean introduce boolean like "useStatsd" and then required params if true?

eugene-krivobokov: > introduce boolean like "useStatsd" and then required params if true

Yes, it will be easier than juggling multiple optional arguments.

subprojects/android-test/test-inhouse-runner/src/test/kotlin/com/avito/android/runner/StubArgsProvider.kt
```diff
@@ -0,0 +1,33 @@
+package com.avito.android.runner
+
+import com.avito.android.test.report.ArgsProvider
+import java.io.Serializable
+
+internal class StubArgsProvider : ArgsProvider {
```
sergioboi, line 6: Fake

subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/runner/TestRunEnvironment.kt
eugene-krivobokov: There is another option. 
We can check it even for an optional argument, but only if a bundle contains the key.

subprojects/android-test/test-inhouse-runner/src/test/kotlin/com/avito/android/runner/StubArgsProvider.kt
dsvoronin: We decided to use Stub everywhere https://github.com/avito-tech/avito-android/discussions/679#discussioncomment-214813

subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/runner/InHouseInstrumentationTestRunner.kt
```diff
@@ -95,10 +95,7 @@ abstract class InHouseInstrumentationTestRunner :
      */
     val testRunEnvironment: TestRunEnvironment by lazy {
         if (isRealRun(instrumentationArguments)) {
-            logger.debug("Instrumentation arguments: $instrumentationArguments")
```
sergioboi: We can't log there because loggerFactory needs the testRunEnvironment

PR #900, opened 2021-04-01
New metrics for running tests

eugene-krivobokov:


docs/content/test_runner/Metrics.md
```diff
@@ -79,36 +79,52 @@ graph LR
...
+
+#### `success`
+
+#### `failure`
+
+#### `skipped`
+
+#### `manual`
+
+#### `lost.`
```
eugene-krivobokov, line 99: I'd like to collect information about runs by all statuses to see the whole picture.
The idea is to visualize the proportion of different statuses and see it in a long term.

Described in docs before implementation to discuss options.
@sboishtyan , @dsvoronin , are you ok with these new metrics? Will do?

dsvoronin: Some of these statuses are in test runner domain, some in report-viewer
Do you really want to mix it?

eugene-krivobokov: No, I'd like to avoid mixing them. Please help me to distinguish them.

Now we send only `lost` cases. They know about report-viewer, isn't it? E.g. `lost.no-file` from the device.

`TestsRunner` returns `SimpleRunTest` as a result. It is also a model from report-viewer.

PR #902, opened 2021-04-01
Compose test action finalizer

eugene-krivobokov:
Split `InstrumentationTestActionFinalizer` responsibilities to make it more composable and testable.


subprojects/common/result/src/main/kotlin/com/avito/android/Result.kt
```diff
@@ -106,7 +106,7 @@ sealed class Result<T> {
 
         override fun component1(): T? = null
 
-        override fun component2(): Throwable? = throwable
+        override fun component2(): Throwable = throwable
```
dsvoronin, line 109: Why so? parent function returns nullable

dsvoronin (approved):

subprojects/common/result/src/main/kotlin/com/avito/android/Result.kt
eugene-krivobokov: If a compiler can infer a more precise type, let's do it.
This use case for Result is rare but valid. 

```kotlin
val result = Result.tryCatch<String> { throw RuntimeException() }
    
if (result is Result.Failure) {
    val (_, error: Throwable) = result
}
```

PR #905, opened 2021-04-02
Project accessors

dsvoronin:
https://docs.gradle.org/7.0-rc-2/userguide/declaring_dependencies.html#sec:type-safe-project-accessors

IMO very useful feature, kotlin dsl just had +1 strong argument


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.8.3-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.0-rc-2-all.zip
```
eugene-krivobokov: Do you want to wait for a stable version?

I'm worried about the extra cost of supporting branches alive.
Personally, I think supporting it in advance unlikely worth it in a cost/benefit ratio.

dsvoronin: gradle 7 could not work with current agp in theory, so it's more like a research of how useful this feature, and how does IDE work with it

support of this particular feature branch is super low cost: on merge just apply remote branch changes and manually add accessors

dsvoronin:
There is too much to merge here, it will be simplier to just do it all over again

PR #908, opened 2021-04-02
MBS-10798 ignore clear directory on device errors

dsvoronin:
shouldn't lead to lost tests

### DIFF
--- a/subprojects/gradle/runner/client/src/main/kotlin/com/avito/runner/scheduler/listener/ArtifactsTestListener.kt
+++ b/subprojects/gradle/runner/client/src/main/kotlin/com/avito/runner/scheduler/listener/ArtifactsTestListener.kt
@@ -43,12 +43,18 @@ internal class ArtifactsTestListener(
         targetPackage: String,
         result: TestCaseRun.Result,
         durationMilliseconds: Long,
-        executionNumber: Int
+        executionNumber: Int,
+        testMetadataDirectory: File,
+        testFolder: String
     ) {
         val testResult = when (result) {
             Passed,
             is Failed.InRun ->
-                pullArtifacts(device, test, targetPackage)
+                pullArtifacts(
+                    device = device,
+                    runnerOutputDirectory = testMetadataDirectory,
+                    testFolder = testFolder
+                )
             is InfrastructureError ->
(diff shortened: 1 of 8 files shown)


subprojects/gradle/runner/client/src/main/kotlin/com/avito/runner/scheduler/listener/ArtifactsTestListener.kt
```diff
@@ -93,7 +93,7 @@ internal class ArtifactsTestListener(
 
             device.clearDirectory(
                 remotePath = testMetadataDirectory.toPath()
-            ).getOrThrow()
+            ).getOrElse { /* ignore errors */ }
```
eugene-krivobokov: Why can we ignore cleaning?
Won't it break the next test run on the device?

dsvoronin: You're right, i thought we somehow cleared state using snapshots, but i was wrong, `om.avito.runner.service.model.intention.State` only knows about device api level, model and installed apps

Both old and new variant are wrong, we need to remove device from pool, but it shouldn't affect current test result

dsvoronin: it is only test metadata directory, but anyway it could lead to undesired state

eugene-krivobokov (approved):

subprojects/gradle/runner/service/src/main/kotlin/com/avito/runner/service/worker/DeviceWorker.kt
```diff
@@ -132,7 +133,7 @@ internal class DeviceWorker(
         stateWorker.clearPackages(currentState)
     }.map { intendedState }
 
-    private fun executeAction(action: InstrumentationTestRunAction): DeviceTestCaseRun {
+    private suspend fun executeAction(action: InstrumentationTestRunAction): DeviceTestCaseRun {
```
sergioboi, line 136: Why is `suspend` appeared?

PR #909, opened 2021-04-02
MBS-10974 Remove excessive critical events from APK signer plugin

eugene-krivobokov:


subprojects/gradle/signer/src/main/kotlin/com/avito/plugin/SignViaServiceAction.kt
```diff
@@ -74,50 +54,29 @@ internal class SignViaServiceAction(
...
+            val stringBody = response.body
                 ?.string()
                 ?: "Cannot read the response body"
 
-            error("Sign service returned: ${this.code}, body: $stringBody")
+            error("Failed to sign APK via service: code ${response.code}, body: $stringBody")
         }
     }
-
-    private fun <T> retry(
```
eugene-krivobokov, line 100: Reuse an interceptor instead of a custom retry

subprojects/gradle/signer/src/main/kotlin/com/avito/plugin/SignViaServiceAction.kt
```diff
@@ -13,45 +11,27 @@ import okhttp3.Request
...
-        return Result.tryCatch {
-            retry(
-                retriesCount = 6,
-                delaySeconds = 0,
-                attemptFailedHandler = { attempt, throwable ->
-                    logger.critical("Attempt $attempt: failed to sign apk via service", throwable)
-                },
-                actionFailedHandler = { throwable ->
-                    val message = "Failed to sign apk via service"
-                    logger.critical(message, throwable)
```
eugene-krivobokov, line 41: It's unnecessary. Task will fail in case of error.

subprojects/gradle/signer/src/test/kotlin/com/avito/plugin/SignViaServiceActionTest.kt
```diff
@@ -68,22 +63,8 @@ class SignViaServiceActionTest {
...
-            assertThat(throwable.message).contains("Failed to sign apk via service")
+            assertThat(throwable.message).contains("Failed to sign APK via service: code 500")
         }
-
-        assertWithMessage("retry to send")
-            .that(server.requestCount).isAtLeast(5)
-    }
-
-    @Test
-    fun `action success - successful request after failed`() {
```
eugene-krivobokov, line 79: Don't need to test retry interceptor.

subprojects/gradle/signer/src/main/kotlin/com/avito/plugin/SignViaServiceAction.kt
dsvoronin: thanks!, don't know why we needed that in first place,
as far as i remember we had `RetryInterceptor` already when added that retry

dsvoronin (approved):

subprojects/gradle/signer/src/main/kotlin/com/avito/plugin/SignArtifactTask.kt
```diff
@@ -51,6 +52,13 @@ abstract class SignArtifactTask @Inject constructor(objects: ObjectFactory) : De
             .connectTimeout(10L, TimeUnit.SECONDS)
             .writeTimeout(40L, TimeUnit.SECONDS)
             .readTimeout(40L, TimeUnit.SECONDS)
+            .addInterceptor(
+                RetryInterceptor(
+                    retries = 3,
+                    allowedMethods = listOf("GET", "POST"),
```
sergioboi, line 58: should we simplify API in scenarious when we want to retry all requests?

eugene-krivobokov: I found only 4 usages and in all cases, we add POST because of default GET. 
It seems we care only about response codes (5xx) but not about methods.

I propose to retry all methods by default.
@dsvoronin , @sboishtyan , what do you think?

sergioboi: I agree

dsvoronin: RetryInterceptor is general purpose interceptor, anyone can use it now
in REST only GET methods can be cached (or be idempotent),
and retry POST methods is generally a mistake

that's the reason behind defaults

Why remove this configurability? For what reason? What does we gain in exchange?

eugene-krivobokov: > retry POST methods is generally a mistake

Current API doesn't give an ability to have such configurations:

- GET <---> { codes }
- POST <---> { codes }

See how clients use it. 
Clients want to retry 5xx errors.  They configure POST because they use POST methods. 

> Why remove this configurability?

No one is talking about removing configurations. 
The proposal is only to change defaults for `allowedMethods` - include all methods by default.
Because this is how it's used now.

PR #910, opened 2021-04-03
Mbs 10889 Use infra version as a release in Sentry events

eugene-krivobokov:
### The problem

We used a commit hash as a release in Sentry events. 

It's hard to understand the health of release because of many tiny releases:

Screenshot from 2021-04-03 15-41-59

Also, I can't resolve an issue and select the next release. Instead of this, I have to suppress the issue until the next release date. It's error-prone.

Screenshot from 2021-04-03 15-31-31

### Changes

- Store an infra version [inside JARs](https://github.com/avito-tech/avito-android/pull/910#discussion_r606674741)
- Use infra version as release in Sentry events

### The next steps

Do the same for Android libraries and events from instrumentation tests.

### DIFF
--- a/build-logic/publication/src/main/kotlin/convention.publish-base.gradle.kts
+++ b/build-logic/publication/src/main/kotlin/convention.publish-base.gradle.kts
@@ -11,7 +11,7 @@ version = providers.gradleProperty("projectVersion")
     .forUseAtConfigurationTime()
     .get()
 
-tasks.withType(Jar::class.java) {
+tasks.withType<Jar> {
     manifest {
         attributes(
             mapOf(
--- a/build-logic/publication/src/main/kotlin/convention.publish-base.gradle.kts
+++ b/build-logic/publication/src/main/kotlin/convention.publish-base.gradle.kts
@@ -1,3 +1,5 @@
+import java.util.jar.Attributes
+
 plugins {
     `maven-publish`
 }
@@ -8,6 +10,16 @@ group = "com.avito.android"
 version = providers.gradleProperty("projectVersion")
     .forUseAtConfigurationTime()
     .get()
+
+tasks.withType(Jar::class.java) {
+    manifest {
+        attributes(
+            mapOf(
+                Attributes.Name.IMPLEMENTATION_VERSION.toString() to project.version
+            )
+        )
+    }
+}
 
 publishing.publications.withType<MavenPublication> {
     pom {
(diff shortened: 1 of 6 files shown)


build-logic/publication/src/main/kotlin/convention.publish-base.gradle.kts
```diff
@@ -9,6 +11,16 @@ version = providers.gradleProperty("projectVersion")
...
 
+tasks.withType(Jar::class.java) {
+    manifest {
+        attributes(
+            mapOf(
+                Attributes.Name.IMPLEMENTATION_VERSION.toString() to project.version
+            )
+        )
+    }
+}
```
eugene-krivobokov, line 22: Added a release version to all JARs.

Possible use cases:
- Metadata for troubleshooting. E.g. then we see a JAR file as an intermediate artifact in Gradle cache.
- A client wants to use the new implementation in advance and it's backward compatible. In this case, they needs to check in runtime.
- We want to add a version-specific logic in infrastructure. Now we can only add a comment `// TODO: do something after 2021.X release`. We could check it programmatically.

eugene-krivobokov: An alternative option is to pass the infra version as a parameter. But it's error-prone and leads to duplication. Clients need to know it while configuring build scripts.

build-logic/publication/src/main/kotlin/convention.publish-base.gradle.kts
```diff
@@ -9,6 +11,16 @@ version = providers.gradleProperty("projectVersion")
     .forUseAtConfigurationTime()
     .get()
 
+tasks.withType(Jar::class.java) {
```
dsvoronin: you can use reified generic version: `tasks.withType<Jar> {`

build-logic/publication/src/main/kotlin/convention.publish-base.gradle.kts
dsvoronin: i like attribute version and agree with use cases,  nice!

subprojects/common/sentry/src/main/kotlin/com/avito/android/sentry/SentryConfig.kt
```diff
@@ -44,7 +44,7 @@ sealed class SentryConfig : Serializable {
         val dsn: String,
         val environment: String,
         val serverName: String,
-        val release: String,
+        val release: String?,
```
dsvoronin, line 47: Can't find why it became nullable

eugene-krivobokov: In Sentry, the release is optional. I see no reasons to make it required in config.
It forces us to handle two states:
- No release
- "unknown" release

build-logic/publication/src/main/kotlin/convention.publish-base.gradle.kts
eugene-krivobokov: Fixed

dsvoronin (approved):

PR #911, opened 2021-04-03
use Result type for readability

dsvoronin:


subprojects/common/test-okhttp/build.gradle.kts
```diff
@@ -10,12 +10,17 @@ dependencies {
     api(project(":common:okhttp"))
```
eugene-krivobokov, line 10: Should we convert the module into testFixtures in `common:okhttp`?

eugene-krivobokov (approved):

PR #915, opened 2021-04-07
Reduce timeout and add retries

sergioboi:


subprojects/common/report-viewer/src/main/kotlin/com/avito/report/ReportsApiFactory.kt
```diff
@@ -12,7 +12,7 @@ import java.util.concurrent.TimeUnit
 
 object ReportsApiFactory {
 
-    private const val TIMEOUT_SEC = 30L
+    private const val TIMEOUT_SEC = 10L
```
sergioboi, line 15: I checked the dashboard and see that the longest time is 5s
So I think we don't need to wait 30 seconds

subprojects/common/report-viewer/src/main/kotlin/com/avito/report/ReportsApiFactory.kt
```diff
@@ -25,7 +25,7 @@ object ReportsApiFactory {
         host: String,
         httpClientProvider: HttpClientProvider,
         loggerFactory: LoggerFactory,
-        retryRequests: Boolean = false
+        retryRequests: Boolean = true
```
sergioboi, line 28: I realized that we retry never, but I think we should
because current amount of failures because of retries absence

dsvoronin (approved):

eugene-krivobokov (approved):

PR #924, opened 2021-04-11
add special case check for clusters with only default namespace

dsvoronin:


subprojects/gradle/kubernetes/src/main/kotlin/com/avito/utils/gradle/Kubernetes.kt
```diff
@@ -40,11 +40,27 @@ fun createKubernetesClient(
                     caCertFile = kubernetesCredentials.caCertFile
                 }
 
-                // working with multiple namespaces/contexts is not supported
-                require(getNamespace() == namespace) {
-                    "kubernetes.context.namespace should be $namespace, but was ${getNamespace()}. " +
+                val namespaceFromKubeConfig = getNamespace()
```
eugene-krivobokov, line 43: `...FromKubeConfig` - we add the context because we are inside `config.apply {}`
Maybe, without `apply` the code will be a bit more obvious and less nested.

```kotlin
val config = Config.fromKubeconfig()
val namespace = config.getNamespace()
....
```

subprojects/gradle/kubernetes/src/main/kotlin/com/avito/utils/gradle/Kubernetes.kt
```diff
@@ -40,11 +40,27 @@ fun createKubernetesClient(
...
+
+                val namespaceDiffErrorMessage = {
+                    "kubernetes.context.namespace should be $namespace, but was $namespaceFromKubeConfig. " +
                         "Namespace hardcoded in plugin, and this check only prevents from using wrong context"
                 }
+
+                if (namespace == "default") {
+                    // special case for clusters with default namespaces
```
eugene-krivobokov: The comment inside the check is redundant. 
We already made this decision. The comment adds nothing more.

subprojects/gradle/kubernetes/src/main/kotlin/com/avito/utils/gradle/Kubernetes.kt
```diff
@@ -40,11 +40,27 @@ fun createKubernetesClient(
...
+                    require(
+                        value = namespaceFromKubeConfig == "default" || namespaceFromKubeConfig.isNullOrBlank(),
+                        lazyMessage = namespaceDiffErrorMessage
+                    )
+                } else {
+                    // working with multiple namespaces/contexts is not supported
```
eugene-krivobokov: Not clear why we mention multiplicity in the comments. 
I see the check "namespaceFromKubeConfig == namespace". 
It looks self-explanatory: we require this namespace.

eugene-krivobokov (approved):

subprojects/gradle/kubernetes/src/main/kotlin/com/avito/utils/gradle/Kubernetes.kt
dsvoronin: yep, i don't know

subprojects/gradle/kubernetes/src/main/kotlin/com/avito/utils/gradle/Kubernetes.kt
dsvoronin: removed both comments; agreed

PR #930, opened 2021-04-15
MBS-11060 Split test-report on few parts: api, dsl-api, dsl, report

sergioboi:


dsvoronin (approved):
The only thing i wanted to mention is missing `inline` for step methods which will cause generation of much more methods in test apk.

It's not a blocker or something now, when we are testing with minsdk 21

PR #935, opened 2021-04-16
MBS-10989 don't crash if can't get previous tests with enabled filter

dsvoronin:


subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/internal/suite/filter/FilterFactoryImpl.kt
```diff
@@ -80,26 +85,33 @@ internal class FilterFactoryImpl(
...
+                            add(
+                                ExcludeByTestSignaturesFilter(
+                                    source = TestsFilter.Signatures.Source.PreviousRun,
+                                    signatures = previousRunTests.filterBy(previousStatuses.excluded)
+                                )
+                            )
+                        }
+                    },
+                    onFailure = { throwable ->
+                        logger.info("Can't get tests from previous run: ${throwable.message}")
```
sergioboi, line 112: warning

subprojects/gradle/tms/src/main/kotlin/com/avito/plugin/MarkReportAsSourceAction.kt
```diff
@@ -26,10 +25,10 @@ class MarkReportAsSourceAction(
     fun mark(reportCoordinates: ReportCoordinates) {
         val testSuiteVersion = timeProvider.nowInMillis()
 
-        when (val result = reportsApi.getReport(reportCoordinates)) {
-            is GetReportResult.Found ->
+        reportsApi.getReport(reportCoordinates).fold(
```
sergioboi, line 28: I see result API as:
getReport
.flatMap { report -> }
.onError { error -> }

sergioboi: This is out of scope but it will improve readability

sergioboi (approved):

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/internal/suite/filter/FilterFactoryImpl.kt
dsvoronin: we have three cases:
1. ok -> apply filter
2. no report found -> info
3. network problem -> warning

i decided to merge 2 and 3 with info level log, because i don't think we want to react on this somehow, and it is for simplicity mainly

warning now for 2+3 i think is not a good alternative

Do you want me to make it 3 cases result?

subprojects/gradle/tms/src/main/kotlin/com/avito/plugin/MarkReportAsSourceAction.kt
dsvoronin: `Result<T>.flatmap()` is used to return another `Result<R>`, and you want it to return `T` here
fold on the other hand returns T

maybe
```
getReport
.onSuccess { report -> }
.onError { error -> }
```

is what you want? just plain side effects expression without return value in mind

RuslanMingaliev: Btw, why don't we use [this](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-result/) implementation of Result? Seems like it solves the same issues

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/internal/suite/filter/FilterFactoryImpl.kt
sergioboi: info now create only noise, we could simply do nothing comparing to `log.info`

subprojects/gradle/tms/src/main/kotlin/com/avito/plugin/MarkReportAsSourceAction.kt
sergioboi: >Result<T>.flatmap() is used to return another Result<R>

Sorry but I see you make another request on Success for me it is a `flatmap`

subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/internal/suite/filter/FilterFactoryImpl.kt
dsvoronin: i don't think so;
it will help us to understand what's going on in build by looking for this message

now you have filters json, but why there is no PreviousRun? look at Elastic

subprojects/gradle/tms/src/main/kotlin/com/avito/plugin/MarkReportAsSourceAction.kt
dsvoronin: > Btw, why don't we use this implementation of Result? Seems like it solves the same issues

- It's experimental and could be somehow changed in future releases
- it requires to enable special compiler option to go without error
- it sometimes exposes in modules API, which will require clients to enable this compiler options 

sorry not experimental but https://stackoverflow.com/a/52632328/981330

dsvoronin: > Sorry but I see you make another request on Success for me it is a flatmap

ah, now i understand. let's see

dsvoronin: #936

eugene-krivobokov: > why don't we use this implementation of Result? Seems like it solves the same issues
> ...

You discussed the reasons. For others, they will remain in this thread.
Let's fix knowledge in `Result` documentation. 
Also, it will help to understand when to switch to Kotlin's Result.

dsvoronin: #944

PR #936, opened 2021-04-16
Mbs 10989 refactor

dsvoronin:
https://github.com/avito-tech/avito-android/pull/935#discussion_r614952655


subprojects/gradle/tms/src/main/kotlin/com/avito/plugin/MarkReportAsSourceAction.kt
```diff
@@ -25,38 +25,31 @@ class MarkReportAsSourceAction(
     fun mark(reportCoordinates: ReportCoordinates) {
         val testSuiteVersion = timeProvider.nowInMillis()
 
-        reportsApi.getReport(reportCoordinates).fold(
-            onSuccess = { report ->
+        reportsApi.getReport(reportCoordinates)
+            .map { report ->
```
dsvoronin, line 29: need to pass `report` through the chain, that's why it's `map` instead of `flatMap`

sergioboi (approved):

PR #938, opened 2021-04-18
MBS-10988 fix test artifacts access

dsvoronin:

PR #941, opened 2021-04-19
MBS-10988 Few fixes

dsvoronin:
#938 follow up

### DIFF
--- a/subprojects/android-test/test-report/src/main/kotlin/com/avito/android/test/report/screenshot/ScreenshotCapturerImpl.kt
+++ b/subprojects/android-test/test-report/src/main/kotlin/com/avito/android/test/report/screenshot/ScreenshotCapturerImpl.kt
@@ -9,8 +9,7 @@ import com.avito.android.test.report.screenshot.ScreenshotCapturer.Capture
 import com.avito.android.test.util.getCurrentActivityOrNull
 import com.avito.android.util.runOnMainThreadSync
 import com.avito.report.TestArtifactsProvider
-import com.avito.report.TestDirGenerator
-import com.avito.report.internal.UniqueDirTestArtifactsProvider
+import com.avito.report.TestArtifactsProviderFactory
 import java.io.File
 import java.io.FileOutputStream
 
@@ -20,10 +19,7 @@ class ScreenshotCapturerImpl(private val testArtifactsProvider: TestArtifactsPro
     // todo remove in avito and then here
     @Suppress("unused")
     constructor(outputDirectory: Lazy<File>) : this(
-        testArtifactsProvider = UniqueDirTestArtifactsProvider(
-            rootDir = outputDirectory,
-            testDirGenerator = TestDirGenerator.Stub
-        )
+        testArtifactsProvider = TestArtifactsProviderFactory.createStub(outputDirectory)
     )
 
     override fun captureBitmap(): Result<Capture> {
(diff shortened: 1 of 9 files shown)


subprojects/common/files/src/main/kotlin/com/avito/utils/Path.kt
```diff
@@ -1,12 +1,13 @@
 package com.avito.utils
 
+import androidx.annotation.RequiresApi
 import com.avito.android.Result
 import java.nio.file.Path
 
 /**
  * WARNING: only android 26+ supported
  */
```
eugene-krivobokov, line 9: Now we can remove the comment

eugene-krivobokov (approved):

subprojects/android-test/test-report/src/main/kotlin/com/avito/android/test/report/screenshot/ScreenshotCapturerImpl.kt
```diff
@@ -20,10 +19,7 @@ class ScreenshotCapturerImpl(private val testArtifactsProvider: TestArtifactsPro
     // todo remove in avito and then here
     @Suppress("unused")
     constructor(outputDirectory: Lazy<File>) : this(
-        testArtifactsProvider = UniqueDirTestArtifactsProvider(
-            rootDir = outputDirectory,
-            testDirGenerator = TestDirGenerator.Stub
-        )
+        testArtifactsProvider = TestArtifactsProviderFactory.createStub(outputDirectory)
```
sergioboi, line 22: I could misunderstand the `createStub` name in the future

sergioboi: Because it's a production code and I don't know what kind of Stub we create

subprojects/common/files/src/main/kotlin/com/avito/utils/Path.kt
```diff
@@ -1,12 +1,13 @@
...
-@Suppress("NewApi")
+@RequiresApi(api = 26)
 fun Path.deleteRecursively(): Result<Unit> {
```
sergioboi, line 11: So what will I use at API lower than 26?

eugene-krivobokov: We don't provide a replacement because we don't have such cases.
The point was only to show the restriction explicitly.

subprojects/android-test/test-report/src/main/kotlin/com/avito/android/test/report/screenshot/ScreenshotCapturerImpl.kt
dsvoronin: It's a temporary  solution for backward compatibility
Take a look at: https://github.com/avito-tech/avito-android/blob/854f5a03b6c9fede37fc41d85ef4f246f79a8408/subprojects/android-test/test-report/src/main/kotlin/com/avito/android/test/report/screenshot/ScreenshotCapturerImpl.kt#L22

It's only usage and will be removed in next release

subprojects/common/files/src/main/kotlin/com/avito/utils/Path.kt
dsvoronin: nothing, this method is a helper for gradle runtime, but nothing will stop you on using it from android because `common:files` comes as transitive dependecy

This check could help you avoid runtime exception

PR #943, opened 2021-04-20
MBS-11034 Prevent accidental releases from develop

eugene-krivobokov:


dsvoronin (approved):

dsvoronin:
In future we can be more specific and introduce `release/*` branches to also cover another cases

PR #947, opened 2021-04-20
MBS-11010 Hide registering unexpected incident logic to report impl

sergioboi:
client must be clear from report state logic
Changed sematics - now unexpected incident after report was written event is a warning not a critical


subprojects/android-test/test-inhouse-runner/src/main/kotlin/com/avito/android/runner/InHouseInstrumentationTestRunner.kt
```diff
@@ -276,14 +275,7 @@ abstract class InHouseInstrumentationTestRunner :
     }
 
     fun tryToReportUnexpectedIncident(incident: Throwable) {
```
eugene-krivobokov: `tryTo` prefix is not needed now. 
We can't know about implementation here.

eugene-krivobokov (approved):

dsvoronin (approved):

PR #948, opened 2021-04-20
MBS-11042 Preserve original error for not found test report for troubleshooting

eugene-krivobokov:


subprojects/common/report-viewer/src/main/kotlin/com/avito/report/ReportsFetchApiImpl.kt
```diff
@@ -110,7 +118,7 @@ internal class ReportsFetchApiImpl(
             { exception ->
                 val isNotFoundError = exception.message?.contains("\"data\":\"not found\"") ?: false
```
eugene-krivobokov, line 119: We are losing the original error which could help in troubleshooting

sergioboi (approved):

PR #949, opened 2021-04-21
Screenshoter cleanup

dsvoronin:
- remove legacy constructor
- make ScreenshotCapturerImpl internal
- create :android-test:instrumentation
- create ActivityProvider
- remove test-report dependency on ui-testing-core


subprojects/android-test/test-report/src/main/kotlin/com/avito/android/test/report/screenshot/ScreenshotCapturerFactory.kt
```diff
@@ -0,0 +1,17 @@
+package com.avito.android.test.report.screenshot
+
+import com.avito.android.instrumentation.ActivityProvider
+import com.avito.report.TestArtifactsProvider
+
+object ScreenshotCapturerFactory {
+
+    fun create(
```
eugene-krivobokov, line 8: Factory can hide arguments\dependencies from clients

subprojects/android-test/ui-testing-core/src/main/kotlin/com/avito/android/test/util/Instrumentations.kt
```diff
@@ -1,30 +1,14 @@
...
-import androidx.test.runner.lifecycle.Stage
+import com.avito.android.instrumentation.ActivityProviderFactory
+
+private val activityProvider = ActivityProviderFactory.create()
 
 fun getCurrentActivity(): Activity {
     return getCurrentActivityOrNull() ?: throw IllegalStateException("Resumed activity not found")
 }
 
 fun getCurrentActivityOrNull(): Activity? {
```
eugene-krivobokov: Can we remove this method? 
It doesn't hide much. `ActivityProvider` is already hidden by an interface.

subprojects/android-test/test-report/src/main/kotlin/com/avito/android/test/report/screenshot/ScreenshotCapturerFactory.kt
dsvoronin: added default values

subprojects/android-test/ui-testing-core/src/main/kotlin/com/avito/android/test/util/Instrumentations.kt
dsvoronin: removed

eugene-krivobokov (approved):

subprojects/android-test/instrumentation/src/main/kotlin/com/avito/android/instrumentation/ActivityProviderImpl.kt
```diff
@@ -0,0 +1,54 @@
...
+) : ActivityProvider {
+
+    override fun getCurrentActivity(): Result<Activity> {
+        var result: Result<Activity> = notFoundResult()
+
+        if (isMainThread()) {
+            result = findResumedActivity()
+        } else {
+            try {
+                instrumentation.value.runOnMainSync {
```
sergioboi, line 23: This function could lead to dead lock so it will be better if we add a timeout and if we don't get an Activity in 10 second then fail or return null

PR #951, opened 2021-04-22
Mbs 6015 Update build scan

eugene-krivobokov:


subprojects/gradle/build-metrics/src/main/kotlin/com/avito/android/plugin/build_metrics/SentryConsumer.kt
```diff
@@ -1,97 +0,0 @@
...
-import com.avito.android.sentry.sentry
-import com.avito.kotlin.dsl.getOptionalStringProperty
-import com.avito.logger.Logger
-import com.avito.logger.LoggerFactory
-import com.avito.logger.create
-import io.sentry.connection.EventSendCallback
-import io.sentry.event.Event
-import org.gradle.BuildResult
-import org.gradle.api.Project
-
```
eugene-krivobokov, line 16: Not used already and found a better way.

subprojects/settings.gradle.kts
```diff
@@ -299,7 +299,7 @@ dependencyResolutionManagement {
 }
 
 plugins {
-    id("com.gradle.enterprise") version "3.5.1"
+    id("com.gradle.enterprise") version "3.6.1"
```
eugene-krivobokov, line 303: - Reduced memory usage for when capturing task file inputs in very large builds
- Avoid occasional build scan capture errors during IDE sync
- The `GRADLE_ENTERPRISE_ACCESS_KEY` environment variable allows specifying the access keys for multiple hosts
- Reduced memory usage for builds with lots of input files

dsvoronin (approved):

PR #952, opened 2021-04-23
Mbs 6580 Add optional RV runId prefix to instrumentation configuration  

RuslanMingaliev:


subprojects/common/report-viewer/src/main/kotlin/com/avito/report/model/RunId.kt
```diff
@@ -3,22 +3,43 @@ package com.avito.report.model
...
- * - build config teamcity домешан, чтобы отличать прогон на одном API на develop
- *  (там так же как в релизе нет targetBranch) от buildRelease
+ * Universal identifier for test suits
+ * @param prefix optional prefix
+ * @param commitHash unique SHA identifier of commit
+ * @param buildTypeId is used to distinguish runs with the same API on `develop` branch
  */
-data class RunId(val commitHash: String, val buildTypeId: String) {
+data class RunId(
+    val prefix: String = "",
```
sergioboi: we don't need a default value

subprojects/common/report-viewer/src/main/kotlin/com/avito/report/model/RunId.kt
```diff
@@ -3,22 +3,43 @@ package com.avito.report.model
...
+
+    override fun toString(): String = when {
+        prefix.isBlank() -> "$commitHash$DELIMITER$buildTypeId"
+        else -> "$prefix$DELIMITER$commitHash$DELIMITER$buildTypeId"
+    }
 
     companion object {
 
-        private const val DELIMITER = '.'
+        const val DELIMITER = ':'
```
sergioboi: internal?

subprojects/common/report-viewer/src/main/kotlin/com/avito/report/model/RunId.kt
```diff
@@ -3,22 +3,43 @@ package com.avito.report.model
...
 
         fun fromString(runId: String): Result<RunId> {
```
sergioboi: Where do we use that fun?

subprojects/common/report-viewer/src/main/kotlin/com/avito/report/model/RunId.kt
```diff
@@ -3,22 +3,43 @@ package com.avito.report.model
...
+    }
+
+    override fun toString(): String = when {
+        prefix.isBlank() -> "$commitHash$DELIMITER$buildTypeId"
+        else -> "$prefix$DELIMITER$commitHash$DELIMITER$buildTypeId"
+    }
 
     companion object {
 
-        private const val DELIMITER = '.'
```
sergioboi, line 16: Why do we need that change?

sergioboi: I scare it may break compatibility with old reports

subprojects/gradle/instrumentation-tests/src/gradleTest/kotlin/com/avito/instrumentation/InstrumentationParamsBuildingTest.kt
```diff
@@ -34,6 +34,7 @@ internal class InstrumentationParamsBuildingTest {
             "buildBranch" to "develop",
             "buildCommit" to commit,
             "runId" to RunId(
+                prefix = "stub",
```
sergioboi, line 37: Do we have a test where the prefix is empty?

subprojects/common/report-viewer/src/main/kotlin/com/avito/report/model/RunId.kt
```diff
@@ -3,22 +3,43 @@ package com.avito.report.model
...
-data class RunId(val commitHash: String, val buildTypeId: String) {
+data class RunId(
+    val prefix: String = "",
+    val commitHash: String,
+    val buildTypeId: String
+) {
 
-    override fun toString(): String = "$commitHash$DELIMITER$buildTypeId"
+    init {
+        require(!prefix.contains(DELIMITER)) { "prefix could not contain '$DELIMITER' character" }
```
eugene-krivobokov:
```suggestion
        require(!prefix.contains(DELIMITER)) { "prefix should not contain '$DELIMITER' character" }
```
Because it's not a suggestion

subprojects/common/report-viewer/src/main/kotlin/com/avito/report/model/RunId.kt
RuslanMingaliev: Why? With default value, we are able to create a new instance with both `RunId(hash, type)` and `RunId(prefix, hash, type)`. It's convenient, considering that `prefix` is optional.

subprojects/common/report-viewer/src/main/kotlin/com/avito/report/model/RunId.kt
RuslanMingaliev: Doesn't matter now. There were difficulties with parsing RunId from string, but it's used nowhere. Brought back the old value.

subprojects/common/report-viewer/src/main/kotlin/com/avito/report/model/RunId.kt
```diff
@@ -1,32 +1,23 @@
...
  */
-data class RunId(val commitHash: String, val buildTypeId: String) {
+data class RunId(
+    private val prefix: String? = null,
+    private val commitHash: String,
+    private val buildTypeId: String
+) {
 
-    override fun toString(): String = "$commitHash$DELIMITER$buildTypeId"
+    fun value(): String = when {
```
sergioboi: `toReportViewerFormat()`

sergioboi (approved):

PR #962, opened 2021-04-27
release hotfix

dsvoronin:
don't fail on unavailable report


subprojects/gradle/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/internal/finalizer/LegacyFinalizer.kt
```diff
@@ -22,43 +22,38 @@ internal class LegacyFinalizer(
 
     override fun finalize(testsExecutionResults: TestsScheduler.Result) {
 
-        val reportedTestsResult = report.getTests(testsExecutionResults.testSuite.testsToRun.map { it.test.name })
+        val testResults = report.getTests(
+            initialSuiteFilter = testsExecutionResults.testSuite.testsToRun.map { it.test.name }
+        ).getOrElse { emptyList() }
```
sergioboi, line 27: But know we will green when we have failed tests?
Or we will fail when read inmemory report

dsvoronin: it is the same logic as it was before; we will fail on LostTestDeterminer in that case

sergioboi (approved):

PR #963, opened 2021-04-28
MBS-10989 extract `:common:report-api` module

dsvoronin:
a bunch of changes around generic report models and classes to separate it from avito-specific


subprojects/common/report-api/src/main/kotlin/com/avito/report/ReportLinkGenerator.kt
```diff
@@ -0,0 +1,26 @@
...
+public interface ReportLinkGenerator {
+
+    public fun generateReportLink(
+        filterOnlyFailtures: Boolean = false,
+        team: String? = null
+    ): String
+
+    public fun generateTestLink(testName: TestName): String
+
+    public class NoOp(
```
sergioboi: impl

subprojects/common/report-api/src/main/kotlin/com/avito/report/TestSuiteNameProvider.kt
```diff
@@ -0,0 +1,11 @@
+package com.avito.report
+
+public interface TestSuiteNameProvider {
+
+    public fun getName(): String
+
+    public class NoOp(private val name: String = "Unknown") : TestSuiteNameProvider {
```
sergioboi: impl

subprojects/common/report-api/build.gradle.kts
```diff
@@ -0,0 +1,14 @@
+plugins {
```
sergioboi, line 1: this module looks non-independent
In module `report-api` I expect to find some interface `Report` with contract
And anyone could create an impl module for that `Report`

What I actually see: bunch of models I don't know how to use

So now module looks like a `report-models` module

subprojects/common/report-viewer/src/main/kotlin/com/avito/report/TestArtifactsProvider.kt
```diff
@@ -14,17 +14,17 @@ import java.io.File
  *
  * use [TestArtifactsProviderFactory] to create instances
  */
-interface TestArtifactsProvider {
+public interface TestArtifactsProvider {
```
sergioboi, line 17: As I see this interface doesn't relate to `report-viwer`

dsvoronin: yes, seems it should be another module

sergioboi (approved):

subprojects/common/report-api/build.gradle.kts
dsvoronin: Sure, just not in single PR;
There are a lot of weird dependencies i am trying to unravel right now )

subprojects/common/report-api/src/main/kotlin/com/avito/report/ReportLinkGenerator.kt
dsvoronin: extracted to separate class with more meaningful name

subprojects/common/report-viewer/src/main/kotlin/com/avito/report/TestArtifactsProvider.kt
dsvoronin: done

PR #965, opened 2021-04-29
MBS-10989 revert incidentElement model change

dsvoronin:
tests in avito were broken by it

### DIFF
--- a/subprojects/common/report-api/src/main/kotlin/com/avito/report/model/IncidentElement.kt
+++ b/subprojects/common/report-api/src/main/kotlin/com/avito/report/model/IncidentElement.kt
@@ -1,5 +1,7 @@
 package com.avito.report.model
 
+import com.google.gson.JsonElement
+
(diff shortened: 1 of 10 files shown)


subprojects/common/report-api/src/main/kotlin/com/avito/report/model/IncidentElement.kt
```diff
@@ -13,5 +15,5 @@ public data class IncidentElement(
     val type: String? = null,
     val origin: String? = null,
     val className: String? = null,
-    val data: String? = null
+    val data: JsonElement? = null
 )
```
eugene-krivobokov, line 19: Let's mark `SerializedName`. RM client parses this model from json.

And what about `com.avito.report.internal.model.IncidentElement` from report-viewer? Are they the same?

eugene-krivobokov (approved):

PR #966, opened 2021-04-29
Mbs 6015 Log unknown Gradle cache errors

eugene-krivobokov:


subprojects/gradle/build-metrics/src/main/kotlin/com/avito/android/plugin/build_metrics/internal/cache/BuildCacheOperationListener.kt
```diff
@@ -147,6 +152,7 @@ internal class BuildCacheOperationListener(
             message.substringAfter(" response status ").substringBefore(':').trim()
                 .toInt()
         } else {
+            logger.warn("Unknown cache store error", error)
```
eugene-krivobokov, line 155: The only change in behavior is here

dsvoronin (approved):

subprojects/gradle/build-metrics/src/main/kotlin/com/avito/android/plugin/build_metrics/internal/cache/BuildCacheOperationListener.kt
sergioboi, line 155: Will we see those warnings at Grafana? Or when and how I need to react on them?

eugene-krivobokov: These events are already in grafana as "unknown" errors.
I want to collect them to Sentry for further troubleshooting.

PR #967, opened 2021-04-29
MBS-11063 Use the original Lint issue severity model

eugene-krivobokov:
We parse an Android lint xml report to find errors and decide what to do.
The model of a report is our own and differs from the original.
I've used the original model from lint API.

### DIFF
--- a/build-logic/kotlin-convention/src/main/kotlin/convention.kotlin-base.gradle.kts
+++ b/build-logic/kotlin-convention/src/main/kotlin/convention.kotlin-base.gradle.kts
@@ -25,3 +25,18 @@ tasks.withType<KotlinCompile>().configureEach {
         freeCompilerArgs = freeCompilerArgs + "-Xopt-in=kotlin.RequiresOptIn"
     }
 }
+
+val kotlinVersion: String = checkNotNull(
+    System.getProperty("kotlinVersion")
+)
+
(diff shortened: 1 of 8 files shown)


build-logic/kotlin-convention/src/main/kotlin/convention.kotlin-base.gradle.kts
```diff
@@ -25,3 +25,18 @@ tasks.withType<KotlinCompile>().configureEach {
...
+
+val kotlinVersion: String = checkNotNull(
+    System.getProperty("kotlinVersion")
+)
+
+configurations.all {
+    resolutionStrategy.eachDependency {
+        if (requested.group == "org.jetbrains.kotlin"
+            && requested.name != "kotlin-stdlib-jre8" // deprecated and not updated anymore. It's replaced by jdk
+            && requested.name != "kotlin-stdlib-jre7"
```
eugene-krivobokov, line 37: Example of usage:

```
:gradle:runner:client > project :gradle:runner:service > com.android.tools.ddms:ddmlib:26.2.0 > com.android.tools:common:26.2.0 > org.jetbrains.kotlin:kotlin-stdlib-jre8:1.2.0
```

It'd be good to udpate.

dsvoronin (approved):

PR #969, opened 2021-04-30
add metrics, retries and increase Slack request timeouts

dsvoronin:
added because of flaky integration tests

### DIFF
--- a/subprojects/common/http-client/src/main/kotlin/com/avito/http/HttpClientProvider.kt
+++ b/subprojects/common/http-client/src/main/kotlin/com/avito/http/HttpClientProvider.kt
@@ -1,7 +1,9 @@
 package com.avito.http
 
 import com.avito.android.stats.StatsDSender
+import com.avito.http.internal.RequestMetadataProvider
 import com.avito.http.internal.StatsHttpEventListener
+import com.avito.http.internal.TagRequestMetadataProvider
 import com.avito.logger.LoggerFactory
 import com.avito.time.TimeProvider
 import okhttp3.OkHttpClient
@@ -9,16 +11,19 @@ import okhttp3.OkHttpClient
 public class HttpClientProvider(
     private val statsDSender: StatsDSender,
     private val timeProvider: TimeProvider,
-    private val loggerFactory: LoggerFactory
+    private val loggerFactory: LoggerFactory,
 ) {
 
     private val builder = OkHttpClient.Builder()
 
-    public fun provide(): OkHttpClient.Builder {
+    public fun provide(
+        requestMetadataProvider: RequestMetadataProvider = TagRequestMetadataProvider()
+    ): OkHttpClient.Builder {
         return builder.eventListenerFactory {
             StatsHttpEventListener(
                 statsDSender = statsDSender,
                 timeProvider = timeProvider,
+                requestMetadataProvider = requestMetadataProvider,
                 loggerFactory = loggerFactory
             )
         }
--- a/subprojects/common/http-client/src/main/kotlin/com/avito/http/internal/StatsHttpEventListener.kt
+++ b/subprojects/common/http-client/src/main/kotlin/com/avito/http/internal/StatsHttpEventListener.kt
@@ -16,7 +16,8 @@ import java.net.SocketTimeoutException
 internal class StatsHttpEventListener(
     private val statsDSender: StatsDSender,
     private val timeProvider: TimeProvider,
-    loggerFactory: LoggerFactory
+    private val requestMetadataProvider: RequestMetadataProvider,
+    loggerFactory: LoggerFactory,
 ) : EventListener() {
 
     private val logger = loggerFactory.create<StatsHttpEventListener>()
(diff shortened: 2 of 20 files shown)


subprojects/common/http-client/src/main/kotlin/com/avito/http/internal/StatsHttpEventListener.kt
```diff
@@ -47,30 +48,36 @@ internal class StatsHttpEventListener(
...
-                "unknown-method"
-            )
-        } else {
-            SeriesName.create(
-                "service",
-                metadata.serviceName,
-                metadata.methodName
-            )
-        }
+                logger.warn(
```
eugene-krivobokov, line 70: How about fail-fast? Is it possible?

subprojects/gradle/lint-report/build.gradle.kts
```diff
@@ -14,6 +14,7 @@ dependencies {
     implementation(libs.okhttp)
     implementation(libs.androidToolsLintApi)
 
+    implementation(project(":common:http-client"))
     implementation(project(":common:okhttp"))
```
eugene-krivobokov, line 18: `http-client` and `okhttp` are on different layers of abstraction.
Could I use only http-client?

subprojects/common/http-client/src/main/kotlin/com/avito/http/internal/StatsHttpEventListener.kt
dsvoronin: not really 
https://square.github.io/okhttp/4.x/okhttp/okhttp3/-event-listener/

All event methods must execute fast, without external locking, **cannot throw exceptions**, attempt to mutate the event parameters, or be re-entrant back into the client. Any IO - writing to files or network should be done asynchronously.

eugene-krivobokov: What will happen if we throw an exception? Will it be swallowed or something else?

eugene-krivobokov (approved):

subprojects/gradle/lint-report/build.gradle.kts
dsvoronin: it server different purposes:
`:common:http-client` only provides pre-baked httpClientBuilders
`:common:okhttp` contains extensions

if you need both - you include both

subprojects/common/http-client/src/main/kotlin/com/avito/http/internal/StatsHttpEventListener.kt
dsvoronin: i don't know
at least it will break contract of upstream functions

subprojects/common/http-client/src/main/kotlin/com/avito/http/internal/StatsHttpEventListener.kt
```diff
@@ -47,30 +48,36 @@ internal class StatsHttpEventListener(
...
+                val urlWithoutParams = request.url.newBuilder()
+                    .query(null)
+                    .build()
 
-        return if (metadata == null) {
-            val urlWithoutParams = request.url.newBuilder()
-                .query(null)
-                .build()
+                val seriesName = SeriesName.create(
+                    "service",
```
sergioboi, line 65: Did you create a view on dashboard to notice the amount of that problems?

subprojects/common/http-client/src/main/kotlin/com/avito/http/HttpClientProvider.kt
```diff
@@ -1,24 +1,29 @@
...
     private val timeProvider: TimeProvider,
-    private val loggerFactory: LoggerFactory
+    private val loggerFactory: LoggerFactory,
 ) {
 
     private val builder = OkHttpClient.Builder()
 
-    public fun provide(): OkHttpClient.Builder {
+    public fun provide(
+        requestMetadataProvider: RequestMetadataProvider = TagRequestMetadataProvider()
```
sergioboi, line 20: I see the problem: we create an `OkHttpClient.Builder` with `RequestMetadataProvider` so it adds an implicit requirement that the request must contain a tag; on the opposite side we describe tag for requests in retrofit so I'm easily could forget to add a tag to requests or create Retrofit service with incorrect `OkHttpClient`

I propose unite Retrofit Services creation with adding `RequestMetadataProvider` for `OkHttpClient` used by those Retrofit services

subprojects/gradle/lint-report/build.gradle.kts
sergioboi: It's not clear what those modules contain
maybe we should rename okhttp to okhttp-extensions

subprojects/common/http-client/src/main/kotlin/com/avito/http/internal/StatsHttpEventListener.kt
dsvoronin: yes, that was the first one i created; look on top of dashboard

subprojects/common/http-client/src/main/kotlin/com/avito/http/HttpClientProvider.kt
dsvoronin: i don't quite understand your proposal; let's discuss it on next sync

p.s.
this module doesn't know about retrofit and also used with services which use okhttp directly (without retrofit)

subprojects/gradle/lint-report/build.gradle.kts
dsvoronin: sure; but there is a problem with such renaming, because it's used in avito
we can change module name, and not change publication artifact name; but that will add even more confusion :(

subprojects/common/http-client/src/main/kotlin/com/avito/http/HttpClientProvider.kt
sergioboi: Ok

subprojects/gradle/lint-report/build.gradle.kts
sergioboi: We could make breaking change with naming

PR #971, opened 2021-04-30
fix latency metric

dsvoronin:
was: call start -> response headers end
now: call start -> call end (with response body download)

### DIFF
--- a/subprojects/common/http-client/src/main/kotlin/com/avito/http/internal/StatsHttpEventListener.kt
+++ b/subprojects/common/http-client/src/main/kotlin/com/avito/http/internal/StatsHttpEventListener.kt
@@ -24,15 +24,29 @@ internal class StatsHttpEventListener(
 
     var callStarted = 0L
 
+    var responseCode: Int? = null
+
     override fun callStart(call: Call) {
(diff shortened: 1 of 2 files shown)


sergioboi (approved):

subprojects/common/http-client/src/main/kotlin/com/avito/http/internal/StatsHttpEventListener.kt
```diff
@@ -24,15 +24,29 @@ internal class StatsHttpEventListener(
...
+        }
+    }
 
-        val latencyMs = response.receivedResponseAtMillis - response.sentRequestAtMillis
-        statsDSender.send(TimeMetric(prefix.append("${response.code}"), latencyMs))
+    override fun callEnd(call: Call) {
+        if (responseCode in 200..299) {
+            val latencyMs = timeProvider.nowInMillis() - callStarted
+            sendCode(call, responseCode.toString(), latencyMs)
+        }
```
eugene-krivobokov, line 49: What for other codes? 
Will `callFailed()` be invoked for all of them?

PR #973, opened 2021-05-05
Mbs 8372 Critical path of a build

eugene-krivobokov:
Added a critical path.
See `BuildTrace.md`. All other comments are about implementation details.


docs/content/projects/BuildTrace.md
```diff
@@ -9,25 +9,83 @@ This plugin collects tasks execution time in a trace event format.
```
eugene-krivobokov, line 9: Here are changes from the user perspective

subprojects/common/graph/build.gradle.kts
```diff
@@ -0,0 +1,9 @@
+plugins {
+    id("convention.kotlin-jvm")
+    id("convention.publish-kotlin-library")
+    id("convention.libraries")
+}
+
+dependencies {
+    implementation(libs.jdgraphtCore)
+}
```
eugene-krivobokov, line 9: New module for graph algorithms.
For now, use it only as a glue for [JGraphT](https://jgrapht.org). I found it convenient and robust.

subprojects/common/graph/src/main/kotlin/com/avito/graph/ShortestPath.kt
```diff
@@ -0,0 +1,174 @@
...
+     *   ↑                                         │
+     *   └-----(2)-- F <--(2)------------- G <-----┘
+     * ```
+     *
+     * - A node weight become an edge weight
+     * - Single source and sink are needed to find the shortest path
+     *     without checking all combinations of roots and leaves
+     * - Dependencies in graph are kinda inverted for faster computing in case we need to compare multiple paths.
+     *   It's more efficient than recomputing paths for different sources.
+     */
```
eugene-krivobokov, line 57: The only non-obvious part in finding the longest\shortest path.

subprojects/common/graph/src/test/kotlin/com/avito/grapth/ShortestPathTest.kt
```diff
@@ -0,0 +1,66 @@
+package com.avito.grapth
+
+import com.avito.graph.Operation
+import com.avito.graph.ShortestPath
+import com.avito.graph.SimpleOperation
+import com.google.common.truth.Truth.assertThat
+import org.junit.jupiter.api.Test
+
+class ShortestPathTest {
+
```
eugene-krivobokov, line 10: Test for algorithm, just in case

subprojects/gradle/build-trace/src/gradleTest/kotlin/com/avito/android/build_trace/CriticalPathTest.kt
```diff
@@ -0,0 +1,280 @@
...
+import com.google.common.truth.Truth.assertThat
+import org.gradle.testkit.runner.TaskOutcome
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Disabled
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.io.TempDir
+import java.io.File
+
+class CriticalPathTest {
+
```
eugene-krivobokov, line 15: Integration tests. 

They are needed mostly to check the correct conversion executed tasks for the algorithm.

subprojects/gradle/build-trace/src/main/kotlin/com/avito/android/build_trace/internal/TraceEventProvider.kt
```diff
@@ -0,0 +1,130 @@
...
+import com.avito.android.trace.CompleteEvent
+import com.avito.android.trace.InstantEvent
+import com.avito.android.trace.TraceEvent
+import org.gradle.api.Project
+import org.gradle.api.Task
+import org.gradle.api.tasks.TaskState
+import java.util.concurrent.TimeUnit
+
+internal class TraceEventProvider {
+
```
eugene-krivobokov, line 15: Here mostly the old logic, moved to another class

docs/content/projects/BuildTrace.md
```diff
@@ -9,25 +9,83 @@ This plugin collects tasks execution time in a trace event format.
...
 
-```groovy
+```kotlin
 plugins {
     id("com.avito.android.build-trace")
 }
+
+buildTrace {
+    enabled.set(true)
+}
```
eugene-krivobokov, line 21: Added a specific extension

docs/content/projects/BuildTrace.md
```diff
@@ -9,25 +9,83 @@ This plugin collects tasks execution time in a trace event format.
...
+[
+    {
+        "path": ":app:assembleAndroidTest",
+        "type": "org.gradle.api.Task",
+        "start": 1620217409525,
+        "finish": 1620217409525,
+        "predecessors": [":avito:assembleDebugAndroidTest"]
+    },
+    ...
+```
```
eugene-krivobokov, line 89: Added a new report for the critical path.

docs/content/projects/BuildTrace.md
```diff
@@ -9,25 +9,83 @@ This plugin collects tasks execution time in a trace event format.
...
+
+Tasks on this path are highlighted in a trace.  
+You can find them by query:
+
+```sql
+SELECT slice.name AS TASK_PATH, slice.ts / 1000000 AS START_MS, slice.dur / 1000000 AS DURATION_MS
+FROM slice JOIN args ON slice.arg_set_id = args.arg_set_id
+WHERE args.flat_KEY = "args.CRITICAL_PATH"
+ORDER BY slice.ts ASC
+```
```
eugene-krivobokov, line 72: Added a critical path to trace events.

subprojects/common/graph/src/main/kotlin/com/avito/graph/ShortestPath.kt
```diff
@@ -0,0 +1,174 @@
...
+
+        checkGraphStructure(graph)
+
+        return graph
+    }
+
+    private fun Graph<Operation, DefaultWeightedEdge>.shortestPath(): GraphPath<Operation, DefaultWeightedEdge> {
+        // We could use negative weights to find the shortest path as solution for the longest path in DAG
+        // Bellman–Ford algorithm supports negative weights but Dijkstra doesn't.
+        // TODO: try topological sorting, it will be more efficient - O(V+E) instead of O(V*E)
```
eugene-krivobokov, line 74: For now, it's not critical, less than a second on the avito app.

subprojects/gradle/build-trace/src/main/kotlin/com/avito/android/build_trace/internal/BuildTraceListener.kt
```diff
@@ -0,0 +1,57 @@
...
+        output.parentFile.mkdirs()
+
+        val report = TraceReport(
+            traceEvents = enrichCriticalPath(events)
+        )
+        TraceReportClient().writeTo(output, report)
+        logger.info("Build trace: ${output.path}")
+    }
+
+    private fun enrichCriticalPath(traceEvents: List<TraceEvent>): List<TraceEvent> {
```
eugene-krivobokov, line 49: This is an old listener. The only new part is this function.

subprojects/gradle/build-trace/src/main/kotlin/com/avito/android/build_trace/internal/TaskExt.kt
```diff
@@ -0,0 +1,29 @@
...
+import org.gradle.api.Task
+import org.gradle.api.internal.TaskInternal
+import org.gradle.api.tasks.TaskDependency
+
+internal val Task.type: Class<out Task>
+    get() = (this as TaskInternal).taskIdentity.type
+
+internal val Task.predecessors: Set<Task>
+    get() {
+        val dependenciesByInputs = project.gradle.taskGraph.getDependencies(this)
```
eugene-krivobokov, line 12: Important knowledge about task dependencies.

This is what we see in a build scan. Didn't find a better way to get them.

subprojects/gradle/build-trace/src/main/kotlin/com/avito/android/build_trace/internal/TaskResult.kt
```diff
@@ -0,0 +1,35 @@
+package com.avito.android.build_trace.internal
+
+@Suppress("ClassName")
+internal sealed class TaskResult {
```
eugene-krivobokov, line 4: Old class, just moved

subprojects/gradle/build-trace/src/main/kotlin/com/avito/android/build_trace/internal/critical_path/CriticalPathListener.kt
```diff
@@ -0,0 +1,47 @@
...
+import com.avito.android.gradle.profile.BuildProfile
+import com.avito.android.gradle.profile.TaskExecution
+import com.avito.graph.ShortestPath
+import org.gradle.BuildResult
+import org.gradle.api.Task
+
+internal class CriticalPathListener(
+    private val serialization: CriticalPathSerialization
+) : AbstractBuildEventsListener(), CriticalPathProvider {
+
```
eugene-krivobokov, line 13: This is a provider for the critical path. It converts executed tasks to the algorithm.

subprojects/gradle/build-trace/src/main/kotlin/com/avito/android/build_trace/internal/critical_path/CriticalPathListener.kt
```diff
@@ -0,0 +1,47 @@
...
+            .map { it.invertTime() }
+    }
+
+    override fun path(): List<TaskOperation> {
+        return criticalPath
+    }
+
+    // Using a TaskExecutionListener instead of OperationCompletionListener
+    //   due to https://github.com/gradle/gradle/issues/15824
+    override fun afterExecute(task: Task, state: TaskExecution) {
```
eugene-krivobokov, line 28: This implementation will be a problem in two cases:

- To share critical path between multiple plugins
- For configuration cache

I'll find a solution in another issue\PR.

docs/content/projects/BuildTrace.md
```diff
@@ -9,25 +9,83 @@ This plugin collects tasks execution time in a trace event format.
...
 --8<--
 plugins-setup.md
 --8<--
 
-Run a build with `--profile` argument. You will get a message in a log:
+Run a build. You will get a message in a log:
 
 ```log
-Build trace: <path to the project>/outputs/trace/build.trace
+Build trace: <path to the project>/outputs/avito/build-trace/build.trace
```
dsvoronin:
```suggestion
Build trace: <path to the project>/outputs/<project name>/build-trace/build.trace
```

docs/content/projects/BuildTrace.md
```diff
@@ -9,25 +9,83 @@ This plugin collects tasks execution time in a trace event format.
...
+Here you can also make [analytical queries by SQL](https://perfetto.dev/docs/analysis/trace-processor):
+
+```sql
+-- Slowest tasks
+SELECT slice.name AS TASK_PATH, slice.dur / 1000000 AS DURATION_MS
+FROM slice
+ORDER BY slice.dur DESC
+```
+
+## Critical path
```
dsvoronin, line 59: An attached image would be helpful imo

dsvoronin (approved):

docs/content/projects/BuildTrace.md
eugene-krivobokov: The plugin works in a root project. It's not a project name but a company name.
This is to mitigate probable clashes with other "outputs".

docs/content/projects/BuildTrace.md
eugene-krivobokov: Updated the first screenshot and added a reference to it.

docs/content/projects/BuildTrace.md
eugene-krivobokov: Removed "avito" prefix

subprojects/common/graph/src/main/kotlin/com/avito/graph/Operation.kt
```diff
@@ -0,0 +1,14 @@
+package com.avito.graph
+
+interface Operation {
+    val id: String
+    val duration: Double
+
+    /**
+     * ids of predecessor operations
+     */
+    val predecessors: Set<String>
```
sergioboi, line 10: I did not meet a critical path vocabulary, is it a common name for graph relations?

subprojects/common/graph/src/main/kotlin/com/avito/graph/Operation.kt
```diff
@@ -0,0 +1,14 @@
...
+}
+
+fun Operation.isLeaf() =
```
sergioboi, line 13: I consider operation is leaf when it doesn't have followers, not predecessors

subprojects/common/graph/src/main/kotlin/com/avito/graph/ShortestPath.kt
```diff
@@ -0,0 +1,174 @@
...
+import org.jgrapht.graph.DirectedWeightedMultigraph
+
+/**
+ * @operations nodes describing a DAG
+ */
+class ShortestPath<T : Operation>(private val operations: Set<T>) {
+
+    private val operationByKey: Map<String, Operation> = operations.map { it.id to it }.toMap()
+
+    fun find(): List<T> {
```
sergioboi, line 16: I expect to see a more linked structure. It's not obvious that we return the path of `Operations` ordered by index
We could wrap the result with some `Path` class

subprojects/common/graph/src/main/kotlin/com/avito/graph/ShortestPath.kt
```diff
@@ -0,0 +1,174 @@
...
+     * F(2) <------------- G(2)
+     * ```
+     *
+     * Root nodes: E, G
+     * Leaf nodes: A, B, F
+     *
+     * This will be represented as:
+     *
+     * ```
+     *   ┌-----(1)-- A <--(3)---┐
```
sergioboi, line 45: I don't understand this notation
Explain it please at the meeting

subprojects/common/graph/src/main/kotlin/com/avito/graph/ShortestPath.kt
```diff
@@ -0,0 +1,174 @@
...
+        checkGraphStructure(graph)
+
+        return graph
+    }
+
+    private fun Graph<Operation, DefaultWeightedEdge>.shortestPath(): GraphPath<Operation, DefaultWeightedEdge> {
+        // We could use negative weights to find the shortest path as solution for the longest path in DAG
+        // Bellman–Ford algorithm supports negative weights but Dijkstra doesn't.
+        // TODO: try topological sorting, it will be more efficient - O(V+E) instead of O(V*E)
+        return BellmanFordShortestPath(this)
```
sergioboi, line 75: I would listen about that algorithm with pleasure too

subprojects/common/graph/src/main/kotlin/com/avito/graph/SimpleOperation.kt
```diff
@@ -0,0 +1,25 @@
+package com.avito.graph
+
+open class SimpleOperation(
+    override val id: String,
+    override val duration: Double = 1.0,
```
sergioboi, line 5: Do we need a default duration? It could lead to bugs

subprojects/common/graph/src/main/kotlin/com/avito/graph/SimpleOperation.kt
```diff
@@ -0,0 +1,25 @@
...
+    override val predecessors: Set<String> = emptySet()
+) : Operation {
+
+    override fun toString(): String {
```
sergioboi, line 9: Why do not you simply use `data class`?

subprojects/common/graph/src/test/kotlin/com/avito/grapth/ShortestPathTest.kt
```diff
@@ -0,0 +1,66 @@
...
+        val intermediate2 = SimpleOperation(id = "intermediate2", duration = 2.0, predecessors = setOf("start"))
+        val intermediate3 = SimpleOperation(id = "intermediate3", duration = 3.0, predecessors = setOf("start"))
+        val end = SimpleOperation(id = "end", predecessors = setOf("intermediate1", "intermediate2", "intermediate3"))
+
+        val longestPath = path(setOf(start, intermediate1, intermediate2, intermediate3, end))
+
+        assertThat(longestPath).containsExactly(start, intermediate1, end)
+    }
+
+    private fun path(operations: Set<Operation>) =
```
sergioboi, line 64: Propose to rename `path` to `computeShortestPath` it will help me read tests

subprojects/common/graph/src/test/kotlin/com/avito/grapth/ShortestPathTest.kt
```diff
@@ -0,0 +1,66 @@
...
+
+    @Test
+    fun `independent chains - the shortest chain`() {
+        val a1 = SimpleOperation(id = "a1", duration = 2.0)
+        val a2 = SimpleOperation(id = "a2", duration = 3.0, predecessors = setOf("a1"))
+
+        val b1 = SimpleOperation(id = "b1", duration = 3.0)
+        val b2 = SimpleOperation(id = "b2", duration = 3.0, predecessors = setOf("b1"))
+
+        val longestPath = path(setOf(a1, a2, b1, b2))
```
sergioboi, line 46: I'm confused, you make `ShortestPath.find` and then save it like a `longestPath`.
Do you find the shortest or longest?

subprojects/gradle/build-trace/src/gradleTest/kotlin/com/avito/android/build_trace/CriticalPathTest.kt
```diff
@@ -0,0 +1,253 @@
...
+    private lateinit var projectDir: File
+
+    @BeforeEach
+    fun setup(@TempDir tempDir: File) {
+        this.projectDir = tempDir
+    }
+
+    @Test
+    fun `has a predecessor task - input-output`() {
```
sergioboi, line 24: I don't understand what we test there
Could we rename the test?

subprojects/gradle/build-trace/src/gradleTest/kotlin/com/avito/android/build_trace/CriticalPathTest.kt
```diff
@@ -0,0 +1,253 @@
...
+            """.trimIndent()
+        )
+
+        val tasks = calculatePath(":consumerTask")
+
+        assertThat(tasks).containsExactly(":producerTask", ":consumerTask")
+    }
+
+    @Test
+    fun `has a predecessor task - dependsOn`() {
```
sergioboi, line 69: I thought we firstly write `the subject` and then `the expected result`
I would write the name of this test like `dependsOn - has a predecessor task`

subprojects/gradle/build-trace/src/main/kotlin/com/avito/android/build_trace/BuildTracePlugin.kt
```diff
@@ -17,22 +16,36 @@ open class BuildTracePlugin : Plugin<Project> {
...
+            output = File(pluginOutputDir(project), "build.trace"),
+            criticalPathProvider = criticalPathListener,
+            loggerFactory = loggerFactory
+        )
+        GradleCollector.initialize(project, listOf(criticalPathListener, buildTraceListener))
     }
+
+    private fun criticalPathListener(project: Project): CriticalPathListener {
+        val writer = CriticalPathSerialization(
+            report = File(pluginOutputDir(project), "critical_path.json")
```
sergioboi, line 44: I prefer to see an extension for  the `file path`

subprojects/gradle/build-trace/src/main/kotlin/com/avito/android/build_trace/internal/TaskExt.kt
```diff
@@ -0,0 +1,29 @@
...
+        return (
+            dependenciesByInputs +
+                taskDependencies.safeDependencies(this) +
+                mustRunAfter.safeDependencies(this) +
+                shouldRunAfter.safeDependencies(this)
+            )
+            .toSet()
+    }
+
+internal fun TaskDependency.safeDependencies(task: Task): Set<Task> {
```
sergioboi, line 22: Where do we use this Set?

subprojects/gradle/build-trace/src/main/kotlin/com/avito/android/build_trace/internal/TaskExt.kt
```diff
@@ -0,0 +1,29 @@
...
+    return try {
+        this.getDependencies(task)
+    } catch (error: Exception) {
```
sergioboi, line 25: Could we don't swallow this exception and use it to describe explicit result type
I prefer to use Failure then emptySet

subprojects/gradle/build-trace/src/main/kotlin/com/avito/android/build_trace/BuildTracePlugin.kt
```diff
@@ -17,22 +16,36 @@ open class BuildTracePlugin : Plugin<Project> {
...
+        val loggerFactory = GradleLoggerFactory.fromPlugin(this, project)
+
+        val criticalPathListener = criticalPathListener(project)
+
+        val buildTraceListener = BuildTraceListener(
+            output = File(pluginOutputDir(project), "build.trace"),
+            criticalPathProvider = criticalPathListener,
+            loggerFactory = loggerFactory
+        )
+        GradleCollector.initialize(project, listOf(criticalPathListener, buildTraceListener))
```
sergioboi, line 39: Do we need to unregister listeners at the end of the build?

subprojects/common/graph/src/main/kotlin/com/avito/graph/Operation.kt
eugene-krivobokov: In this case, "predecessor" defines incoming edges.
I found usage of this term in similar cases: in Gradle build scan and in the Activity-on-node diagram.

- https://en.wikipedia.org/wiki/Critical_path_method#Visualizing_critical_path_schedule
- https://project-management-knowledge.com/definitions/a/activity-on-node/

subprojects/common/graph/src/main/kotlin/com/avito/graph/Operation.kt
eugene-krivobokov: On the other hand, we have the opposite case. For now, they are named as "roots".
To avoid confusion, I'll use the correct original terms for DAG - sink and source.

subprojects/common/graph/src/main/kotlin/com/avito/graph/ShortestPath.kt
eugene-krivobokov: > we return the path of Operations ordered by index

This is by definition a path from a source to a sink in DAG. 
This path can't be ordered by any other reasons or attributes.

eugene-krivobokov: > We could wrap the result with some Path class

Good idea. Will do.

Originally didn't want to expose `org.jgrapht.GraphPath` to hide from client implementation details with the whole graph.

subprojects/common/graph/src/main/kotlin/com/avito/graph/SimpleOperation.kt
eugene-krivobokov: Can't eliminate it all because it's a weight of an edge. 
I'll try to make it `0`.

subprojects/common/graph/src/main/kotlin/com/avito/graph/SimpleOperation.kt
eugene-krivobokov: Sure, will do

subprojects/common/graph/src/main/kotlin/com/avito/graph/SimpleOperation.kt
sergioboi: I don't understand why we need a default value?
Why can't we set duration when creating an instance of the `SimpleOperation`?

subprojects/common/graph/src/test/kotlin/com/avito/grapth/ShortestPathTest.kt
eugene-krivobokov: Thanks, this is just a typo. In this class, we work with the shortest path.

I mixed it up by mistake because clients use it for the longest path. The "invert" operations for it.

subprojects/gradle/build-trace/src/main/kotlin/com/avito/android/build_trace/internal/TaskExt.kt
eugene-krivobokov: In this file, for example. 
To resolve dependent tasks described as `dependsOn` / `mustRunAfter` / `shouldRunAfter` / `finalizedBy`

subprojects/gradle/build-trace/src/main/kotlin/com/avito/android/build_trace/internal/TaskExt.kt
eugene-krivobokov: Ok, will wrap it. But clients anyway will ignore it.

subprojects/gradle/build-trace/src/main/kotlin/com/avito/android/build_trace/BuildTracePlugin.kt
eugene-krivobokov: Sure, this is the responsibility of `GradleCollector`.

subprojects/gradle/build-trace/src/main/kotlin/com/avito/android/build_trace/internal/TaskExt.kt
sergioboi: Clients should log can't build critical path because of `exception` and skip the critical path phase

subprojects/gradle/build-trace/src/main/kotlin/com/avito/android/build_trace/internal/TaskExt.kt
sergioboi: So it's incorrect to return emptySet

subprojects/gradle/build-trace/src/main/kotlin/com/avito/android/build_trace/internal/TaskExt.kt
eugene-krivobokov: We faced a bug in resolving dependencies: https://issuetracker.google.com/issues/142268503
I will change the logic and skip only this specific case. Otherwise, we will lose all the data. This bug was not specific for a critical path but for resolving dependencies in general.

subprojects/gradle/build-trace/src/main/kotlin/com/avito/android/build_trace/internal/TaskExt.kt
eugene-krivobokov: In general - incorrect
For https://issuetracker.google.com/issues/142268503 we have to preserve at least most of the data.
I'll change the logic:
https://github.com/avito-tech/avito-android/pull/973/files/2dbe5c6cd2cc7f4ceb0d37655aaa4b7200902a6b#r631650972

subprojects/common/graph/src/main/kotlin/com/avito/graph/SimpleOperation.kt
eugene-krivobokov: Agree, will remove

subprojects/gradle/build-trace/src/gradleTest/kotlin/com/avito/android/build_trace/CriticalPathTest.kt
eugene-krivobokov: As I see, we use mostly this pattern `state under test - expected behaviour - [when]`.

Samples:

```kotlin
@Test
fun `success runs - returns number of passed test runs`() {}

@Test
fun `read instrumentation output - completes stream - with failed test`() {}

@Test
fun `create summary - marks test as matched - all tests passed`() {}
```
 
I tend to think this way:
- "critical path" - state under test. I've omitted this part because of repeating.
-  "has a predecessor task" - expected behavior. This is ok.
- "dependsOn" - when part. This is what confuses me. I don't know to interpret this - as a condition or as a subject of testing.

> I would write the name of this test like dependsOn - has a predecessor task

Agree that "dependsOn" is what we really test here, in the wider context of a critical path.

Fixed in https://github.com/avito-tech/avito-android/pull/985/commits/7117e5971efda650b54c0cc5845ffb3bbe128128

subprojects/common/graph/src/main/kotlin/com/avito/graph/Operation.kt
eugene-krivobokov: Fixed in https://github.com/avito-tech/avito-android/pull/985/commits/f2038794bedf48b0e4f02583c5ad2f4826547df7

subprojects/common/graph/src/main/kotlin/com/avito/graph/ShortestPath.kt
eugene-krivobokov: Wrapped in model in https://github.com/avito-tech/avito-android/pull/985/commits/2afbb9a4146c27e65054c5d19f28c9e37580e4cb

subprojects/common/graph/src/main/kotlin/com/avito/graph/ShortestPath.kt
eugene-krivobokov: Described the notation on a simpler case in https://github.com/avito-tech/avito-android/pull/985/commits/f2038794bedf48b0e4f02583c5ad2f4826547df7

subprojects/common/graph/src/main/kotlin/com/avito/graph/SimpleOperation.kt
eugene-krivobokov: Fixed in https://github.com/avito-tech/avito-android/pull/985/commits/dc3d4bfb937bd64e9a9f58799f131da8871013a5

subprojects/common/graph/src/main/kotlin/com/avito/graph/SimpleOperation.kt
eugene-krivobokov: Fixed in https://github.com/avito-tech/avito-android/pull/985/commits/518d8ab1ef7375db97009368f70eac5c29fffa70

subprojects/common/graph/src/test/kotlin/com/avito/grapth/ShortestPathTest.kt
eugene-krivobokov: Renamed to `shortestPath` in https://github.com/avito-tech/avito-android/pull/985/commits/675c094180aaa4e130576c64ef4572add497eb73

subprojects/common/graph/src/test/kotlin/com/avito/grapth/ShortestPathTest.kt
eugene-krivobokov: Fixed in https://github.com/avito-tech/avito-android/pull/985/commits/675c094180aaa4e130576c64ef4572add497eb73

subprojects/gradle/build-trace/src/gradleTest/kotlin/com/avito/android/build_trace/CriticalPathTest.kt
eugene-krivobokov: Renamed to `input-output dependent task - path has a dependent task`
in https://github.com/avito-tech/avito-android/pull/985/commits/7117e5971efda650b54c0cc5845ffb3bbe128128

See also https://github.com/avito-tech/avito-android/pull/973#discussion_r631990853

subprojects/gradle/build-trace/src/main/kotlin/com/avito/android/build_trace/BuildTracePlugin.kt
eugene-krivobokov: Added an output dir to the extension in https://github.com/avito-tech/avito-android/pull/985/commits/beea22b429de726c0b41af61594ebc60d9f72441

subprojects/gradle/build-trace/src/main/kotlin/com/avito/android/build_trace/internal/TaskExt.kt
eugene-krivobokov: Removed the check at all because couldn't reproduce the issue.
https://github.com/avito-tech/avito-android/pull/985/commits/649bc7548b8c45d8cfbd158643d2681249186ab6

subprojects/gradle/build-trace/src/main/kotlin/com/avito/android/build_trace/internal/TaskExt.kt
eugene-krivobokov: Removed the check at all because couldn't reproduce the issue.
https://github.com/avito-tech/avito-android/pull/985/commits/649bc7548b8c45d8cfbd158643d2681249186ab6

eugene-krivobokov:
Look what I found.
Interesting to see similar decisions.

Screenshot from 2021-05-25 20-21-17

PR #977, opened 2021-05-11
log "unknown" network exceptions

dsvoronin:


eugene-krivobokov (approved):

subprojects/common/http-client/src/main/kotlin/com/avito/http/internal/StatsHttpEventListener.kt
```diff
@@ -58,6 +58,7 @@ internal class StatsHttpEventListener(
             statsDSender.send(TimeMetric(prefix.append("timeout"), latencyMs))
         } else {
             statsDSender.send(TimeMetric(prefix.append("unknown"), latencyMs))
+            logger.warn("Unexpected network problem", ioe)
```
sergioboi, line 61: I think we need more granularity
And send only really `Unexpected` problems

PR #984, opened 2021-05-12
fix "could not find dependency" error

materkey:
steps to reproduce:
- make stage_ui_tests verbose=DEBUG kubernetesContext=minikube

image

### DIFF
--- a/subprojects/android-test/ui-testing-core-app/build.gradle.kts
+++ b/subprojects/android-test/ui-testing-core-app/build.gradle.kts
@@ -35,11 +35,6 @@ android {
     }
 }
 
-@Suppress("UnstableApiUsage")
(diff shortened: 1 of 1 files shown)


dsvoronin:
Thanks @materkey !

To add more of the context:

i changed what version is used in `ui-testing-core-app` here https://github.com/avito-tech/avito-android/pull/963/commits/0e82f7fa1113209f3ddf6115614d3fbcaf0d2be1 to prevent a problem with different version of common test-runner libraries in gradle and in android test classpaths. 
It is a good change and should be returned, but for now we lost a way to test runner changes via maven local publication.

It should be done in separate module like it was tried in https://github.com/avito-tech/avito-android/pull/954

dsvoronin (approved):

PR #985, opened 2021-05-13
Mbs 8372 Critical path - code review fixes

eugene-krivobokov:
Fixes for #973
It's easier to review by commits


subprojects/common/graph/src/main/kotlin/com/avito/graph/SimpleOperation.kt
```diff
@@ -1,15 +1,11 @@
...
-    override val duration: Double = 1.0,
+    override val duration: Double,
     override val predecessors: Set<String> = emptySet()
 ) : Operation {
 
-    override fun toString(): String {
-        return "SimpleOperation(id=$id, duration=$duration, predecessors=$predecessors)"
-    }
-
     override fun equals(other: Any?): Boolean {
```
sergioboi, line 9: equals is redundant too

sergioboi (approved):

subprojects/common/graph/src/main/kotlin/com/avito/graph/SimpleOperation.kt
eugene-krivobokov: Keep it for now to compare only by ids

subprojects/detekt.yml
```diff
@@ -635,7 +635,7 @@ style:
         maxLineLength: 120
         excludePackageStatements: true
         excludeImportStatements: true
-        excludeCommentStatements: false
+        excludeCommentStatements: true
```
eugene-krivobokov, line 638: Excluded comments.
It's needed for code blocks and diagrams in KDoc.
I don't want to suppress the whole method\class because of it.

PR #991, opened 2021-05-14
MBS-7611 Fix missing mkdocs image

eugene-krivobokov:
Somehow an image has gone from a registry.

I've uploaded a new one to reuse a pipeline for images.


eugene-krivobokov:
Recovered in registry

PR #997, opened 2021-05-17
Fix client builder reusing: create copy of the builder for each `provide`

sergioboi:


dsvoronin (approved):

subprojects/common/http-client/src/main/kotlin/com/avito/http/HttpClientProvider.kt
```diff
@@ -14,15 +14,15 @@ public class HttpClientProvider(
...
 ) {
 
-    private val sharedBuilder = OkHttpClient.Builder()
+    private val client = OkHttpClient.Builder().build()
 
     public fun provide(
         requestMetadataProvider: RequestMetadataProvider = defaultRequestMetadataProvider()
     ): OkHttpClient.Builder {
-        return modifyExisting(sharedBuilder, requestMetadataProvider)
+        return provide(client.newBuilder(), requestMetadataProvider)
```
eugene-krivobokov, line 22: Let's add unit tests for this contract.

We can easily simulate the problem:
- First client gets a builder, modifies it
- Second client gets a builder, must not see changes

eugene-krivobokov (approved):

subprojects/common/http-client/src/main/kotlin/com/avito/http/HttpClientProvider.kt
sergioboi: Ok, I will make different PR with that test

PR #999, opened 2021-05-17
MBS-11063 Fix lint-report error - missing Severity class

eugene-krivobokov:
Fix lint report failure caused by missing class.
We can't use original lint dependencies in Gradle plugins due to a custom classloader in Lint (#989).

### DIFF
--- a/subprojects/gradle/cd/src/gradleTest/kotlin/com/avito/ci/steps/LintReportTest.kt
+++ b/subprojects/gradle/cd/src/gradleTest/kotlin/com/avito/ci/steps/LintReportTest.kt
@@ -1,0 +1,80 @@
+package com.avito.ci.steps
+
+import com.avito.test.gradle.TestProjectGenerator
+import com.avito.test.gradle.TestResult
+import com.avito.test.gradle.gradlew
+import com.avito.test.gradle.module.AndroidAppModule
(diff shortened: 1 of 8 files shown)


subprojects/gradle/cd/src/gradleTest/kotlin/com/avito/ci/steps/LintReportTest.kt
```diff
@@ -0,0 +1,80 @@
...
+    @Test
+    fun `lint check - does not send error report - valid project`() {
+        generateProject()
+
+        val buildResult = runBuild()
+        buildResult.assertThat()
+            .buildSuccessful()
+            .taskWithOutcome(":app:lintRelease", TaskOutcome.SUCCESS)
+            .taskWithOutcome(":app:lintReportToChannel", TaskOutcome.SUCCESS)
+    }
```
eugene-krivobokov, line 33: This test reproduces the problem.

I didn't test a transport in case or failed lint. It's covered partially by `LintSlackAlertIntegrationTest`.

In a Gradle test kit faced with restrictions:
- integTests don't support Gradle test kit
- Can't run integTests with slack locally

I've filed a separate issue to support these scenarios - MBS-11278

dsvoronin (approved):

PR #1001, opened 2021-05-17
MBS-11240-parallel-entry-loading

sergioboi:


subprojects/common/file-storage/src/main/kotlin/com/avito/filestorage/SettableFuture.kt
```diff
@@ -5,30 +5,46 @@ import java.util.concurrent.atomic.AtomicBoolean
...
     private val isSet = AtomicBoolean()
+    @Volatile
     private var slot: T? = null
 
     override fun get(): T {
         latch.await()
         return slot!!
     }
 
+    override fun tryGet(): T? {
```
dsvoronin, line 44: why this strange logic with 500ms delay instead of `latch.await(500ms)` with catching interruptException?

sergioboi: because `latch.await` will block the Thread

dsvoronin: ok, i will remove Future at all after enabling the flag

dsvoronin (approved):

PR #1002, opened 2021-05-17
MBS-11270 use jdoctor for error messages

dsvoronin:


subprojects/common/problem/src/test/kotlin/com/avito/android/SimpleProblemBuilderTest.kt
```diff
@@ -0,0 +1,46 @@
...
+                .withShortDescription { "something went wrong" }
+                .withLongDescription { "something really went wrong" }
+                .addSolution { builder -> builder.withShortDescription { "try to turn off and on again" } }
+                .addSolution { builder -> builder.withShortDescription { "is it really a problem?" } }
+        )
+
+        val result: RuntimeException = problem.asRuntimeException()
+
+        assertThat(result.message).isEqualTo(
+            """
```
dsvoronin, line 29: example of an error message

PR #1003, opened 2021-05-18
MBS-11268 remove suppressed failed tests from verdict

dsvoronin:
Suppressed failed tests was by mistake added to `Verdict.Failure`. Fixed it and added testcase

Also slightly changed text

was:
```
Failed. There are 1 not reported tests
Failed. There are 272 not suppressed failed tests
```

now:
```
Test suite failed. Problems:
 - Not reported tests: 1
 - Not suppressed failed tests: 272
```

LOST -> NOT REPORTED in problem tests

### DIFF
--- a/subprojects/test-runner/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/internal/finalizer/verdict/Verdict.kt
+++ b/subprojects/test-runner/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/internal/finalizer/verdict/Verdict.kt
@@ -19,7 +19,7 @@ internal sealed class Verdict {
 
     data class Failure(
         override val testResults: Collection<AndroidTest>,
-        val lostTests: Collection<AndroidTest.Lost>,
-        val failedTests: Collection<TestStaticData>
+        val notReportedTests: Collection<AndroidTest.Lost>,
+        val unsuppressedFailedTests: Collection<TestStaticData>
     ) : Verdict()
 }
--- a/subprojects/test-runner/instrumentation-tests/src/test/kotlin/com/avito/instrumentation/internal/finalizer/action/WriteTaskVerdictActionTest.kt
+++ b/subprojects/test-runner/instrumentation-tests/src/test/kotlin/com/avito/instrumentation/internal/finalizer/action/WriteTaskVerdictActionTest.kt
@@ -76,20 +76,20 @@ internal class WriteTaskVerdictActionTest {
                 ),
                 failedTest
             ),
-            failedTests = setOf(failedTest),
-            lostTests = emptyList()
+            unsuppressedFailedTests = setOf(failedTest),
+            notReportedTests = emptyList()
         )
         val action = createWriteTaskVerdictAction(tempDir)
         action.action(verdict)
 
         val verdictContent = readFile(tempDir)
 
-        assertThat(verdictContent).contains("Failed. There are 1 not suppressed failed tests")
+        assertThat(verdictContent).contains("Not suppressed failed tests: 1")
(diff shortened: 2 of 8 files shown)


subprojects/test-runner/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/internal/finalizer/verdict/Verdict.kt
```diff
@@ -19,7 +19,7 @@ internal sealed class Verdict {
 
     data class Failure(
         override val testResults: Collection<AndroidTest>,
-        val lostTests: Collection<AndroidTest.Lost>,
-        val failedTests: Collection<TestStaticData>
+        val notReportedTests: Collection<AndroidTest.Lost>,
+        val unsuppressedFailedTests: Collection<TestStaticData>
```
eugene-krivobokov, line 23: > unsuppressed

Should and can we know at this level the reason of failure?

subprojects/test-runner/instrumentation-tests/src/test/kotlin/com/avito/instrumentation/internal/finalizer/action/WriteTaskVerdictActionTest.kt
```diff
@@ -106,16 +106,46 @@ internal class WriteTaskVerdictActionTest {
...
         val verdictContent = readFile(tempDir)
 
-        assertThat(verdictContent).contains("Failed. There are 1 not reported tests")
-        assertThat(verdictContent).contains("com.test.Test2.test API22 LOST")
+        assertThat(verdictContent).contains("Not reported tests: 1")
+        assertThat(verdictContent).contains("com.test.Test2.test API22 NOT REPORTED")
+    }
+
+    @Test
+    fun `file - contains lost message without suppressed tests - verdict is failure`(@TempDir tempDir: File) {
```
eugene-krivobokov, line 122: I didn't get the naming pattern.
Is "verdict is failure" a condition? In this case, it doesn't explain it.

https://avito-tech.github.io/avito-android/contributing/Testing/#naming-conventions

eugene-krivobokov (approved):

subprojects/test-runner/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/internal/finalizer/verdict/Verdict.kt
dsvoronin: No need for now; If we will need it in future it's possible to pass whole `AndroidTest` object

subprojects/test-runner/instrumentation-tests/src/test/kotlin/com/avito/instrumentation/internal/finalizer/action/WriteTaskVerdictActionTest.kt
dsvoronin: yes it is. misleading test name. will fix it

subprojects/test-runner/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/internal/finalizer/verdict/Verdict.kt
eugene-krivobokov: I mean that on this level of abstraction it looks like `failed tests`. 
For clients of the verdict, these tests are the reason of failure. It's impossible to say here why they failed. Thus, knowledge about "suppressing" here looks excessive, at least.

PR #1005, opened 2021-05-18
MBS-8372 Suppress dependency resolving known issues in build trace plugin

eugene-krivobokov:
Now we suppress explicitly known issues with dependency resolution.
Couldn't reproduce them in a test project, but have them in avito project.
They are in rare conditions, not flaky but with uknonwn workaround.

### DIFF
--- a/subprojects/gradle/build-trace/src/main/kotlin/com/avito/android/build_trace/internal/TaskDependenciesResolutionResult.kt
+++ b/subprojects/gradle/build-trace/src/main/kotlin/com/avito/android/build_trace/internal/TaskDependenciesResolutionResult.kt
@@ -1,0 +1,61 @@
+package com.avito.android.build_trace.internal
+
+import com.avito.composite_exception.CompositeException
+import org.gradle.api.Task
+import org.gradle.api.internal.tasks.TaskDependencyResolveException
+
+/**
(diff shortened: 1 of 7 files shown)


subprojects/gradle/build-trace/src/main/kotlin/com/avito/android/build_trace/internal/TaskDependenciesResolutionResult.kt
```diff
@@ -0,0 +1,61 @@
...
+
+        private fun unexpectedError(reasons: List<Throwable>): CompositeException {
+            val message = "Unexpected dependency resolution errors. " +
+                "Possible solution: suppress it explicitly. See TaskDependenciesResolutionResult implementation."
+            return CompositeException(message, throwables = reasons.toTypedArray())
+        }
+
+        private fun Throwable.getKnownCauseOrNull(): Throwable? {
+            val cause = cause
+            // https://issuetracker.google.com/issues/188457864
```
eugene-krivobokov, line 50: Have this problem only in one sample app. It even doesn't have instrumentation tests.
Didn't find any loss of information. The final dependencies list is the same as we see in a build scan.

dsvoronin (approved):

PR #1008, opened 2021-05-19
MBS-11288 add more info for infrastructure errors

dsvoronin:
extract Problem.kt

### DIFF
--- a/subprojects/test-runner/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/internal/report/listener/Problem.kt
+++ b/subprojects/test-runner/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/internal/report/listener/Problem.kt
@@ -1,0 +1,85 @@
+package com.avito.instrumentation.internal.report.listener
+
+/**
+ * Inspired by https://github.com/melix/jdoctor
+ */
+internal data class Problem(
+    val shortDescription: String,
+    val context: String,
+    val because: String? = null,
+    val possibleSolutions: List<String> = emptyList(),
+    val documentedAt: String? = null,
+    val throwable: Throwable? = null
+) {
+
+    fun newBuilder(): Builder {
+        return Builder(this)
+    }
+
+    class Builder(private val shortDescription: String, private val context: String) {
+
+        private var because: String? = null
+        private val possibleSolutions = mutableListOf<String>()
+        private var documentedAt: String? = null
+        private var throwable: Throwable? = null
+
+        constructor(source: Problem) : this(source.shortDescription, source.context) {
+            this.because = source.because
+            this.possibleSolutions.addAll(source.possibleSolutions)
+            this.documentedAt = source.documentedAt
+            this.throwable = source.throwable
+        }
+
+        fun because(reason: String): Builder {
+            this.because = reason
+            return this
+        }
+
+        fun addSolution(solution: String): Builder {
+            this.possibleSolutions.add(solution)
+            return this
+        }
+
+        fun documentedAt(link: String): Builder {
+            this.documentedAt = link
+            return this
+        }
+
+        fun throwable(throwable: Throwable): Builder {
+            this.throwable = throwable
+            return this
+        }
+
+        fun build(): Problem {
+            return Problem(
+                shortDescription = this.shortDescription,
+                context = this.context,
+                because = because,
+                possibleSolutions = possibleSolutions,
+                documentedAt = documentedAt,
+                throwable = throwable,
+            )
+        }
+    }
+}
+
+internal fun Problem.asPlainText(): String = buildString {
+    appendLine(shortDescription)
+    appendLine("Where : $context")
+    appendLine("Why? : $because")
+    if (possibleSolutions.isNotEmpty()) {
+        appendLine("Possible solutions:")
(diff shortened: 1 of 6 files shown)


subprojects/test-runner/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/internal/report/listener/Problem.kt
```diff
@@ -0,0 +1,85 @@
+package com.avito.instrumentation.internal.report.listener
+
+/**
+ * Inspired by https://github.com/melix/jdoctor
+ */
+internal data class Problem(
```
dsvoronin, line 6: Will move to separate module in near future

subprojects/test-runner/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/internal/report/listener/ReportProcessorImpl.kt
```diff
@@ -52,85 +52,79 @@ internal class ReportProcessorImpl(
...
                             metricsSender.sendFailedOnStartInstrumentation()
+                            problemBuilder.because("Can't start test")
+                        }
 
-                        is TestCaseRun.Result.Failed.InfrastructureError.Timeout ->
+                        is TestCaseRun.Result.Failed.InfrastructureError.Timeout -> {
                             metricsSender.sendTimeOut()
+                            problemBuilder.because(
+                                "Test didn't finish in time. " +
+                                    "Test Runner has hardcoded timeout of $timeoutMin minutes"
```
eugene-krivobokov, line 104: > Test Runner has hardcoded timeout

This duplicates knowledge about implementation details. On this level, we shouldn't know it.

eugene-krivobokov (approved):

subprojects/test-runner/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/internal/report/listener/ReportProcessorImpl.kt
dsvoronin: I don't know how to deal with it better. 
On another level (where result is forming) i don't know about message detalization, it's just an exception message, not the same rich message for the end user like this `Problem`
Maybe it's a good idea to form problem on another level altogether

eugene-krivobokov: > Maybe it's a good idea to form problem on another level altogether

Or make problems more composable.
It makes sense to append different pieces of knowledge incrementally on different layers of abstraction.

Maybe even compose `Prolem` itself to explicitly express the difference.

```
- Problem("short description for end-user", "possible solution on high level - disable the whole feature")
caused by:
- Problem("short description for devs", "possible solution on implementation level - support\change these implementation details; check something")
```
The point is to give pieces on different levels of understanding. The more I know, the more I can understand from it.

dsvoronin: It can be done. But it would be really hard to present such error. Like who want 40 lines of problem descriptions, even chained ones.

subprojects/test-runner/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/internal/report/listener/ReportProcessorImpl.kt
```diff
@@ -52,85 +52,79 @@ internal class ReportProcessorImpl(
...
+                            )
+                        }
 
-                        is TestCaseRun.Result.Failed.InfrastructureError.Unexpected ->
+                        is TestCaseRun.Result.Failed.InfrastructureError.Unexpected -> {
                             metricsSender.sendUnexpectedInfraError()
+                        }
                     }
 
+                    val problem = problemBuilder.build()
```
sergioboi, line 113: Do we send this failure to the ReportViewer?

dsvoronin: Yes, as `incident.message`; and also as the message to kibana and sentry

sergioboi: Only message? What about a stack trace?

dsvoronin: Stacktrace passed as before, nothing changed in this regard

subprojects/test-runner/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/internal/report/listener/ReportProcessorImpl.kt
eugene-krivobokov: Agree, sound reasonable

PR #1010, opened 2021-05-19
extract `:common:problem` module

dsvoronin:
fix incident trace


subprojects/test-runner/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/internal/report/listener/ReportProcessorImpl.kt
```diff
@@ -153,7 +156,8 @@ internal class ReportProcessorImpl(
                         incident = Incident(
                             type = Incident.Type.INFRASTRUCTURE_ERROR,
                             timestamp = now,
-                            trace = problem.asRuntimeException().stackTraceToList(),
+                            trace = problem.throwable?.stackTraceToList()
+                                ?: problem.asRuntimeException().stackTraceToList(),
```
dsvoronin, line 160: @sboishtyan that would act as before, i introduced another layer of exception on top of original cause, which is kinda useless;

subprojects/common/problem/src/test/kotlin/com/avito/android/ProblemKtTest.kt
```diff
@@ -0,0 +1,37 @@
+package com.avito.android
+
+import com.google.common.truth.Truth.assertThat
+import org.junit.jupiter.api.Test
+
+internal class ProblemKtTest {
```
eugene-krivobokov:
```suggestion
internal class ProblemTest {
```

subprojects/common/problem/src/main/kotlin/com/avito/android/Problem.kt
```diff
@@ -12,45 +12,45 @@ internal data class Problem(
...
-    class Builder(private val shortDescription: String, private val context: String) {
+    public class Builder(private val shortDescription: String, private val context: String) {
 
         private var because: String? = null
         private val possibleSolutions = mutableListOf<String>()
         private var documentedAt: String? = null
         private var throwable: Throwable? = null
 
-        constructor(source: Problem) : this(source.shortDescription, source.context) {
+        internal constructor(source: Problem) : this(source.shortDescription, source.context) {
```
eugene-krivobokov, line 26:
```suggestion
        private constructor(source: Problem) : this(source.shortDescription, source.context) {
```
?

dsvoronin: can't ; private constructor is not available for `newBuilder()`

eugene-krivobokov (approved):

PR #1012, opened 2021-05-20
MBS-11299 Check that we realy create the dir

sergioboi:


dsvoronin (approved):

subprojects/test-runner/test-report-artifacts/src/main/kotlin/com/avito/report/UniqueDirTestArtifactsProvider.kt
```diff
@@ -17,8 +17,18 @@ internal class UniqueDirTestArtifactsProvider(
...
+                    require(parent.mkdirs()) {
+                        "Can't create parent dir for dir $this"
+                    }
+                }
+                require(!exists()) {
+                    "Dir $this already existed"
+                }
+                require(mkdir()) {
+                    "Can't create dir $this"
+                }
```
eugene-krivobokov, line 31: If this will help to diagnose the problem, let's schedule to use it as a common way. Maybe lint check or abstraction above File.

eugene-krivobokov (approved):

PR #1016, opened 2021-05-20
Make ReportState.kt thread-safe

sergioboi:


subprojects/test-runner/test-report/src/main/kotlin/com/avito/android/test/report/ReportState.kt
```diff
@@ -40,8 +46,8 @@ sealed class ReportState {
                 var dataSet: DataSet? = null,
                 var startTime: Long,
                 var endTime: Long = 0,
-                var preconditionStepList: MutableList<StepResult> = mutableListOf(),
-                var testCaseStepList: MutableList<StepResult> = mutableListOf()
+                var preconditionStepList: CopyOnWriteArrayList<StepResult> = CopyOnWriteArrayList(),
```
dsvoronin, line 49: Do we need to know about CopyOnWriteArrayList in constructor signature?

dsvoronin (approved):

subprojects/test-runner/test-report/src/main/kotlin/com/avito/android/test/report/ReportState.kt
sergioboi: Yes, because users could pass a wrong `MutableList` without that restriction

eugene-krivobokov: Do we need to make the same in `StepAttachments`?

sergioboi: StepAttachments doesn't allow to pass collections through constructor, so no

PR #1017, opened 2021-05-20
MBS-11263 fix k8s get pods request metadata; add verbose logging

dsvoronin:
Almost ready sample project; still some problems with getting successful suite run

### DIFF
--- a/samples/gradle.properties
+++ b/samples/gradle.properties
@@ -1,3 +1,6 @@
+avito.logging.verbosity=DEBUG
+avito.build=local
+#
 android.useAndroidX=true
 # Mandatory properties in root project for insturmentation plugin TODO remove all of them
 teamcityBuildId=1
--- a/samples/test-runner/build.gradle.kts
+++ b/samples/test-runner/build.gradle.kts
@@ -1,4 +1,7 @@
+import com.avito.instrumentation.reservation.request.Device.CloudEmulator
+import com.avito.kotlin.dsl.getMandatoryStringProperty
 import com.avito.kotlin.dsl.getOptionalStringProperty
+import com.avito.utils.gradle.KubernetesCredentials
 import com.avito.utils.gradle.kubernetesCredentials
 
 plugins {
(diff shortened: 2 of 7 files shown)


samples/gradle.properties
```diff
@@ -1,3 +1,6 @@
+avito.logging.verbosity=DEBUG
+avito.build=local
```
dsvoronin, line 2: @sboishtyan That was the reason for "teamcity" deployment label. new project and i forgot to set this flag. We even have TODO message about local by default

samples/test-runner/build.gradle.kts
```diff
@@ -29,12 +32,16 @@ android {
...
     androidTestImplementation("com.avito.android:test-inhouse-runner")
+    androidTestImplementation("com.avito.android:test-report")
+    androidTestImplementation("com.avito.android:logger")
     androidTestImplementation(libs.truth)
 
     androidTestUtil(libs.testOrchestrator)
 }
 
 instrumentation {
+    output = "$buildDir/reports/test-runner"
```
dsvoronin, line 43: using `buildDir` is much much better choice for local development

samples/test-runner/build.gradle.kts
```diff
@@ -73,8 +80,8 @@ instrumentation {
 
                     testsCountBasedReservation {
                         device = emulator29
-                        maximum = 2
-                        minimum = 1
```
dsvoronin, line 77: We have some weird bug with 1 min device. Deployments generation never succeed with it

samples/test-runner/build.gradle.kts
eugene-krivobokov: We can use it for CI build also

samples/test-runner/build.gradle.kts
eugene-krivobokov: File an issue, please

samples/test-runner/build.gradle.kts
dsvoronin: yep

eugene-krivobokov (approved):

PR #1025, opened 2021-05-25
Delete component1, component2 API because it pushes users to check manually what is null

sergioboi:
Add inline - helps using with coroutines

### DIFF
--- a/subprojects/common/result/src/main/kotlin/com/avito/android/Result.kt
+++ b/subprojects/common/result/src/main/kotlin/com/avito/android/Result.kt
@@ -13,10 +13,6 @@ import com.avito.composite_exception.CompositeException
  */
 sealed class Result<T> {
 
-    abstract operator fun component1(): T?
-
-    abstract operator fun component2(): Throwable?
-
     fun getOrThrow(): T = when (this) {
         is Success -> value
         is Failure -> throw throwable
@@ -27,30 +23,34 @@ sealed class Result<T> {
         is Failure -> func()
     }
 
-    inline fun <R> map(func: (value: T) -> R): Result<R> = when (this) {
+    inline fun <R> map(
+        mapSuccess: (value: T) -> R
+    ): Result<R> = when (this) {
         is Success -> try {
-            Success(func(value))
+            Success(mapSuccess(value))
         } catch (throwable: Throwable) {
             Failure(throwable)
         }
         is Failure -> Failure(throwable)
     }
 
-    inline fun <R> flatMap(func: (value: T) -> Result<R>): Result<R> = when (this) {
+    inline fun <R> flatMap(
+        mapSuccess: (value: T) -> Result<R>
+    ): Result<R> = when (this) {
         is Success -> try {
-            func(getOrThrow())
+            mapSuccess(value)
         } catch (e: Throwable) {
             Failure(e)
         }
         is Failure -> Failure(throwable)
     }
 
-    fun <R> fold(onSuccess: (value: T) -> R, onFailure: (throwable: Throwable) -> R): R = when (this) {
+    inline fun <R> fold(onSuccess: (value: T) -> R, onFailure: (throwable: Throwable) -> R): R = when (this) {
         is Success -> onSuccess(value)
         is Failure -> onFailure(throwable)
     }
 
-    fun recover(func: (Throwable) -> T): Result<T> = when (this) {
+    inline fun recover(func: (Throwable) -> T): Result<T> = when (this) {
         is Success -> this
         is Failure -> try {
             Success(func(throwable))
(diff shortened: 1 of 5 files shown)


subprojects/common/result/src/main/kotlin/com/avito/android/Result.kt
```diff
@@ -27,30 +23,36 @@ sealed class Result<T> {
...
     }
 
-    fun <R> fold(onSuccess: (value: T) -> R, onFailure: (throwable: Throwable) -> R): R = when (this) {
+    inline fun <R> fold(onSuccess: (value: T) -> R, onFailure: (throwable: Throwable) -> R): R = when (this) {
         is Success -> onSuccess(value)
         is Failure -> onFailure(throwable)
     }
 
-    fun recover(func: (Throwable) -> T): Result<T> = when (this) {
+    inline fun recover(func: (Throwable) -> T): Result<T> = when (this) {
```
sergioboi, line 53: inline for corouines

subprojects/common/result/src/main/kotlin/com/avito/android/Result.kt
```diff
@@ -13,10 +13,6 @@ import com.avito.composite_exception.CompositeException
  */
 sealed class Result<T> {
 
-    abstract operator fun component1(): T?
```
sergioboi, line 16: This API bad because it pushes users to check manually what is null and that leads to bugs

subprojects/common/result/src/main/kotlin/com/avito/android/Result.kt
```diff
@@ -27,30 +23,36 @@ sealed class Result<T> {
         is Failure -> func()
     }
 
-    inline fun <R> map(func: (value: T) -> R): Result<R> = when (this) {
+    inline fun <R> map(
+        mapFailure: (Throwable) -> Throwable = { it -> it },
```
sergioboi: We often have a case when we want to add details for throwable

subprojects/test-runner/test-report/src/main/kotlin/com/avito/android/test/report/video/VideoCapturer.kt
```diff
@@ -37,29 +37,18 @@ class VideoCapturerImpl(
     override fun start(): Result<Unit> {
         // checks if execute start() concurrently
         return if (state is State.Idling) {
-
-            val (videoFile, videoError) = testArtifactsProvider.generateUniqueFile("mp4")
-
-            if (videoError != null) {
```
sergioboi, line 43: for example, we check that the video error is not null so assume that the video file is null
It pushed us to use force unwrap

subprojects/test-runner/test-report/src/main/kotlin/com/avito/android/test/report/video/VideoCapturer.kt
```diff
@@ -37,29 +37,18 @@ class VideoCapturerImpl(
...
-                        Result.Success(Unit)
-                    } catch (t: Throwable) {
-                        if (videoFile!!.exists()) {
-                            videoFile.delete()
-                        }
-                        Result.Failure(IllegalStateException("Can't start video capturing", t))
-                    }
-                }
-            }
+            testArtifactsProvider.generateUniqueFile("mp4").flatMap(
```
sergioboi: As for me in that code harder to make a bug because the compile check all nulls for you

subprojects/common/result/src/main/kotlin/com/avito/android/Result.kt
```diff
@@ -27,30 +23,36 @@ sealed class Result<T> {
         is Failure -> func()
     }
 
-    inline fun <R> map(func: (value: T) -> R): Result<R> = when (this) {
+    inline fun <R> map(
+        mapFailure: (Throwable) -> Throwable = { failure -> failure },
+        mapSuccess: (value: T) -> R
```
eugene-krivobokov, line 27: let's switch positions for the sake of consistency:

```
1. success
2. failure
```

In other functions, we stick to this order.

eugene-krivobokov (approved):

subprojects/common/result/src/main/kotlin/com/avito/android/Result.kt
sergioboi: I can't switch because this order helps to write code like
`result.map { }` because we have lambda as a last parameter

subprojects/test-runner/service/src/main/kotlin/com/avito/runner/service/worker/DeviceWorker.kt
```diff
@@ -136,62 +123,62 @@ internal class DeviceWorker(
...
+                device = device,
+                test = action.test,
+                targetPackage = action.targetPackage,
+                result = deviceTestCaseRun.testCaseRun.result,
+                durationMilliseconds = deviceTestCaseRun.testCaseRun.durationMilliseconds,
+                executionNumber = action.executionNumber,
+                testArtifactsDir = testArtifactsDir
+            )
 
-        testArtifactsDir.map { dir ->
```
sergioboi, line 187: I found two bugs that:
- If `testArtifactsDir` was Failed then map `would` never be called
- After onDeviceDied we must end the coroutine with `return` invocation.

PR #1028, opened 2021-05-26
MBS-11331 Fix previous run filter

sergioboi:


subprojects/test-runner/report/src/main/kotlin/com/avito/android/runner/report/internal/AvitoReport.kt
```diff
@@ -117,10 +116,8 @@ internal class AvitoReport(
         }
     }
 
-    override fun getTests(initialSuiteFilter: List<TestName>): Result<List<SimpleRunTest>> {
-        return reportsApi.getTestsForRunId(reportCoordinates).map { list ->
-            list.filter { TestName(it.className, it.methodName) in initialSuiteFilter }
```
eugene-krivobokov, line 122: For the context
We were losing tests from a report here. The list was empty.

eugene-krivobokov (approved):

PR #1032, opened 2021-05-28
MBS-11309 Use common prefix for build metrics

eugene-krivobokov:


docs/content/projects/internal/BuildMetrics.md
```diff
@@ -27,31 +27,25 @@ They will be referred in docs as `<placeholder>`.
 
 [Http build cache](https://docs.gradle.org/current/userguide/build_cache.html#sec:build_cache_configure_remote) errors:
 
-- `<namespace>.build.cache.errors.[load|store].<http status code>`: errors counter
+- `<namespace>.<environment>.<node>.build.cache.errors.[load|store].<http status code>`: errors counter
```
eugene-krivobokov, line 30: Added common prefix to metrics.
These attributes turned out to be helpful for all build metrics. Thus, made them common.
Further decomposition can be done in specific metrics.

docs/content/projects/internal/BuildMetrics.md
```diff
@@ -27,31 +27,25 @@ They will be referred in docs as `<placeholder>`.
...
-- `build-time.total` (time in ms): total build time
+- `<namespace>.<environment>.<node>.id.<build status>.init_configuration.total` (time in ms): initialization with configuration time
+- `<namespace>.<environment>.<node>.id.<build status>.build-time.total` (time in ms): total build time
 
 ### Tasks metrics
 
-Prefix: `<namespace>.<environment>.<node>.id.<build status>.tasks`
-
-- `executed.<module path>.<task name>.total` (time in ms): top slowest tasks execution time
-- `from_cache.miss` (gauge 0..100): executed tasks ratio.
```
eugene-krivobokov, line 50: Removed because they already replaced by `build.cache` metrics.

docs/content/projects/internal/BuildMetrics.md
```diff
@@ -27,31 +27,25 @@ They will be referred in docs as `<placeholder>`.
...
-This is not a real cache misses. See a bug MBS-7244.
+- `<namespace>.<environment>.<node>.id.<build status>.tasks.executed.<module path>.<task name>.total` (time in ms): 
+  top slowest tasks execution time. It will be removed in MBS-11309.
 
 ### Specific build events
 
-Prefix: `<namespace>.<environment>.<node>.id.<build status>`
```
eugene-krivobokov, line 55: Inlined in docs. Found harder to compose mentally the final value.

subprojects/common/statsd/src/main/kotlin/com/avito/android/stats/StatsDSender.kt
```diff
@@ -11,8 +11,6 @@ interface StatsDSender {
 
     fun send(metric: StatsMetric)
 
-    fun send(prefix: SeriesName, metric: StatsMetric)
```
eugene-krivobokov, line 14: Want to simplify the interface. Building an metric should the responsibility of clients.

subprojects/gradle/build-environment/src/main/kotlin/com/avito/utils/gradle/Environment.kt
```diff
@@ -1,11 +1,8 @@
 package com.avito.utils.gradle
 
-/**
- * @param publicName is used in graphite events
```
eugene-krivobokov, line 4: Common models shouldn't know about this specific use case.

subprojects/gradle/impact/build.gradle.kts
```diff
@@ -15,6 +15,7 @@ dependencies {
     implementation(project(":common:math"))
     implementation(project(":gradle:git"))
     implementation(project(":gradle:build-environment"))
+    implementation(project(":gradle:build-metrics"))
```
eugene-krivobokov: - [x] extract to a module. Shouldn't consume plugin

dsvoronin: it should be done in this PR or postponed?

subprojects/gradle/build-metrics/src/test/kotlin/com/avito/android/plugin/build_metrics/BuildMetricTrackerTest.kt
```diff
@@ -0,0 +1,84 @@
...
+import com.avito.android.stats.StubStatsdSender
+import com.avito.android.stats.TimeMetric
+import com.avito.utils.gradle.Environment
+import com.google.common.truth.Truth.assertThat
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+
+internal class BuildMetricTrackerTest {
+
+    private lateinit var statsd: StubStatsdSender
```
dsvoronin:
```suggestion
    private val statsd = StubStatsdSender()
```

`lateinit` only adds complexity here; junit default blocking runner creates new instance of `BuildMetricTrackerTest` for each test invocation

image

subprojects/gradle/build-metrics/src/test/kotlin/com/avito/android/plugin/build_metrics/BuildMetricTrackerTest.kt
```diff
@@ -0,0 +1,84 @@
...
+        val metrics = statsd.getSentMetrics()
+        assertThat(metrics).hasSize(1)
+        assertThat(metrics.first().name.toString()).isEqualTo("local.user.build.metric")
+    }
+
+    @Test
+    fun `build - CI environment`() {
+        val tracker = metricTracker(
+            environment = Environment.CI,
+            node = "agent-3c23034b"
```
dsvoronin: why node value is missing in assertion? was it intended? if so it not clear for me from test name

dsvoronin (approved):

subprojects/gradle/impact/build.gradle.kts
eugene-krivobokov: Moved to `build-metrics-tracker` module

subprojects/gradle/build-metrics/src/test/kotlin/com/avito/android/plugin/build_metrics/BuildMetricTrackerTest.kt
eugene-krivobokov: It has state. I want to recreate it explicitly. It seems easier for me while reading the code, to don't rely on implicit contract from test runner.

subprojects/gradle/build-metrics/src/test/kotlin/com/avito/android/plugin/build_metrics/BuildMetricTrackerTest.kt
eugene-krivobokov: Yes, it's intended behavior. A node in CI is ephemeral because of a container. It's no use in metrics.

Added to test name.

subprojects/gradle/build-metrics/src/test/kotlin/com/avito/android/plugin/build_metrics/BuildMetricTrackerTest.kt
dsvoronin: i see, i prefer private factory function then, also to avoid `lateinit`

dsvoronin (approved):

subprojects/gradle/build-metrics/src/test/kotlin/com/avito/android/plugin/build_metrics/BuildMetricTrackerTest.kt
eugene-krivobokov: Fixed, changed to a factory and create in each test explicitly.

docs/content/projects/internal/BuildMetrics.md
sergioboi, line 30: I want to see series name builder compatible with this concept
For example:
```
SeriesNameBuilder()
.addNamespace("namespace")
.addEnvironment(Environment.CI)
addNode() - I don't know what it is
.addProduct/Service("build.cache")
.addMetric("errors.load")
.addMetricState("200")
```

eugene-krivobokov: I see the idea. I've considered this model also but decided to use only original stastd model. At least for now.

On this example:
- Namespace: it is on low level of statsd transport. Client can't override it in a metric.
- Environment, node (this is mostly a "user" for simplicity): want to see them as required. I tried couple of time to make them optional, but alway found that we need them in all metrics.
- Product/service VS metric: I can't describe the different between these terms. If I introduce a new term or abstraction, I need to understand cases, problems around it. For now, the difference is not clear and more confuses me.
- Metric state: this is what metric type describes

So, it turns out a couple of options:

```
buildMetricSeriesNameBuilder   // with required defaults
	.addMetric("build.cache.errors.load")    // series name
	.build() 				   // builds series name
```

Or an abstraction above stasd metric:

```
buildMetricBuilder	
	.addMetric("build.cache.errors.load") 		// series name
	.timeMs() |gauge() | count()    // builds statsd metric
```

eugene-krivobokov: Discussed internally, the idea:

```
@Service("build")
interface Build {
	@Path("cache")	
	val cache: Cache
	interface Cache {
		@Path("errors")
		val cache: Errors
		interface Errors {
			@GaugeMetric("load")
			fun load(value: Long)
		}
	}
}
build.cache.errors.load(100)
```

PR #1035, opened 2021-05-28
MBS-11322 Optimize process runner

sergioboi:


subprojects/gradle/process/src/main/kotlin/com/avito/utils/ProcessRunner.kt
```diff
@@ -1,129 +1,147 @@
...
+    fun run(command: String, timeout: Long, unit: TimeUnit): Result<String>
+
     fun run(command: String): Result<String>
 
+    fun spawn(command: String, outputTo: File?): Process
+
     class Real(
         private val workingDirectory: File?,
-        private val loggerFactory: LoggerFactory
+        private val cliOutputReaderThreadPool: ExecutorService =
```
sergioboi, line 30: Pool for threads that reads cli output:
before -  2 new thread for each cli operation without caching
now - max 300 threads, keep alive time=30s. Zero impact after 30 sec idle

subprojects/gradle/process/src/main/kotlin/com/avito/utils/ProcessRunner.kt
```diff
@@ -1,129 +1,147 @@
...
+                workingDirectory = workingDirectory
+            )
+            val outputFuture = cliOutputReaderThreadPool.submit(
+                CLIOutputReaderTask(
+                    stdout = process.inputStream,
+                    stderr = process.errorStream
+                )
+            )
+            val output = if (timeout != null) {
+                process.waitFor(timeout.time, timeout.unit)
```
sergioboi: before we could hang and that blocks thread from what we invoke `process run` for example from KubernetesReservationProvider when initializing device

Now we could configure the timeout for CLI command if need

subprojects/test-runner/device-provider/src/main/kotlin/com/avito/android/runner/devices/internal/AbstractDevice.kt
```diff
@@ -25,13 +28,15 @@ internal abstract class AbstractDevice(protected val loggerFactory: LoggerFactor
             " ${tags.joinToString(" ")}"
         }
 
-        executeNonBlockingCommand(
-            command = "logcat$tagsString",
-            redirectOutputTo = file
+        // TODO stop endless process
```
sergioboi, line 31: Bug: we don't know how this process could load our builder when we free pod connected with that process

subprojects/test-runner/device-provider/src/main/kotlin/com/avito/android/runner/devices/internal/AbstractDevice.kt
```diff
@@ -25,13 +28,15 @@ internal abstract class AbstractDevice(protected val loggerFactory: LoggerFactor
...
+        processRunner.spawn(
+            command = "$adb -s $serial logcat$tagsString",
+            outputTo = file
         )
     }
 
-    protected fun isBootCompleted() = executeCommand(CHECK_BOOT_COMPLETED_COMMAND)
+    protected fun isBootCompleted() =
+        processRunner.run(CHECK_BOOT_COMPLETED_COMMAND, timeout = 10, unit = TimeUnit.SECONDS)
```
sergioboi: Wait for boot only 10 seconds per operation then TimeoutException

subprojects/test-runner/device-provider/src/main/kotlin/com/avito/android/runner/devices/internal/RemoteDeviceImpl.kt
```diff
@@ -5,20 +5,22 @@ import com.avito.logger.LoggerFactory
...
     override val adb: Adb,
-    loggerFactory: LoggerFactory
-) : AbstractDevice(loggerFactory), RemoteDevice {
+    loggerFactory: LoggerFactory,
+    processRunner: ProcessRunner
+) : AbstractDevice(loggerFactory, processRunner), RemoteDevice {
 
-    override fun disconnect(): Result<String> = processRunner.run(command = "$adb disconnect $serial")
+    override fun disconnect(): Result<String> =
+        processRunner.run(command = "$adb disconnect $serial", timeout = 10, unit = SECONDS)
```
sergioboi: Wait for disconnect only 10 seconds per operation then TimeoutException

subprojects/test-runner/device-provider/src/main/kotlin/com/avito/android/runner/devices/internal/RemoteDeviceImpl.kt
```diff
@@ -5,20 +5,22 @@ import com.avito.logger.LoggerFactory
...
 
     override fun connect(): Result<String> {
         disconnect()
 
-        return processRunner.run("$adb connect $serial")
+        return processRunner.run("$adb connect $serial", timeout = 30, unit = SECONDS)
```
sergioboi: Wait for connect only 30 seconds per operation then TimeoutException

subprojects/gradle/process/src/main/kotlin/com/avito/utils/ProcessRunner.kt
```diff
@@ -1,129 +1,137 @@
...
+        @VisibleForTesting
+        fun splitCommand(source: String): Array<String> = Commandline.translateCommandline(source)
+
+        private fun spawnProcess(
+            command: String,
+            workingDirectory: File? = null,
+            outputTo: File? = null
+        ): Process {
+            return ProcessBuilder(*splitCommand(command))
+                .redirectErrorStream(true)
```
sergioboi, line 61: Redirect stderr to stdout, it helps to read the whole output from `process.inputStream`

subprojects/gradle/process/src/main/kotlin/com/avito/utils/ProcessRunner.kt
```diff
@@ -1,129 +1,137 @@
...
+            val process = spawnProcess(
+                command = command,
+                workingDirectory = workingDirectory
+            )
+            val outputFuture = cliOutputReaderThreadPool.submit(
+                CLIOutputReaderTask(
+                    stdout = process.inputStream
+                )
+            )
+            val output = if (timeout != null) {
```
eugene-krivobokov: How about making timeout mandatory with reasonable defaults?
I don't know cases for real infinite commands.
If so, I want to move this decision to a client-side and make it explicit.

subprojects/gradle/process/src/main/kotlin/com/avito/utils/ProcessRunner.kt
```diff
@@ -1,129 +1,137 @@
...
+        val stdoutBuffer = BufferedReader(InputStreamReader(stdout))
+        while (true) {
+            val line = stdoutBuffer.readLine()
+            if (line != null) {
+                result.appendLine(line)
+            } else {
+                break
+            }
         }
+        return result.toString().trim()
```
eugene-krivobokov, line 117: Why do we trim the stdout?
What if spaces are meaningful? We can't know it on this level.

sergioboi: I don't know but trim works only for the first and last space, so it's not critical but I could remove this logic

subprojects/gradle/process/src/main/kotlin/com/avito/utils/ProcessRunner.kt
sergioboi: I agree with you

subprojects/gradle/process/src/main/kotlin/com/avito/utils/ProcessRunner.kt
```diff
@@ -1,129 +1,137 @@
...
+                    }
+                    if (outputTo != null) {
+                        builder.redirectOutput(outputTo)
+                    }
+                }
+                .start()
+        }
 
-    val output = processOutput.toString().trim()
+        private class Timeout(val time: Long, val unit: TimeUnit)
```
eugene-krivobokov: Can we use `java.time.Duration` instead?
I know only about restrictions for Android libraries due to java stdlib incompleteness.
As I see this library is only for Gradle plugins.

eugene-krivobokov (approved):

dsvoronin (approved):

PR #1036, opened 2021-06-01
MBS-11263 Provide logcat from Incomplete tests

dsvoronin:
Previously it was only available via LogcatBuffer, tailing logcat file during test execution; now it's needed to provide access with already populated string with logcat of Incomplete test

Logcat will be fully available only under "stdout" section of ReportViewer
Previous stdout/stderr separation was silly and misguiding (take a look and separation logic in diff)

Possible future improvements: to make logcat more readable some tags could be filtered out somewhere near LogcatAccessor logic, instead of feeding configuration with whitelist


subprojects/common/report-viewer/src/main/kotlin/com/avito/report/internal/ReportsAddApiImpl.kt
```diff
@@ -213,12 +209,12 @@ internal class ReportsAddApiImpl(private val client: JsonRpcClient) : ReportsAdd
...
             // тут происходит магия "с помощью оператора добавляется в массив новое значение билда"
-            // todo узнать у Жени как можно это лучше сделать
             "report_data" to reportData,
             // todo onlyIf present
             "console" to mapOf(
-                "stdout" to stdout,
-                "stderr" to stderr
+                "stdout" to logcat,
+                // non empty value causes warning sign to appear in UI
+                "stderr" to ""
```
eugene-krivobokov, line 216: How does it look?
Should we express it explicitly for a client instead of empty value?

subprojects/test-runner/client/src/main/kotlin/com/avito/runner/scheduler/listener/ArtifactsTestListener.kt
```diff
@@ -63,12 +63,20 @@ internal class ArtifactsTestListener(
                 }
                 TestResult.Complete(artifacts)
             }
+
             is InfrastructureError ->
-                TestResult.Incomplete(result)
+                TestResult.Incomplete(
+                    infraError = result,
+                    logcat = device.logcat(LOGCAT_LINES_TO_FETCH)
```
eugene-krivobokov:
```suggestion
                    logcat = device.logcat(LOGCAT_LAST_LINES)
```
The current behavior is obvious. Let's explain the motivation.

subprojects/test-runner/client/src/main/kotlin/com/avito/runner/scheduler/listener/ArtifactsTestListener.kt
```diff
@@ -63,12 +63,20 @@ internal class ArtifactsTestListener(
...
+                )
+
             Ignored ->
                 TestResult.Incomplete(
                     InfrastructureError.Unexpected(
                         IllegalStateException("Instrumentation executed Ignored test")
+                    ),
+                    logcat = Result.Failure(
```
eugene-krivobokov, line 78: It looks that Result is not suitable for this property.
Clients can't distinguish "empty\disabled" values from real errors.

subprojects/test-runner/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/internal/logcat/LogcatAccessor.kt
```diff
@@ -0,0 +1,12 @@
+package com.avito.instrumentation.internal.logcat
+
+internal interface LogcatAccessor {
+
+    val isAvailable: Boolean
+
+    /**
+     * Logcat will be populated with empty lists if not available;
+     * Check [isAvailable] to avoid useless work
```
eugene-krivobokov: It's easy to forget about "if" before usage. Nothing will remind or force to do it.
It'd be good to express this by types, e.g. sealed class.

subprojects/test-runner/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/internal/logcat/ResultLogcatAccessor.kt
```diff
@@ -0,0 +1,13 @@
...
+import com.avito.android.Result
+
+internal class ResultLogcatAccessor(val result: Result<String>) : LogcatAccessor {
+
+    override val isAvailable: Boolean
+        get() = result.isSuccess()
+
+    override fun getLogs(): Logcat {
+        return result.map { Logcat(it) }.getOrElse { Logcat.STUB }
+    }
```
eugene-krivobokov, line 23: Example for https://github.com/avito-tech/avito-android/pull/1036/files#r643426727

I will invoke `getLogs()` and get empty string.

subprojects/common/report-viewer/src/main/kotlin/com/avito/report/internal/ReportsAddApiImpl.kt
dsvoronin: image

Like this. We can but it will be an invalid link probably

dsvoronin: I wanted to add "All logcat output goes to stdout" message to all stderr's, but it seems like a huge waste of traffic. It will be added to tens of thousands tests a day... 
Should be probably handled on report frontend side

subprojects/test-runner/client/src/main/kotlin/com/avito/runner/scheduler/listener/ArtifactsTestListener.kt
dsvoronin: Agreed, please take a look at new `LogcatResult`

subprojects/test-runner/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/internal/logcat/ResultLogcatAccessor.kt
dsvoronin: https://github.com/avito-tech/avito-android/pull/1036#discussion_r643462103

subprojects/test-runner/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/internal/logcat/LogcatAccessor.kt
dsvoronin: https://github.com/avito-tech/avito-android/pull/1036#discussion_r643462103

subprojects/common/report-viewer/src/main/kotlin/com/avito/report/internal/ReportsAddApiImpl.kt
eugene-krivobokov: > Should be probably handled on report frontend side

Indeed

subprojects/test-runner/client/src/main/kotlin/com/avito/runner/scheduler/listener/ArtifactsTestListener.kt
eugene-krivobokov: I see, `LogcatResult` looks solving the issue.
Can we use it here?

eugene-krivobokov (approved):

subprojects/test-runner/client/src/main/kotlin/com/avito/runner/scheduler/listener/ArtifactsTestListener.kt
dsvoronin: we should, but can't right now, it needs to be refactored in a way that most internal classes of logcat and report processing moved to `:client` module

subprojects/test-runner/client/src/main/kotlin/com/avito/runner/scheduler/listener/ArtifactsTestListener.kt
```diff
@@ -84,3 +92,5 @@ internal class ArtifactsTestListener(
         }
     }
 }
+
+private const val LOGCAT_LAST_LINES = 300
```
sergioboi, line 96: Why 300? Could we send all lines?

subprojects/test-runner/service/src/main/kotlin/com/avito/runner/service/worker/device/adb/AdbDevice.kt
```diff
@@ -435,6 +435,28 @@ data class AdbDevice(
         }
     )
 
+    override fun logcat(lines: Int): Result<String> {
+        return retryAction.retry(
```
sergioboi, line 439: One more `retry` action (

subprojects/test-runner/client/src/main/kotlin/com/avito/runner/scheduler/listener/ArtifactsTestListener.kt
sergioboi: I don't get why we can't call `logcatBuffers.get(key)` when test is incomplete?

subprojects/test-runner/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/internal/logcat/BufferLogcatAccessor.kt
```diff
@@ -0,0 +1,16 @@
+package com.avito.instrumentation.internal.logcat
+
+import com.avito.android.Problem
+
+internal class BufferLogcatAccessor(private val logcatBuffer: LogcatBuffer?) : LogcatAccessor {
```
sergioboi, line 5: We could check `logcatBuffers.get(key)` and if there is `null` then call `device.logcat()`

subprojects/test-runner/client/src/main/kotlin/com/avito/runner/scheduler/listener/ArtifactsTestListener.kt
dsvoronin: because logcat buffers needs start and stop events, and such incomplete tests doesn't have these callbacks called

subprojects/test-runner/client/src/main/kotlin/com/avito/runner/scheduler/listener/ArtifactsTestListener.kt
dsvoronin: all lines count is unknown, and could be a big file; in fact:

```
adb logcat -g

main: ring buffer is 2 MiB (1 MiB consumed), max entry is 5120 B, max payload is 4068 B
system: ring buffer is 2 MiB (307 KiB consumed), max entry is 5120 B, max payload is 4068 B
crash: ring buffer is 2 MiB (0 B consumed), max entry is 5120 B, max payload is 4068 B
```
for our 29 emulator

we could alter this by `adb logcat -G <SIZE>` command

PR #1040, opened 2021-06-02
MBS-11263 expose suppress flags

dsvoronin:
…sion


dsvoronin:
This implementation is ugly, don't want to merge it

Temporary workaround:

afterEvaluate {
   find needed instrumentation task and set `suppressFailure` or `suppressFlaky` directly, see example here: https://github.com/avito-tech/avito-android/blob/7b6e591e936c589e0bc12786d8c633f4980306c6/subprojects/gradle/cd/src/main/kotlin/com/avito/ci/steps/UiTestCheck.kt#L36
}

cc @materkey

PR #1042, opened 2021-06-03
MBS-11263 add experimental flag for fetchLogcatForIncompleteTests

dsvoronin:
there were some (maybe unrelated) problems on avito integration
false by default

### DIFF
--- a/subprojects/test-runner/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/InstrumentationTestsTask.kt
+++ b/subprojects/test-runner/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/InstrumentationTestsTask.kt
@@ -84,6 +84,9 @@ public abstract class InstrumentationTestsTask @Inject constructor(
     public val useInMemoryReport: Property<Boolean> = objects.property()
 
     @Input
+    public val fetchLogcatForIncompleteTests: Property<Boolean> = objects.property()
+
+    @Input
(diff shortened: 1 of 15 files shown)


subprojects/test-runner/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/InstrumentationTestsTask.kt
```diff
@@ -83,6 +83,9 @@ public abstract class InstrumentationTestsTask @Inject constructor(
     @Input
     public val useInMemoryReport: Property<Boolean> = objects.property()
 
+    @Input
+    public val fetchLogcatForIncompleteTests: Property<Boolean> = objects.property()
+
```
eugene-krivobokov, line 88: How about grouping experimental toggles in an abstraction?
We could pass it as one container on different levels.

eugene-krivobokov (approved):

subprojects/test-runner/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/InstrumentationTestsTask.kt
dsvoronin: i thought about it;
different flags goes to different abstractions so it's probably make sense to create multiple `flag bundles` based on it's destination. For example some of the flags goes to executor, some to reports.

Maybe it's not a big deal and we can move around full blob of experimental flags everywhere.

PR #1044, opened 2021-06-03
MBS-11263 update junit 5 + platform versions

dsvoronin:
https://junit.org/junit5/docs/current/release-notes/index.html#release-notes-5.7.2

platform version was out of sync with junit5


libraries/src/main/kotlin/com/avito/android/ExternalLibrariesExtension.kt
```diff
@@ -12,8 +12,8 @@ abstract class ExternalLibrariesExtension @Inject constructor(private val provid
         const val sentry = "1.7.23"
         const val retrofit = "2.9.0"
         const val androidXTest = "1.3.0"
-        const val junit5 = "5.7.1"
-        const val junit5Platform = "1.6.0"
+        const val junit5 = "5.7.2"
+        const val junit5Platform = "1.7.2"
```
eugene-krivobokov, line 16: > platform version was out of sync with junit5

I want to keep them aligned automatically.
We could use [bom](https://repo1.maven.org/maven2/org/junit/junit-bom/5.7.2/junit-bom-5.7.2.pom) spec.
I've added this case to MBS-10416.

eugene-krivobokov (approved):

PR #1045, opened 2021-06-04
MBS-11279 remove spamming "skip sending event" message

dsvoronin:


subprojects/common/statsd/src/main/kotlin/com/avito/android/stats/StatsDSender.kt
```diff
@@ -83,8 +83,6 @@ interface StatsDSender {
 
             if (config is StatsDConfig.Enabled) {
                 logger.debug("${metric.type}:${config.namespace}.$aspect:${metric.value}")
-            } else {
-                logger.debug("Skip sending event: ${metric.type}:<namespace>.$aspect:${metric.value}")
```
eugene-krivobokov, line 87: It's used in tests. Need to provide an alternative or switch them to messages in enabled state.

dsvoronin: ok, i didn't know that skipped events being asserted too.
Well, i'll try to change assertion mechanism then, as we discussed: not by logs, but by network calls to stub api

dsvoronin: it's not so straightforward task, statsd uses datagram sockets
will dive into this problem later

PR #1048, opened 2021-06-05
MBS-11322 Make test lost when device die when executes

sergioboi:


subprojects/test-runner/service/src/main/kotlin/com/avito/runner/service/worker/DeviceWorker.kt
```diff
@@ -83,7 +83,7 @@ internal class DeviceWorker(
                             result = result
                         )
                     }.onFailure { failure ->
-                        onDeviceDieOnIntention(intention, failure)
+                        onDeviceDieWhenActionExecuted(intention, failure)
```
sergioboi: Before: If we have a bug on execution logic then the device will die and test re-executes. So it leads to infinite re-execution
Now: If we have a failure at the execution phase then we will kill a device and make test execution lost

sergioboi: In the future, we should make device dying more accurate

dsvoronin (approved):

subprojects/test-runner/service/src/main/kotlin/com/avito/runner/service/worker/DeviceWorker.kt
```diff
@@ -197,6 +201,32 @@ internal class DeviceWorker(
...
+    private suspend fun onDeviceDieWhenExecutingTest(
+        intention: Intention,
+        reason: Throwable
+    ) {
+        deviceListener.onDeviceDied(
+            device = device,
+            message = "DeviceWorker died when executed intention: $intention",
+            reason = reason
+        )
+        deviceListener.onTestCompleted(
```
eugene-krivobokov, line 213: Does it mean that in case of emulator issues test won't retry?

sergioboi: Yes this attempt will be LOST

PR #904, opened 2021-04-02
Update to gradle 7

dsvoronin:
https://docs.gradle.org/7.0/release-notes.html

## Benchmark

see `performance.scenarios` for options

| gradle version  | fastCheck median |
| ------------- | ------------- |
| 6.9 | 24,621.41 ms |
| 7.0.2  | 24,955.83 ms |

## TODO

- [X] check kotlin daemon params gradle/gradle#17075 potential blocker; should check before upgrade (FIXED in 7.2) Only blocks avito update
- [x] https://github.com/avito-tech/avito-android/pull/1050 blocker

PR #1052, opened 2021-06-07
MBS-11395 some gradle7 preparations

dsvoronin:
extracted from https://github.com/avito-tech/avito-android/pull/904

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,6 @@ gh-pages-generated/
 *.iml
 .idea/*
 !.idea/dictionaries/
+# gradle-profiler
+gradle-user-home
+profile-out*
--- a/Makefile
+++ b/Makefile
@@ -259,3 +259,6 @@ check_avito_configuration:
 
 dependency_updates:
 	$(docker_command) ./gradlew -p $(project) $(log_level) $(params) dependencyUpdates -Drevision=release
+
+benchmark_fast_check:
+	gradle-profiler --benchmark --project-dir subprojects --scenario-file gradle/performance.scenarios fastCheck
--- a/gradle/performance.scenarios
+++ b/gradle/performance.scenarios
@@ -1,0 +1,7 @@
+fastCheck {
+    title = "fastCheck"
+    tasks = ["compileAll", "detektAll", "test"]
+    cleanup-tasks = ["clean"]
+    gradle-args = ["--no-build-cache", "--parallel"]
+    daemon = warm
+}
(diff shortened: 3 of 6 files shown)


Makefile
```diff
@@ -259,3 +259,6 @@ check_avito_configuration:
 
 dependency_updates:
 	$(docker_command) ./gradlew -p $(project) $(log_level) $(params) dependencyUpdates -Drevision=release
+
+benchmark_fast_check:
+	gradle-profiler --benchmark --project-dir subprojects --scenario-file gradle/performance.scenarios fastCheck
```
eugene-krivobokov, line 264:
```suggestion
	gradle-profiler --benchmark --project-dir subprojects --gradle-user-home ~/.gradle-user-home --output-dir build/reports/gradle-profiler  --scenario-file gradle/performance.scenarios fastCheck
```

- `gradle-user-home` helps to reuse profiler between different builds
- `build/` output because it's well known and common

.gitignore
```diff
@@ -11,3 +11,6 @@ gh-pages-generated/
 *.iml
 .idea/*
 !.idea/dictionaries/
+# gradle-profiler
+gradle-user-home
```
eugene-krivobokov, line 15: It'd be better to avoid this issue but not ignore
https://github.com/avito-tech/avito-android/pull/1052/files#r646526650

eugene-krivobokov (approved):

Makefile
dsvoronin: outputs redirect has one downside: previously it created a new directory for every new run, and after redirection it just overwrites existing results

dsvoronin: gradle-user-home should probably be project specific, i'am not sure though

eugene-krivobokov: > and after redirection it just overwrites existing results

Why do you need multiple results? It's hard to compare independent runs.
To compare two implementations we need to run two scenarios in one run.

eugene-krivobokov: > gradle-user-home should probably be project specific

Didn't see any problems sharing it. It's like gradle GRADLE_USER_HOME

dsvoronin: > Didn't see any problems sharing it. It's like gradle GRADLE_USER_HOME

from doc:
```
--gradle-user-home: The Gradle user home. Defaults to <project-dir>/gradle-user-home to isolate performance tests from your other builds.
```

maybe i misinterpreted this comment, though is is a good idea to keep this home dir isolated from other projects

dsvoronin: > Why do you need multiple results? It's hard to compare independent runs.
> To compare two implementations we need to run two scenarios in one run.

Just running multiple checks locally, modifying parameters
I keep forgetting to clear folder before new benchmark starts

eugene-krivobokov: > Just running multiple checks locally, modifying parameters

Sorry, still didn't get why do we need multiple runs\results. 
For now, I see only one reason to rerun - to compare different implementations \ scenarios.

dsvoronin: >  For now, I see only one reason to rerun - to compare different implementations \ scenarios.

yes, exactly what i'm doing

eugene-krivobokov: When we have two scenarios in one gradle-profiler's report we have built-in math:
Screenshot from 2021-06-07 20-27-20

`97.52` - is a confidence calculated by Mann–Whitney U-test.

When we have multiple independent gradle-profiler results, we need to calculate the same by hand. Otherwise, it will be incorrect to compare two distributions. This is why I think it's worth overriding results, to not encourage manual comparison.

PR #1056, opened 2021-06-08
MBS-11263 run samples tests on CI

dsvoronin:
Added separate teamcity configuration, that will break PR's with sample instrumentation tests problems

### DIFF
--- a/ci/_environment.sh
+++ b/ci/_environment.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-ANDROID_BUILDER_TAG=28e6bacd68
+ANDROID_BUILDER_TAG=eddf45adfa
 
 if [[ -z "${DOCKER_REGISTRY+x}" ]]; then
     # using dockerhub for public availability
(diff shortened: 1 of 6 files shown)


ci/_environment.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-ANDROID_BUILDER_TAG=28e6bacd68
+ANDROID_BUILDER_TAG=eddf45adfa
```
eugene-krivobokov, line 3: What has been changed?

dsvoronin: https://github.com/avito-tech/avito-android/pull/813
it is changes from that pr

sergioboi:
why do we need a separate teamcity configuration?

dsvoronin:
> why do we need a separate teamcity configuration?

it is temporary; i'am not sure that sample project is stable enough to be a PR blocker right now, wanted to look at it for ~ a week

eugene-krivobokov (approved):

PR #1059, opened 2021-06-09
Mbs 11392 runner cleanup

dsvoronin:
remove `shared-test` module
make stubs instead of generator functions
move stuff to testFixtures


subprojects/test-runner/client/src/test/kotlin/com/avito/test/ListWithDefault.kt
```diff
@@ -0,0 +1,5 @@
+package com.avito.test
+
+internal fun <T : Any> listWithDefault(size: Int, value: T): List<T> {
+    return (0 until size).map { value }
```
eugene-krivobokov, line 4:
```suggestion
    return List(size) { value }
```

subprojects/test-runner/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/service/TestRunnerWorkAction.kt
```diff
@@ -38,7 +40,13 @@ public abstract class TestRunnerWorkAction : WorkAction<TestRunnerWorkAction.Par
 
         val testResults = parameters.service.get().runTests(params, legacyTestRunParams)
 
-        factory.provideFinalizer().finalize(testResults)
+        val buildFailer: BuildFailer = BuildFailer.RealFailer()
+
+        when (val result = factory.provideFinalizer().finalize(testResults)) {
+            Finalizer.Result.Ok -> {
+            }
```
eugene-krivobokov, line 47:
```suggestion
                // no op
            }
```

subprojects/test-runner/service/build.gradle.kts
```diff
@@ -32,7 +32,7 @@ dependencies {
     testImplementation(project(":common:truth-extensions"))
     testImplementation(project(":common:resources"))
     testImplementation(project(":gradle:test-project"))
-    testImplementation(project(":test-runner:shared-test"))
+    testImplementation(project(":test-runner:stub"))
```
eugene-krivobokov, line 35: I didn't get why we use a standalone module for stubs.
Do we use it only in tests?
If so, can't we share it by `testFixtures`?

eugene-krivobokov (approved):

subprojects/test-runner/service/build.gradle.kts
dsvoronin: we use it in production code, i'll try to move to `device-provider`

subprojects/test-runner/client/src/test/kotlin/com/avito/test/ListWithDefault.kt
dsvoronin: done in next PR, removed this extension completely in favour of stdlib fun

subprojects/test-runner/service/build.gradle.kts
dsvoronin: done in next PR

PR #1068, opened 2021-06-15
MBS-11446 Expose signer timeout setting

dsvoronin:
- expose signer http client 'readWriteTimeoutSec' to plugin extension
- url property instead of host

### DIFF
--- a/subprojects/gradle/signer/src/gradleTest/kotlin/com/avito/plugin/SignServicePluginTest.kt
+++ b/subprojects/gradle/signer/src/gradleTest/kotlin/com/avito/plugin/SignServicePluginTest.kt
@@ -40,7 +40,6 @@ class SignServicePluginTest {
         generateTestProject(
             signServiceExtension = """
              signService {
-                // host
                 apk(android.buildTypes.release, "signToken")
                 bundle(android.buildTypes.release, "signToken")
              }
@@ -56,7 +55,31 @@ class SignServicePluginTest {
 
         result.assertThat()
             .buildFailed()
-            .outputContains("host")
+            .outputContains("Invalid signer url value: ''")
+    }
+
+    @Test
+    fun `plugin apply - fails - configuration with invalid url`() {
+        generateTestProject(
+            signServiceExtension = """
+             signService {
+                url = "some_incorrect_url"
+                apk(android.buildTypes.release, "signToken")
+                bundle(android.buildTypes.release, "signToken")
+             }
+        """.trimIndent()
+        )
+
+        val result = ciRun(
+            testProjectDir,
+            ":app:signApkViaServiceRelease",
+            dryRun = true,
+            expectFailure = true
+        )
+
+        result.assertThat()
+            .buildFailed()
+            .outputContains("Invalid signer url value: 'some_incorrect_url'")
     }
 
     @Test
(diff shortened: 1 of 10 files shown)


subprojects/gradle/signer/src/gradleTest/kotlin/com/avito/plugin/SignServicePluginTest.kt
```diff
@@ -56,7 +55,31 @@ class SignServicePluginTest {
 
         result.assertThat()
             .buildFailed()
-            .outputContains("host")
+            .outputContains("Invalid signer url value: ''")
+    }
+
+    @Test
+    fun `plugin apply - fails - configuration with invalid`() {
```
eugene-krivobokov:
```suggestion
    fun `plugin apply - fails - configuration with invalid url`() {
```

subprojects/gradle/signer/src/main/kotlin/com/avito/plugin/SignArtifactTask.kt
```diff
@@ -8,22 +8,26 @@ import com.avito.logger.create
...
+    @get:Input
+    abstract val tokenProperty: Property<String>
 
-    @Input
-    val serviceUrl = objects.property<String>()
+    @get:Input
+    abstract val serviceUrl: Property<String>
+
+    @get:Input
+    abstract val readWriteTimeoutSec: Property<Long>
```
eugene-krivobokov, line 30: It is an ad-hoc parameter but may be more reusable.
We could customize HTTP clients for each external service. And in all cases, a config may be the same.

The idea in properties:

```
avito.service.sign.http.readTimeoutSec=40
avito.service.sign.http.writeTimeoutSec=10

avito.service.another.http.readTimeoutSec=30
avito.service.another.http.writeTimeoutSec=10
```

Thus, I can override HTTP parameters for any service. Naming is a pattern that is supported in common HTTP models.

Why properties? Because we didn't agree yet to migrate to extensions in MBS-11039, it is only a hypothesis.

subprojects/gradle/signer/src/gradleTest/kotlin/com/avito/plugin/SignServicePluginTest.kt
```diff
@@ -56,7 +55,31 @@ class SignServicePluginTest {
 
         result.assertThat()
             .buildFailed()
-            .outputContains("host")
+            .outputContains("Invalid signer url value: ''")
+    }
+
+    @Test
+    fun `plugin apply - fails - configuration with invalid url`() {
```
eugene-krivobokov, line 62: `plugin apply - fails`, but we schedule a task: `:app:signApkViaServiceRelease --dry-run` 

It doesn't look like a project configuration. Is it correct?

eugene-krivobokov (approved):

subprojects/gradle/signer/src/gradleTest/kotlin/com/avito/plugin/SignServicePluginTest.kt
dsvoronin: it is, dry run will not run any actual tasks, but goes through configuration phase
i need this combination here, because there is no other way to trigger task configuration without it's run

subprojects/gradle/signer/src/main/kotlin/com/avito/plugin/SignArtifactTask.kt
dsvoronin: i really don't like these implicit parameters, which is only known from documentation

eugene-krivobokov: I understand the benefits of explicit config but don't know cases when we really need unique parameters for the same external service in different plugins.

dsvoronin: this `readWriteTimeoutSec` is bad example, 
MBS-11039 proposal with reusable nested extension blocks should be fine though!

PR #1069, opened 2021-06-16
Test runner creation refactoring

sergioboi:


subprojects/test-runner/client/src/main/kotlin/com/avito/runner/scheduler/TestRunnerFactory.kt
```diff
@@ -15,35 +11,30 @@ import com.avito.runner.scheduler.runner.TestRunnerExecutionState
...
-            prefix = config.metricsConfig.runnerPrefix
-        ),
-        timeProvider = timeProvider,
-        loggerFactory = config.loggerFactory
-    )
+
+    /**
+     * i.e {projectDir}/output/test-runner/
+     */
+    private val testRunnerOutputDir: File by lazy {
```
dsvoronin, line 30: Please take a look at: https://github.com/avito-tech/avito-android/blob/68d514f0eb4e56c244946151d2f41716c4162e8d/subprojects/test-runner/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/InstrumentationTestsPlugin.kt#L129-L132
Does it clashes somehow? Maybe move more logic from gradle plugin there?
On plugin side it's not clear what outputDir user should specify

dsvoronin (approved):

subprojects/test-runner/client/src/main/kotlin/com/avito/runner/scheduler/TestRunnerFactory.kt
sergioboi: As I understand we have root output-dir and some sub-dirs. In that case, Factory need a `root output dir` to create a sub-dir.
So in the plugin, we could rename the property as `pluginRootOutputDir`

PR #1070, opened 2021-06-16
MBS-11288 fix inmemory report unexpected LOST cases

dsvoronin:
Problem was that AndroidTest.Lost type was treated like actual Lost(not reported) tests, but in fact we treat it like InfrastructureError, and suppressFailures should work for it

added tests for this cases and also more integration tests on new Finalizer

### DIFF
--- a/subprojects/test-runner/client/src/test/kotlin/com/avito/runner/finalizer/FinalizerImplTest.kt
+++ b/subprojects/test-runner/client/src/test/kotlin/com/avito/runner/finalizer/FinalizerImplTest.kt
@@ -1,0 +1,91 @@
+package com.avito.runner.finalizer
+
+import com.avito.report.NoOpReportLinkGenerator
+import com.avito.report.model.AndroidTest
+import com.avito.report.model.Incident
+import com.avito.report.model.TestRuntimeDataPackage
+import com.avito.report.model.TestStaticDataPackage
+import com.avito.report.model.createStubInstance
+import com.avito.runner.finalizer.action.WriteTaskVerdictAction
+import com.avito.runner.scheduler.runner.model.TestSchedulerResult
+import com.avito.truth.isInstanceOf
+import com.google.common.truth.Truth.assertThat
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.io.TempDir
+import java.io.File
+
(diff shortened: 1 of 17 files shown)


subprojects/test-runner/client/src/testFixtures/kotlin/com/avito/runner/finalizer/StubFinalizerImpl.kt
```diff
@@ -0,0 +1,21 @@
...
+
+import com.avito.runner.finalizer.action.FinalizeAction
+import com.avito.runner.finalizer.verdict.VerdictDeterminer
+import com.avito.runner.finalizer.verdict.VerdictDeterminerImpl
+import com.avito.runner.finalizer.verdict.createStubInstance
+import java.io.File
+
+internal fun FinalizerImpl.Companion.createStubInstance(
+    actions: List<FinalizeAction> = emptyList(),
+    verdictFile: File = File("."),
```
eugene-krivobokov, line 11: Shouldn't we pass a temporary file?

subprojects/test-runner/client/src/test/kotlin/com/avito/runner/finalizer/FinalizerImplTest.kt
```diff
@@ -0,0 +1,91 @@
...
+        )
+
+        val result = finalizerImpl.finalize(testSchedulerResult)
+
+        assertThat(result).isInstanceOf<Finalizer.Result.Failure>()
+        assertThat(verdictFile.readText()).contains("${test.name} ${test.device} NOT REPORTED")
+    }
+
+    @Test
+    fun `finalized - ok - single test failure`(@TempDir output: File) {
```
eugene-krivobokov, line 55:
```suggestion
    fun `finalized - failure - single test failure`(@TempDir output: File) {
```

eugene-krivobokov, line 55:
```suggestion
    fun `finalized - ok - single test with incident`(@TempDir output: File) {
```
?
I'm not sure if "test failure" describes the case.

eugene-krivobokov (approved):

sergioboi (approved):

PR #1073, opened 2021-06-17
MBS-11457 signer plugin enhancements

dsvoronin:
- add documentation
- describe with Problem's
- use Worker API
- fix missing http errors data


RuslanMingaliev (approved):

subprojects/gradle/signer/src/main/kotlin/com/avito/plugin/StringExt.kt
```diff
@@ -0,0 +1,15 @@
+package com.avito.plugin
+
+import kotlin.contracts.ExperimentalContracts
+import kotlin.contracts.contract
+
+@OptIn(ExperimentalContracts::class)
+internal fun String?.hasContent(): Boolean {
```
sergioboi, line 7: "I have no idea what I watching"

dsvoronin: it's about kotlin contracts api?

PR #1076, opened 2021-06-18
MBS-11461 hide finalizer in test-runner:client

dsvoronin:
remove InstrumentationTestsAction
rename InstrumentationTestsActionParams -> RunnerInputParams


eugene-krivobokov (approved):

subprojects/test-runner/client/src/main/kotlin/com/avito/runner/scheduler/runner/scheduler/TestSchedulerFactoryImpl.kt
```diff
@@ -1,47 +1,229 @@
...
+            metricsConfig = metricsConfig,
+            reportFactory = reportFactory,
+            timeProvider = timeProvider
+        )
+
+        val androidDebugBridgeProvider = createAndroidDebugBridgeProvider(params.loggerFactory)
+
+        val emulatorsLogsReporterProvider = createEmulatorsLogsReporterProvider(params)
+
+        val testRunnerFactoryProvider = createTestRunnerFactoryProvider(
```
sergioboi: I think that is not what `TestShecdulerFactory` should do
`TestSchedulerFactory` must only create the `TestScheduler` and get `testRunnerFactoryProvider` as an input parameter

subprojects/test-runner/client/src/main/kotlin/com/avito/runner/scheduler/runner/scheduler/TestSchedulerFactoryImpl.kt
```diff
@@ -1,47 +1,229 @@
...
+            finalizer = finalizerFactory.create(),
         )
     }
 
-    private fun createTestSuiteProvider(): TestSuiteProvider = TestSuiteProvider.Impl(
-        report = report,
-        targets = params.instrumentationConfiguration.targets,
-        reportSkippedTests = params.instrumentationConfiguration.reportSkippedTests,
-        filterFactory = FilterFactory.create(
+    private fun createFinalizerFactory(
```
sergioboi: `TestShedulerFactory` must not know how to create a Finalizer
Make FinalizerFactory as an input parameter

subprojects/test-runner/client/src/main/kotlin/com/avito/runner/scheduler/runner/scheduler/TestSchedulerFactoryImpl.kt
```diff
@@ -1,47 +1,229 @@
...
+    }
+
+    private fun createStatsdSender(params: RunnerInputParams): StatsDSender {
+        return StatsDSender.Impl(
+            config = params.statsDConfig,
+            loggerFactory = params.loggerFactory
+        )
+    }
+
+    private fun createTestSuiteLoader(): TestSuiteLoader {
```
sergioboi: Useless

sergioboi (changes requested):

subprojects/test-runner/client/src/main/kotlin/com/avito/runner/scheduler/runner/scheduler/TestSchedulerFactoryImpl.kt
dsvoronin: Finalizer is now hidden in TestScheduler implementation, so i can only extract it like this:

```
TestSchedulerFactoryImpl(finalizerFactory (params) -> Filnalizer) {

   override fun create(params: RunnerInputParams): TestScheduler {
        val finalizerFactory = finalizerFactory.invoke(params)
   }
}
```

which is odd

subprojects/test-runner/client/src/main/kotlin/com/avito/runner/scheduler/runner/scheduler/TestSchedulerFactoryImpl.kt
dsvoronin: hides implementation, i had plans to look why Implementation class is here, and not hidden in some module

subprojects/test-runner/client/src/main/kotlin/com/avito/runner/scheduler/runner/scheduler/TestSchedulerFactoryImpl.kt
dsvoronin: so we need something like TestSchedulerFactoryFactory, or entrypoint for runner?

subprojects/test-runner/client/src/main/kotlin/com/avito/runner/scheduler/runner/scheduler/TestSchedulerFactoryImpl.kt
sergioboi: I don't want to debate lets discuss it at the call

sergioboi (approved):

PR #1078, opened 2021-06-18
MBS-10600 setup gradle cache node

dsvoronin:
add usage and setup docs


ci/docker/gradle-cache-node-github/image.yaml
```diff
@@ -0,0 +1,2 @@
+registry: DOCKER_REGISTRY
+image: android/gradle-cache-node-github
```
eugene-krivobokov: I'd want to avoid a specific image for an application. It's generic and versionable, but not related to the app.

subprojects/settings.gradle.kts
```diff
@@ -301,6 +301,61 @@ dependencyResolutionManagement {
...
+    return uri(url)
+}
+
+fun isRemoteCachePushEnabled(): Boolean {
+    val result = avitoGithubRemoteCachePush?.toBoolean() ?: false
+    return if (result) {
+        logger.lifecycle("[RemoteCache] push enabled")
+        true
+    } else {
+        logger.lifecycle("[RemoteCache] push disabled")
```
eugene-krivobokov: Let's remove diagnostic messages. They duplicate what you can easily find in the build scan.

subprojects/settings.gradle.kts
```diff
@@ -301,6 +301,61 @@ dependencyResolutionManagement {
...
+        url = remoteCacheUrl()
+        isEnabled = hasRemoteCacheUrl()
+        isPush = isRemoteCachePushEnabled()
+        isAllowUntrustedServer = true
+        isAllowInsecureProtocol = true
+    }
+}
+
+fun remoteCacheUrl(): java.net.URI {
+    val url = "http://${avitoGithubRemoteCacheHost.orNull}/cache/"
```
eugene-krivobokov: I'd like to use more composable structure for properties, to make them more self-describable:

`avitoGithub.gradle.buildCache.remote.push`

docs/content/contributing/internal/RemoteCache.md
```diff
@@ -0,0 +1,42 @@
...
+    Persistence volume not set for now, deleting a pod will lead to losing cache.
+
+Official docs: https://docs.gradle.com/build-cache-node
+
+## How to enable locally
+
+???+ Prerequisites 
+    Ask for <GRADLE_CACHE_NODE_HOST> in internal chat
+    
+Set environment variable `GRADLE_CACHE_NODE_HOST`
```
eugene-krivobokov, line 16: Why not project property? It's used only in Gradle process and can be hidden from others.

docs/content/contributing/internal/RemoteCache.md
```diff
@@ -0,0 +1,42 @@
...
+
+## How to deploy
+
+???+ Prerequisites
+    - DOCKER_REGISTRY env set
+    - GRADLE_CACHE_NODE_HOST env set
+    - [kubernetes credentials configured](http://links.k.avito.ru/kubectl)
+    - Write access to `gradle-remote-cache` namespace in cluster
```
eugene-krivobokov: Is it feasible?
See comments to MBS-5444.
The solution should be accessible for everyone.

Makefile
```diff
@@ -270,3 +271,40 @@ dependency_updates:
...
+	cd ./ci/k8s/gradle-remote-cache && \
+	sed -e 's|GRADLE_CACHE_NODE_HOST|$(GRADLE_CACHE_NODE_HOST)|g' -e 's|NODE_IMAGE|$(DOCKER_REGISTRY)/android/gradle-cache-node-github:$(GRADLE_CACHE_NODE_TAG)|g' github-project.yaml | kubectl apply -f - && \
+	echo "Gradle Cache Node web interface should be available soon here: http://$(GRADLE_CACHE_NODE_HOST)"
+
+delete_gradle_cache_node:
+	cd ./ci/k8s/gradle-remote-cache && \
+	kubectl delete -f github-project.yaml
+
+check-host-env:
+ifndef GRADLE_CACHE_NODE_HOST
```
eugene-krivobokov: We have also an imperative check - `check_defined`.
I'd like to use only one way to do something or have explanations for both if they are not identical.

ci/gradle.properties
```diff
@@ -0,0 +1,2 @@
+org.gradle.jvmargs=-Xmx16g -XX:+UseGCOverheadLimit -XX:GCTimeLimit=10 -Dfile.encoding=UTF-8
```
eugene-krivobokov, line 1: Do we really need 16G?

docs/content/contributing/internal/RemoteCache.md
eugene-krivobokov: I see, it's also used in deploy

ci/gradle.properties
dsvoronin: I'll do some testing.
An idea for this change was to specify CI xmx separately from local, and that we have more than enough ram per build agent
And ci builds consume a lot of ram because of massive GradleTestKit runs in parallel

Makefile
dsvoronin: Thanks, fixed
already forgot about this func

image

docs/content/contributing/internal/RemoteCache.md
dsvoronin: mentioned this issue in doc

subprojects/settings.gradle.kts
dsvoronin: i agree, the problem is i wanted to use delegation property syntax `someVar by settings`, but it doesn't work with `.` in name
Not so much profit in this mechanism though, will change

subprojects/settings.gradle.kts
dsvoronin: removed, was thinking to left it until i see that all is working ok

ci/docker/gradle-cache-node-github/image.yaml
dsvoronin: i want it too, but configuration is still a part of an image and i can't find another method to specify it
Tags are also taken by out "hash" publication mechanism, so only image name left

eugene-krivobokov: > but configuration is still a part of an image

Do you mean `config.yaml`? Can't we mount it?

dsvoronin: yes, config.yaml
It applies on container start

dsvoronin: i don't mind two images, honestly

eugene-krivobokov: Still don't understand why use two different images if all the difference can be customized in `docker run` command.

eugene-krivobokov: Also, why official image `gradle/build-cache-node` didn't fit the purpose?

dsvoronin: image

phew, finally executed some intiContainer-shell kung-fu to make it work

eugene-krivobokov: I think such implementation with printf is not worth it. I was hoping to apply a config from file like in docker run.

eugene-krivobokov (approved):

ci/docker/gradle-cache-node-github/image.yaml
dsvoronin: how so? we should either download file in init-container or place it somewhere in cluster to mount.
But persistent mount volumes is not enabled right now.

We need PVC for cache node though, it enables us to do updates without losing cache.
Let's discuss it on another sync

I think it's ok to left only UI configuratble settings (it's like 2 values really) if we'll have PVC, but for now i see `printf` as a valuable compromise (easy installation, easy enough configuration)

eugene-krivobokov: > how so? we should either download file in init-container or place it somewhere in cluster to mount.

I also don't know how to do it. I was just asking about implementation and didn't want to force any implementation. The goal is to make it easy to support.

PR #1083, opened 2021-06-21
MBS-7663 upgrade avito cache node + instructions

dsvoronin:


sergioboi (approved):

eugene-krivobokov:
Will you add a CI configuration for cache warmup?
See example in `Gradle remote build cache warmup`.
It might be crucial for local builds.

dsvoronin:
> Will you add a CI configuration for cache warmup?
> See example in `Gradle remote build cache warmup`.
> It might be crucial for local builds.

Looks like we run all set of tasks in CI builds; let's take a look at local scans to know what is missing in cache

PR #1085, opened 2021-06-21
MBS-11456 agp 4.2 preparations: support signer plugin

dsvoronin:
hack around https://issuetracker.google.com/issues/174678813

### DIFF
--- a/subprojects/signer/src/main/kotlin/com/avito/plugin/SignServicePlugin.kt
+++ b/subprojects/signer/src/main/kotlin/com/avito/plugin/SignServicePlugin.kt
@@ -2,7 +2,6 @@ package com.avito.plugin
 
 import com.android.build.api.artifact.ArtifactType
 import com.android.build.api.variant.Variant
-import com.android.build.gradle.api.ApplicationVariant
 import com.avito.android.Problem
 import com.avito.android.androidCommonExtension
 import com.avito.android.asRuntimeException
@@ -11,6 +10,7 @@ import com.avito.android.withAndroidApp
 import org.gradle.api.Plugin
 import org.gradle.api.Project
 import org.gradle.api.tasks.TaskContainer
+import org.gradle.api.tasks.TaskProvider
 import org.gradle.kotlin.dsl.create
 import org.gradle.kotlin.dsl.register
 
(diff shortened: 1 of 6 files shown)


eugene-krivobokov (approved):

subprojects/signer/src/main/kotlin/com/avito/plugin/SignServicePlugin.kt
```diff
@@ -21,71 +21,89 @@ public class SignServicePlugin : Plugin<Project> {
...
+            appExtension.applicationVariants.all { variant: com.android.build.gradle.api.ApplicationVariant ->
+
+                val buildTypeName = variant.buildType.name
+                val apkToken: String? = extension.apkSignTokens[buildTypeName]
+                val bundleToken: String? = extension.bundleSignTokens[buildTypeName]
+
+                variant.outputsAreSigned = apkToken.hasContent() || bundleToken.hasContent()
+
+                target.tasks.signedApkTaskProvider(variant.name).configure { signApkTask ->
+                    signApkTask.dependsOn(variant.packageApplicationProvider)
```
jdochez, line 39: you should not have to do this. Since you are using the new variant api to wire the task input to the APK, it will be automatically a dependent.

subprojects/signer/src/main/kotlin/com/avito/plugin/SignServicePlugin.kt
```diff
@@ -21,71 +21,89 @@ public class SignServicePlugin : Plugin<Project> {
...
             target.androidCommonExtension.onVariantProperties {
+                val variant = this
 
-                artifacts.use(target.tasks.signedApkTaskProvider(this))
-                    .wiredWithDirectories(
-                        taskInput = SignApkTask::unsignedDirProperty,
-                        taskOutput = SignApkTask::signedDirProperty
-                    )
-                    .toTransform(ArtifactType.APK)
-
```
jdochez, line 66: here, you should be able to do :
artifacts.use(target.tasks.signedApkTaskProvider(this)).configure {
                    signedDirProperty.set(File("/path/to/where/you/want/your/signed/files"))
                }

same for bundle

subprojects/signer/src/main/kotlin/com/avito/plugin/SignServicePlugin.kt
dsvoronin: Yes, you're right, removed this manual wiring;

What about `variant.outputsAreSigned`? 

It's not deprecated, but look like it should be somehow reflected in new API's, right?

How to say properly to AGP(post 7.0) that i want to handle signing myself?

subprojects/signer/src/main/kotlin/com/avito/plugin/SignServicePlugin.kt
dsvoronin: `artifacts.use` returns `TaskBasedOperation` which don't have `configure`; but probably you meant that i should configure our custom `signedApkTaskProvider`?

We overwrite final signed file in standard (AGP provided) location, i don't really know why so and can change this logic to output in custom directory and then wire signed file consumers accordingly.

@eugene-krivobokov maybe you remember why we need to override?

eugene-krivobokov: No, I don't remember. 
I assume only for convenience, to have singed artifact in one place, regardless of this plugin.
When no other plugins are consuming this artifact it doesn't look important.

jdochez: yes my example was based on using the TaskProvider for your signing task.

PR #1089, opened 2021-06-22
MBS-11456 don't wire signer tasks manually

dsvoronin:


eugene-krivobokov (approved):

dsvoronin:
Looks like MBS-11484 (Optinal signer plugin in CiSteps) should be done right here

PR #1105, opened 2021-06-28
MBS-9992 Add a flag to disable intents stubbing

RuslanMingaliev:


subprojects/test-runner/test-inhouse-runner/src/main/kotlin/com/avito/android/rule/InHouseScenarioScreenRule.kt
```diff
@@ -16,13 +16,16 @@ import org.junit.rules.TestRule
 import org.junit.runner.Description
 import org.junit.runners.model.Statement
 
-abstract class InHouseScenarioScreenRule<A : Activity>(activityClass: Class<A>) : TestRule {
+abstract class InHouseScenarioScreenRule<A : Activity>(
+    activityClass: Class<A>,
+    stubIntents: Boolean = true
```
RuslanMingaliev: Current behaviour

subprojects/test-runner/test-inhouse-runner/src/main/kotlin/com/avito/android/rule/ActivityScenarioRule.kt
```diff
@@ -9,15 +9,20 @@ import com.avito.android.rule.base.BaseActivityScenarioRule
...
     initialTouchMode = true,
     launchActivity = false,
 ) {
 
     override fun afterActivityLaunched() {
         Intents.init()
-        com.avito.android.test.Intents.stubEverything()
+        if (stubIntents) {
+            com.avito.android.test.Intents.stubEverything()
+        }
```
eugene-krivobokov, line 25: Should we consider this flag in `afterActivityFinished` for  `Intents.release()` ?

subprojects/test-runner/test-inhouse-runner/src/main/kotlin/com/avito/android/rule/InHouseScenarioScreenRule.kt
eugene-krivobokov: Will we make it disabled by default in the next release? What are your plans?

eugene-krivobokov: I'm asking because this new flag will help to disable stubs but won't help to diagnose the problem.
The question is about making a less error-prone API.

eugene-krivobokov (approved):

subprojects/test-runner/test-inhouse-runner/src/main/kotlin/com/avito/android/rule/ActivityScenarioRule.kt
```diff
@@ -9,15 +9,20 @@ import com.avito.android.rule.base.BaseActivityScenarioRule
  * actions that send out intents which need to be verified or stubbed. This is similar to
  * MockitoAnnotations.initMocks.
  */
-class ActivityScenarioRule<T : Activity>(activityClass: Class<T>) : BaseActivityScenarioRule<T>(
+class ActivityScenarioRule<T : Activity>(
+    activityClass: Class<T>,
+    private val stubIntents: Boolean
```
sergioboi, line 14: Do we need a default value there?

subprojects/test-runner/test-inhouse-runner/src/main/kotlin/com/avito/android/rule/ActivityScenarioRule.kt
```diff
@@ -9,15 +9,20 @@ import com.avito.android.rule.base.BaseActivityScenarioRule
...
     activityClass = activityClass,
     initialTouchMode = true,
     launchActivity = false,
 ) {
 
     override fun afterActivityLaunched() {
         Intents.init()
-        com.avito.android.test.Intents.stubEverything()
+        if (stubIntents) {
+            com.avito.android.test.Intents.stubEverything()
```
sergioboi, line 24: Could we write a stubber that will write logs with what it stubbed? It will help If someone use stub but forget about that behaviour

subprojects/test-runner/test-inhouse-runner/src/main/kotlin/com/avito/android/rule/ActivityScenarioRule.kt
RuslanMingaliev: `Intents.release()` must be called after `Intents.init()`. We can make intents initialization optional at all, but it will break existing tests.

subprojects/test-runner/test-inhouse-runner/src/main/kotlin/com/avito/android/rule/InHouseScenarioScreenRule.kt
RuslanMingaliev: I was afraid if I turned it off by default, I'd break the tests. I'll check the integration against Avito

subprojects/test-runner/test-inhouse-runner/src/main/kotlin/com/avito/android/rule/ActivityScenarioRule.kt
RuslanMingaliev: I think no, we don't. I've added `internal` modifier to this class. It's used only in one place.

subprojects/test-runner/test-inhouse-runner/src/main/kotlin/com/avito/android/rule/InHouseScenarioScreenRule.kt
eugene-krivobokov: > if I turned it off by default, I'd break the tests

Most probably, but this is not a blocker to change behavior. It can be done in multiple steps.
The question is not about changes in this PR but rather about the final goal.

RuslanMingaliev: Integration build succeeded. Changed the flag value to `false`

subprojects/test-runner/test-inhouse-runner/src/main/kotlin/com/avito/android/rule/ActivityScenarioRule.kt
RuslanMingaliev: I disabled the stubbing by default. Now it should be more clear for end-users. I agree that logs will be very useful here, we can add them later.

subprojects/test-runner/test-inhouse-runner/src/main/kotlin/com/avito/android/rule/InHouseScenarioScreenRule.kt
RuslanMingaliev: I was wrong, the build failed.

subprojects/test-runner/test-inhouse-runner/src/main/kotlin/com/avito/android/rule/ActivityScenarioRule.kt
RuslanMingaliev: Added logs

PR #1106, opened 2021-06-28
MBS-9712 cross app test summary and tms plugin work

dsvoronin:
removed tms plugin, tms step now do all the work

actual tms/summary work done in single task root project per planSlug-jobSlug

**NEED TO RE-CONFIGURE AFTER UPGRADE**

- `com.avito.android.test-summary` now should be applied to root project only
- `TestSummaryExtension` should be configured on root project 
- old hack should be removed
- all demo-apps should register all 3 ci steps: tms/flaky/summary


eugene-krivobokov:
> CONFIGURATION IS NOT BACKWARD COMPATIBLE

Proposal: keep empty implementation to preserve only compilation.
Ideally, we shouldn't check changes against the internal repository, but for now, it's risky to lose this feedback completely.

subprojects/gradle/test-summary/src/main/kotlin/com/avito/test/summary/TestSummaryPlugin.kt
```diff
@@ -1,115 +1,52 @@
...
-                    timeProvider = timeProvider,
-                    loggerFactory = loggerFactory
-                )
+        if (!target.isRoot()) {
+            val problem = Problem(
+                shortDescription = "TestSummaryPlugin should be applied to root project",
+                context = "TestSummaryPlugin applied to ${target.path}",
+                because = "Summary tasks now registered from CiStep on root project, one per PlanSlug+JobSlug key, " +
+                    "to make cross-app dependency on single report possible"
             )
```
eugene-krivobokov, line 31: > CONFIGURATION IS NOT BACKWARD COMPATIBLE

Proposal: keep empty stubs for old extensions to preserve compilation\configuration.
Ideally, we shouldn't check changes against the internal repository, but for now, it's risky to lose this feedback completely.

dsvoronin: Old extensions will work here for plugin, but not for CiStep

eugene-krivobokov: I just want to preserve backward compatibility and avoid breaking changes. At least in configuration.
Don't know how to make it better and trying to speculate.

dsvoronin: Done, no more breaking changes

subprojects/gradle/cd/src/gradleTest/kotlin/com/avito/ci/steps/TestSummaryStepTest.kt
```diff
@@ -29,24 +30,62 @@ internal class TestSummaryStepTest {
...
-            )
-            .inOrder()
+            .buildSuccessful().run {
+                tasksShouldBeTriggered(
+                    ":app1:instrumentationUi",
+                    ":testSummarySomePlanSomeJob",
+                    ":app1:fullCheck"
+                ).inOrder()
+
+                tasksShouldBeTriggered(
```
RuslanMingaliev, line 40: Is the same check called twice here?

RuslanMingaliev (approved):

subprojects/gradle/cd/src/gradleTest/kotlin/com/avito/ci/steps/TestSummaryStepTest.kt
eugene-krivobokov: It is for different apps - "app1", "app2"

subprojects/gradle/cd/src/main/kotlin/com/avito/ci/steps/TestSummaryPluginBuildStep.kt
```diff
@@ -0,0 +1,65 @@
...
+    override fun registerTask(project: Project, rootTask: TaskProvider<out Task>) {
+        require(project.rootProject.pluginManager.hasPlugin(testSummaryPluginId)) {
+            "$stepName step can't be initialized without $testSummaryPluginId plugin applied to root project"
+        }
+
+        require(configuration.isNotBlank()) {
+            "$stepName step can't be initialized without provided instrumentation configuration"
+        }
+
+        if (useImpactAnalysis && !project.internalModule.isModified()) return
```
eugene-krivobokov, line 43: out of scope, but can be checked in execution phase

subprojects/gradle/tms/src/main/kotlin/com/avito/plugin/TmsPlugin.kt
```diff
@@ -3,19 +3,11 @@ package com.avito.plugin
...
     override fun apply(target: Project) {
-        val extension = target.extensions.create<TmsExtension>("tms")
-
-        target.tasks.register<MarkReportAsSourceTask>(markReportAsSourceTaskName) {
-            reportsHost.set(extension.reportsHost)
-        }
-
-        target.afterEvaluate {
-            requireNotNull(extension.reportsHost.orNull) { "tms.reportsHost should be set" }
-        }
```
eugene-krivobokov, line 19: We could print a warning here about wrong usage

eugene-krivobokov:
> all demo-apps should register all 3 ci steps: tms/flaky/summary

I didn't get this from the code. 
Is it required?
Can we check this contract automatically?
Can we configure it in the root project one time? It's strange that we configure the same in multiple apps.
(we can discuss it at call)

subprojects/gradle/cd/src/main/kotlin/com/avito/ci/steps/TestSummaryPluginBuildStep.kt
dsvoronin: it's common pattern in steps, i don't like it too

subprojects/gradle/cd/src/gradleTest/kotlin/com/avito/ci/steps/TestSummaryStepTest.kt
RuslanMingaliev: ah, didn't notice

subprojects/gradle/cd/src/main/kotlin/com/avito/ci/internal/ReportKey.kt
```diff
@@ -0,0 +1,20 @@
...
+import com.avito.report.model.ReportCoordinates
+import java.util.Locale
+
+/**
+ * RunId doesn't matter here, because it's stable for a single build
+ */
+internal data class ReportKey(val planSlug: String, val jobSlug: String) {
+
+    override fun toString(): String =
+        "${planSlug.capitalize(Locale.getDefault())}${jobSlug.capitalize(Locale.getDefault())}"
```
sergioboi: Why do we need implicit capitalization in `toString`?

dsvoronin: changed to more explicit API

dsvoronin:
> > all demo-apps should register all 3 ci steps: tms/flaky/summary
> 
> I didn't get this from the code.
> Is it required?
> Can we check this contract automatically?
> Can we configure it in the root project one time? It's strange that we configure the same in multiple apps.
> (we can discuss it at call)

dsvoronin:
Sorry for notifications, closed by mistake

dsvoronin:
> > all demo-apps should register all 3 ci steps: tms/flaky/summary
> 
> I didn't get this from the code.
> Is it required?
> Can we check this contract automatically?
> Can we configure it in the root project one time? It's strange that we configure the same in multiple apps.
> (we can discuss it at call)

i want to merge it like this, because common config handled by convention plugin, or demo-apps subprojects block for now, so it's fairly easy to maintain

root project config will introduce wiring by ReportKey, which should be identical between all projects and root configuration, which is in my opinion more fragile

subprojects/gradle/cd/src/main/kotlin/com/avito/ci/internal/ReportKey.kt
```diff
@@ -0,0 +1,21 @@
...
+
+import com.avito.report.model.ReportCoordinates
+import java.util.Locale
+
+/**
+ * RunId doesn't matter here, because it's stable for a single build
+ */
+internal data class ReportKey(val planSlug: String, val jobSlug: String) {
+
+    fun appendToTaskName(taskName: String): String {
```
sergioboi, line 11: So, I don't think that this class should contains this function

sergioboi: We have an abstraction leak here

dsvoronin: why so? it's an internal data class, only to make a clear intention

sergioboi: I will try to share my opinion.
What I see, we have a model `ReportKey` it's a data for report-viewer
Should the report-viewer know about the Gradle abstraction as a task? I don't think so
But Gradle task or Gradle plugin could know about the ReportKey and how modify it to a `task` name

dsvoronin: i don't see it,
but i'll just copy paste a function

sergioboi: I don't have an idea to control everything in our codebase but If you asked me I would move this function to the plugin

PR #1107, opened 2021-06-28
MBS-11293 Use test retry Gradle plugin for unit tests

eugene-krivobokov:
https://github.com/gradle/test-retry-gradle-plugin

Now we can retry unstable tests (unit, gradle, integration).


build-logic/testing-convention/src/main/kotlin/convention.unit-testing.gradle.kts
```diff
@@ -35,6 +40,14 @@ tasks.withType<Test>().configureEach {
         "isInvokedFromIde",
         gradle.startParameter.allInitScripts.find { it.name.contains("ijtestinit") } != null
     )
+    if (isCi.getOrElse(false)) {
+        retry {
+            maxRetries.set(1)
+            filter {
+                includeAnnotationClasses.add("com.avito.test.Flaky")
```
dsvoronin, line 47: i like this API so much!
we should reuse it for our ui tests

subprojects/common/junit-utils/src/main/java/com/avito/test/Flaky.kt
```diff
@@ -0,0 +1,9 @@
+package com.avito.test
+
+/**
+ * Unstable test marker.
+ * It's used for retry logic.
+ */
+annotation class Flaky(
```
dsvoronin, line 7: I think you should specify retention

dsvoronin (approved):

sergioboi (approved):

subprojects/common/junit-utils/src/main/java/com/avito/test/Flaky.kt
eugene-krivobokov: Ok, I'll check it out in next PR

sergioboi: I don't understand what the retention meaning is in that context

eugene-krivobokov: `kotlin.annotation.AnnotationRetention`
As I think, the idea is to use the minimum required retention, to store only needed information.

PR #1109, opened 2021-06-29
MBS-11503 Split device-provider

sergioboi:
That's not all. Next steps will be:
- rename test-runner:service to test-runner:device-worker
- Make dependencies between modules explicit and obvious


subprojects/settings.gradle.kts
```diff
@@ -116,6 +111,13 @@ include(":test-runner:file-storage")
 include(":test-runner:report-viewer")
 include(":test-runner:report-api")
 include(":test-runner:test-model")
+include(":test-runner:client")
+include(":test-runner:device-provider:model")
```
sergioboi, line 115: Structure change

subprojects/test-runner/client/build.gradle.kts
```diff
@@ -18,7 +18,7 @@ dependencies {
     implementation(project(":test-runner:instrumentation-tests-dex-loader"))
     implementation(project(":test-runner:test-report-artifacts"))
     implementation(project(":test-runner:report"))
-    implementation(project(":test-runner:device-provider"))
+    implementation(project(":test-runner:device-provider:impl"))
```
sergioboi, line 21: Want to avoid this dependency in the future. API must be enough for client

subprojects/test-runner/device-provider/impl/build.gradle.kts
```diff
@@ -8,12 +8,9 @@ plugins {
     id("convention.libraries")
 }
 
-publish {
```
sergioboi, line 11: Delete because I don't know for what reason this exists

subprojects/test-runner/instrumentation-tests/build.gradle.kts
```diff
@@ -25,7 +25,7 @@ dependencies {
     implementation(project(":test-runner:instrumentation-tests-dex-loader"))
     implementation(project(":test-runner:report"))
     implementation(project(":test-runner:test-annotations"))
-    implementation(project(":test-runner:device-provider"))
+    implementation(project(":test-runner:device-provider:model"))
```
sergioboi, line 28: Now depends only on the model

subprojects/test-runner/instrumentation-tests/build.gradle.kts
```diff
@@ -37,7 +37,7 @@ dependencies {
     testImplementation(testFixtures(project(":test-runner:report-api")))
     testImplementation(testFixtures(project(":test-runner:report-viewer")))
     testImplementation(testFixtures(project(":test-runner:client")))
-    testImplementation(testFixtures(project(":test-runner:device-provider")))
+    testImplementation(testFixtures(project(":test-runner:device-provider:impl")))
```
sergioboi, line 40: Will be deleted in the future

subprojects/test-runner/report-viewer/src/main/kotlin/com/avito/report/internal/model/KtulhuTestCase.kt
```diff
@@ -3,4 +3,4 @@ package com.avito.report.internal.model
 /**
  * Прилетает в prepared_data после анализа совпадения с тестохранилкой если есть test_case_id
  */
-internal data class TestCase(val description: String?)
+internal data class KtulhuTestCase(val description: String?)
```
sergioboi: Rename because of the name clashing with test-runner TestCase

dsvoronin: It is called `Cthulhu`, and we can call it just `AvitoTestCase`

subprojects/test-runner/device-provider/impl/build.gradle.kts
dsvoronin: It defines artifactId of maven module, when you applying a dependency in project it's:

`com.avito.android:<WHAT IS SET HERE>:version`

please check if this module should be applied as dependency of it is only supposed to be included transitively

sergioboi: As I understand if Integration compile in Avito was done that mean everything is ok

dsvoronin: yes

dsvoronin (approved):

PR #1110, opened 2021-06-29
MBS-11302 Stabilize flaky tests

eugene-krivobokov:

PR #1111, opened 2021-06-29
MBS-11302 increase gradle test kit parallel performance

dsvoronin:
~4x improvement, see comment in GradleTestKit.kt for details

### DIFF
--- a/build-logic/testing-convention/src/main/kotlin/convention.gradle-testing.gradle.kts
+++ b/build-logic/testing-convention/src/main/kotlin/convention.gradle-testing.gradle.kts
@@ -34,12 +34,40 @@ val gradleTestTask = tasks.register<Test>("gradleTest") {
 
     useJUnitPlatform()
 
-    @Suppress("MagicNumber")
-    maxParallelForks = 8
+    /**
+     * timings provided only as relative to each other
+     * gradleTest:
+     *  1: 2m14s
+     *  16: 3m16s
+     */
+    maxParallelForks = 1
+
+    /**
+     * usually there is a small amount of test classes per module, and tests are not so memory hungry
+     */
+    setForkEvery(null)
+
+    jvmArgs = listOf("-XX:+UseGCOverheadLimit", "-XX:GCTimeLimit=10")
+
+    minHeapSize = "128m"
+    maxHeapSize = "256m"
 
     failFast = false
 
+    /**
+     * fix for multiple `WARNING: Illegal reflective access`
+     */
+    jvmArgs = listOf(
+        "--add-opens",
+        "java.base/java.lang=ALL-UNNAMED",
+        "--add-opens",
+        "java.base/java.lang.invoke=ALL-UNNAMED",
+        "--add-opens",
+        "java.base/java.util=ALL-UNNAMED"
+    )
+
     systemProperty("rootDir", "${project.rootDir}")
+    systemProperty("buildDir", "$buildDir")
     systemProperty("kotlinVersion", kotlinVersion)
     systemProperty("compileSdkVersion", libs.compileSdkVersion)
     systemProperty("buildToolsVersion", libs.buildToolsVersion)
@@ -93,7 +121,7 @@ configure<KotlinJvmProjectExtension> {
     target.compilations.getByName("gradleTest")
         .associateWith(target.compilations.getByName("main"))
 
-    plugins.withType<JavaTestFixturesPlugin>() {
+    plugins.withType<JavaTestFixturesPlugin> {
         target.compilations.getByName("gradleTest")
             .associateWith(target.compilations.getByName("testFixtures"))
     }
--- a/subprojects/gradle/critical-path/critical-path/src/gradleTest/kotlin/com/avito/android/critical_path/CriticalPathTest.kt
+++ b/subprojects/gradle/critical-path/critical-path/src/gradleTest/kotlin/com/avito/android/critical_path/CriticalPathTest.kt
@@ -111,8 +111,7 @@ internal class CriticalPathTest {
         // no op
     }
 
-    // TODO: this test is flaky but enabled to collect more info
-    //  If it's failed, please add info to MBS-11302
+    @Disabled("Fails constantly")
     @Test
     fun `independent routes - path has the longest one`() {
         setupTasks(
--- a/subprojects/gradle/test-project/src/main/kotlin/com/avito/test/gradle/GradleTestKit.kt
+++ b/subprojects/gradle/test-project/src/main/kotlin/com/avito/test/gradle/GradleTestKit.kt
@@ -44,8 +44,25 @@ fun gradlew(
 
     val isInvokedFromIde = System.getProperty("isInvokedFromIde")?.toBoolean() ?: false
 
+    /**
+     * default is single temp dir for all tests using test kit, like /tmp/test-kit-dir-username
+     * it leads to concurrent issues between multiple modules in parallel:
+     *   "The file lock is held by a different Gradle process"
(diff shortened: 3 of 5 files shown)


subprojects/gradle/critical-path/critical-path/src/gradleTest/kotlin/com/avito/android/critical_path/CriticalPathTest.kt
```diff
@@ -111,8 +111,7 @@ internal class CriticalPathTest {
         // no op
     }
 
-    // TODO: this test is flaky but enabled to collect more info
-    //  If it's failed, please add info to MBS-11302
+    @Disabled("Fails constantly")
```
eugene-krivobokov, line 114: This test is already deleted in #1110

subprojects/gradle/test-project/src/main/kotlin/com/avito/test/gradle/GradleTestKit.kt
```diff
@@ -44,8 +44,25 @@ fun gradlew(
...
 
+    /**
+     * default is single temp dir for all tests using test kit, like /tmp/test-kit-dir-username
+     * it leads to concurrent issues between multiple modules in parallel:
+     *   "The file lock is held by a different Gradle process"
+     * and silent lock awaiting, increasing our test times on all gradleTest tasks in parallel to ~8min(for scale)
+     * probably because test kit dir also used as gradle-user-home for test
+     *
+     * Sharing test kit dir between different processes, using maxParallelForks>1, inside single module - is ok
+     * unique test-kit-dir per every tempDir in test is not ok! (19min)
```
eugene-krivobokov, line 55:
```suggestion
     * unique test-kit-dir per Gradle test kit runner instance in test is not ok! (19min)
```
```suggestion
     * unique test-kit-dir per test is not ok! (19min)
```
?

build-logic/testing-convention/src/main/kotlin/convention.gradle-testing.gradle.kts
```diff
@@ -34,12 +34,40 @@ val gradleTestTask = tasks.register<Test>("gradleTest") {
...
+     * gradleTest:
+     *  1: 2m14s
+     *  16: 3m16s
+     */
+    maxParallelForks = 1
+
+    /**
+     * usually there is a small amount of test classes per module, and tests are not so memory hungry
+     */
+    setForkEvery(null)
```
eugene-krivobokov, line 48: > usually

How about protecting from it by setting a big enough value?
Will it affect execution when it's greater than a usual count of tests in a module?

build-logic/testing-convention/src/main/kotlin/convention.gradle-testing.gradle.kts
```diff
@@ -34,12 +34,40 @@ val gradleTestTask = tasks.register<Test>("gradleTest") {
 
     useJUnitPlatform()
 
-    @Suppress("MagicNumber")
-    maxParallelForks = 8
+    /**
+     * timings provided only as relative to each other
```
eugene-krivobokov, line 38: > relative to each other

Didn't get what it means. 
Is it about changing only one parameter while optimization?

build-logic/testing-convention/src/main/kotlin/convention.gradle-testing.gradle.kts
```diff
@@ -34,12 +34,40 @@ val gradleTestTask = tasks.register<Test>("gradleTest") {
...
+     * gradleTest:
+     *  1: 2m14s
+     *  16: 3m16s
```
eugene-krivobokov, line 41: I'd like to see also how to optimize and objective constraints - CPU, memory?
Exact values can be optimal only to one machine but not for others.
Comparing only them without explaining constraints makes exact values less reusable.

build-logic/testing-convention/src/main/kotlin/convention.gradle-testing.gradle.kts
```diff
@@ -34,12 +34,40 @@ val gradleTestTask = tasks.register<Test>("gradleTest") {
...
+
+    jvmArgs = listOf("-XX:+UseGCOverheadLimit", "-XX:GCTimeLimit=10")
+
+    minHeapSize = "128m"
+    maxHeapSize = "256m"
 
     failFast = false
 
+    /**
+     * fix for multiple `WARNING: Illegal reflective access`
```
eugene-krivobokov, line 58: I'd like to clarify one point.

These parameters are also applied in convention plugin for unit tests:

```kotlin
tasks.withType<Test>().configureEach {
```

If I'm not mistaken, the unit tests convention plugin applies to all modules through the kotlin convention plugin.
Thus, it's not clear why we duplicate the same configuration in `tasks.register<Test>`.
I don't want such implicit overriding but I'd expect it from the code. 
If it's accidental, such relations between convention plugins is error-prone.
Please correct me if I'm wrong.

eugene-krivobokov (approved):

dsvoronin:
will fix review notes afterwards

build-logic/testing-convention/src/main/kotlin/convention.gradle-testing.gradle.kts
sergioboi: How do you measure?

subprojects/gradle/critical-path/critical-path/src/gradleTest/kotlin/com/avito/android/critical_path/CriticalPathTest.kt
dsvoronin: i don't see it in develop so it's merged ok

build-logic/testing-convention/src/main/kotlin/convention.gradle-testing.gradle.kts
dsvoronin: it was a default before;
i saw no impact setting it to 50

i don't want to change it in this PR

build-logic/testing-convention/src/main/kotlin/convention.gradle-testing.gradle.kts
dsvoronin: You're absolutely right.
I removed all duplicates and found an error in setting jvmArgs, which is replacing extra(added in buildscript) set of args, instead of adding to it

The only problem left is values overwrite.
For example `maxParallelForks` set to `halfofavailableProcessors` in unit tests and then overwritten configuring `gradleTest` task with value `1`. It works that way only because of plugin apply ordering, which is error prone.

build-logic/testing-convention/src/main/kotlin/convention.gradle-testing.gradle.kts
dsvoronin: I tried to explain my decision better in comment, take a look

build-logic/testing-convention/src/main/kotlin/convention.gradle-testing.gradle.kts
dsvoronin: i have commited a performance scenario for gradle profiler and also redone the measurements with more accurate results

build-logic/testing-convention/src/main/kotlin/convention.gradle-testing.gradle.kts
eugene-krivobokov: > It works that way only because of plugin apply ordering, which is error prone

Can we check this contract explicitly inside gradle testing convention plugin? (`withPlugin` or similar constructs)

PR #1120, opened 2021-07-01
MBS-11502 remove legacy reporting code and inMemoryReport flag

dsvoronin:


subprojects/common/time/src/main/kotlin/com/avito/time/TimeProvider.kt
```diff
@@ -3,13 +3,13 @@ package com.avito.time
...
 
-    fun nowInSeconds(): Long
+    public fun nowInSeconds(): Long
 
-    fun now(): Date
+    public fun now(): Date
 
-    fun toDate(seconds: Long): Date
+    public fun toDate(seconds: Long): Date
 }
```
eugene-krivobokov, line 15: for the record, this interface is too shallow and should have only one method - `nowMs`. 
All other methods are just conversion, pure functions.

eugene-krivobokov (approved):

PR #1122, opened 2021-07-01
MBS-10881 use version catalog

dsvoronin:


RuslanMingaliev (approved):

eugene-krivobokov:
https://docs.gradle.org/7.0/userguide/platforms.html#sub:central-declaration-of-dependencies

build-logic/kotlin-convention/src/main/kotlin/convention.kotlin-base.gradle.kts
```diff
@@ -1,12 +1,11 @@
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 
 plugins {
-    id("convention.libraries")
     id("convention.unit-testing")
 }
 
 dependencies {
-    add("implementation", libs.kotlinStdlib)
+    add("implementation", "org.jetbrains.kotlin:kotlin-stdlib:1.5.20")
```
eugene-krivobokov, line 8: Can we share such versions between convention plugins through properties?

PR #1126, opened 2021-07-02
MBS-10881 build-logic refactorings

dsvoronin:
remove nebulaPluginVersion property duplicate
rename build-logic modules, to exclude verbose "-convention" suffix
create build-logic/checks module
add "convention.detekt"
remove systemprop.kotlinVersion
remove keeper plugin setup
workaround to use version catalog in buildscript and convention plugins


build-logic/kotlin/src/main/kotlin/convention.kotlin-base.gradle.kts
```diff
@@ -4,15 +4,10 @@ plugins {
...
-    add("implementation", "org.jetbrains.kotlin:kotlin-stdlib:1.5.20")
-}
-
 /**
- * 1.4 used because of kotlin dsl bundled version
- * todo use 1.5
+ * overrides 1.4, that comes from kotlin dsl bundled version
  */
-val kotlinLanguageVersion = "1.4"
+val kotlinLanguageVersion = "1.5"
```
eugene-krivobokov: I'd good to make this change in a separate PR.
I don't see how it's related to other changes.

eugene-krivobokov (approved):

build-logic/kotlin/src/main/kotlin/convention.kotlin-base.gradle.kts
dsvoronin: i agree, let me separate it

build-logic/testing/src/main/kotlin/convention.unit-testing.gradle.kts
```diff
@@ -1,3 +1,5 @@
+import gradle.kotlin.dsl.accessors._aefff6f539276b188c30d8843d1e61e9.testImplementation
+import gradle.kotlin.dsl.accessors._aefff6f539276b188c30d8843d1e61e9.testRuntimeOnly
```
eugene-krivobokov, line 2: What is this package? Does it expected?

dsvoronin: nope, it could lead to accessor problems. let me think if we can defent from it via detekt or something similar

dsvoronin: https://github.com/avito-tech/avito-android/pull/1131

PR #1128, opened 2021-07-05
MBS-9841 Add checks for View and RecyclerView

RuslanMingaliev:


eugene-krivobokov (approved):

subprojects/android-test/ui-testing-core/src/main/kotlin/com/avito/android/test/checks/Checks.kt
```diff
@@ -58,6 +59,8 @@ interface Checks : LabelChecks {
 
     fun isNotVisible()
 
+    fun hasVisibility(visibility: ViewMatchers.Visibility)
```
sergioboi, line 62: I was confused that we have `isNotVisible` and `hasVisibility`. We should have one API to do one thing

subprojects/android-test/ui-testing-core/src/main/kotlin/com/avito/android/test/matcher/RecyclerViewMatcher.kt
```diff
@@ -48,6 +49,15 @@ class RecyclerViewMatcher {
         }
     }
 
+    fun hasViewTypeAtPosition(position: Int, viewType: Int): Matcher<View> = object : RecyclerViewMatcher() {
```
sergioboi, line 52: Let's make anonymous object separate class or file

subprojects/android-test/ui-testing-core/src/main/kotlin/com/avito/android/test/matcher/RecyclerViewMatcher.kt
```diff
@@ -48,6 +49,15 @@ class RecyclerViewMatcher {
...
+        override fun performMatching(recyclerView: RecyclerView): Boolean {
+            val adapter = recyclerView.adapter ?: error("Adapter is not attached to RecyclerView")
```
sergioboi, line 54: requireNotNull(recyclerView.adapter) { "Adapter is not attached to RecyclerView" }

subprojects/android-test/ui-testing-core/src/main/kotlin/com/avito/android/test/matcher/RecyclerViewMatcher.kt
sergioboi: The idea was that we want to check any parameter of view, not only the `viewType`
I want to see hasView(position, matcher)

subprojects/android-test/ui-testing-core/src/main/kotlin/com/avito/android/test/matcher/RecyclerViewMatcher.kt
```diff
@@ -48,6 +49,15 @@ class RecyclerViewMatcher {
...
+            return adapter.getItemViewType(position) == viewType
```
sergioboi, line 55: What will happen when `adapter` doesn't contain view at position?

subprojects/android-test/ui-testing-core/src/main/kotlin/com/avito/android/test/checks/Checks.kt
RuslanMingaliev: I've added a more generic way to check visibility that I think to be correct. But I can't delete the old ones, because it will break backward compatibility.

sergioboi: I think the current setup adds confusion for users and I don't see a value for having `hasVisibility` fun

eugene-krivobokov: > can't delete the old ones, because it will break backward compatibility.

For the record, changing API is possible and not a blocker. 
We only need to do it in multiple steps with deprecation and changing clients.

PR #1130, opened 2021-07-05
MBS-11502 extract reports

dsvoronin:
First part of extraction. Current state is:

image

to remove `:client` dependency on `:report-viewer` two things needs to be addressed:
1.  Filter logic should not depend on report viewer fetch, so there should be another task(preferably separate plugin), that will produce previous report results as an output
2. Actions logic should be extracted to separate task/plugin also; getting current suite or it's params (coordinates) as inputs and then wired by CiStep


subprojects/test-runner/client/build.gradle.kts
```diff
@@ -16,7 +16,6 @@ dependencies {
     implementation(projects.testRunner.testAnnotations)
     implementation(projects.testRunner.instrumentationTestsDexLoader)
     implementation(projects.testRunner.testReportArtifacts)
-    implementation(projects.testRunner.report)
```
eugene-krivobokov, line 19: In the description of PR `client` uses `report`

eugene-krivobokov: Have you tried to use `make project_graph_report` to draw the current state?

eugene-krivobokov (approved):

sergioboi:
Sorry i can't review this PR because remote branch doesn't exist anymore

dsvoronin:
> Sorry i can't review this PR because remote branch doesn't exist anymore

I don't know why branches deleted. 
It wasn't before? You reviewed pretty old merged PR's as far as i remembered
Or it's only a IDE review thing?

sergioboi:
Before I make a review only in the Github UI
Maybe I have an opportunity to review but must find it

PR #1131, opened 2021-07-06
MBS-10881 remove direct accessor import, add detekt check to prevent it

dsvoronin:
add umbrella task `:detektAll` on root project, that runs detekt for all included builds

### DIFF
--- a/build-logic/testing/src/main/kotlin/convention.unit-testing.gradle.kts
+++ b/build-logic/testing/src/main/kotlin/convention.unit-testing.gradle.kts
@@ -1,5 +1,3 @@
-import gradle.kotlin.dsl.accessors._aefff6f539276b188c30d8843d1e61e9.testImplementation
-import gradle.kotlin.dsl.accessors._aefff6f539276b188c30d8843d1e61e9.testRuntimeOnly
 import org.jetbrains.kotlin.gradle.dsl.KotlinJvmProjectExtension
 import org.jetbrains.kotlin.gradle.plugin.KotlinBasePluginWrapper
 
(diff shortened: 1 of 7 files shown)


build.gradle.kts
```diff
@@ -29,3 +33,9 @@ tasks.register<Exec>("installGitHooks") {
 
 val initialTaskNames: List<String> = project.gradle.startParameter.taskNames
 project.gradle.startParameter.setTaskNames(initialTaskNames + listOf("installGitHooks"))
+
+tasks.register("detektAll").configure {
+    dependsOn(gradle.includedBuild("build-logic").task(":detektAll"))
+    dependsOn(gradle.includedBuild("subprojects").task(":detektAll"))
+    dependsOn(gradle.includedBuild("samples").task(":detektAll"))
+}
```
eugene-krivobokov, line 41: Can't we iterate through all included builds?
`getIncludedBuilds()`

eugene-krivobokov (approved):

build.gradle.kts
dsvoronin: yep, fixed
thanks

build-logic/testing/src/main/kotlin/convention.unit-testing.gradle.kts
```diff
@@ -59,12 +57,12 @@ if (project.name != "gradle-kotlin-dsl-accessors") {
 
     plugins.withType<KotlinBasePluginWrapper> {
         dependencies {
-            testImplementation(libs.junitJupiterApi)
-            testImplementation(libs.truth)
+            add("testImplementation", libs.junitJupiterApi)
```
sergioboi, line 60: What is the difference?

build.gradle.kts
```diff
@@ -29,3 +33,9 @@ tasks.register<Exec>("installGitHooks") {
 
 val initialTaskNames: List<String> = project.gradle.startParameter.taskNames
 project.gradle.startParameter.setTaskNames(initialTaskNames + listOf("installGitHooks"))
+
+tasks.register("detektAll").configure {
```
sergioboi, line 37: Why do we need register the detektAll if we register it in the convention plugin

samples/build.gradle.kts
```diff
@@ -0,0 +1,24 @@
+plugins {
+    id("convention.detekt")
+}
+
+buildscript {
```
sergioboi, line 5: Too many workarounds )))

build-logic/testing/src/main/kotlin/convention.unit-testing.gradle.kts
dsvoronin: these acessors (testImplementation), was imported from wrong file, and in fact are not accessible here, because needed plugin (kotlin) not available during convention plugin resolution

build.gradle.kts
dsvoronin: it's a chiken and egg problem. detektAll registered inside convention plugin of this project, but it needs detekt task itself

samples/build.gradle.kts
dsvoronin: It's till 7.2, all fixed according to github milestone

PR #1132, opened 2021-07-07
MBS-11583 fix not reported verdict determiner condition

dsvoronin:
Take a look at case, described in a jira task first..

There were two problems, possible clashing:
 - using kotlin `subtract` on different classes leads to always full first collection (maybe not a case in our production, but was easily reproduced in testcase)
 - not reported test (actual, not our converted Lost's) leads to different logic branch and to unexpected fail even with `suppressedFailures` on
 
added problematic test case, should be ok now

also somehow RunnerArgs dump was not posted in teamcity artifacts; move file inside configuration dir; hope it will now


eugene-krivobokov (approved):

sergioboi:
Won't review there are no remote branch

PR #1135, opened 2021-07-08
MBS-11502 extract report processor

dsvoronin:


subprojects/test-runner/client/build.gradle.kts
```diff
@@ -11,36 +11,35 @@ publish {
...
-    implementation(projects.testRunner.fileStorage)
+    implementation(projects.common.math)
+    implementation(projects.common.problem)
+    implementation(projects.common.result)
+    implementation(projects.gradle.process)
+    implementation(projects.gradle.traceEvent)
+    implementation(projects.testRunner.deviceProvider.api)
+    implementation(projects.testRunner.deviceProvider.impl)
+    implementation(projects.testRunner.instrumentationTestsDexLoader)
+    implementation(projects.testRunner.reportProcessor)
```
sergioboi, line 29: So could we make a next step and remove dependencies on `testReportArtifacts` and `reportProcessor`

sergioboi (approved):

subprojects/test-runner/client/build.gradle.kts
dsvoronin: actually we can't
let's discuss it tomorrow

PR #1138, opened 2021-07-09
MBS-10151 update fabric8io kubernetes client

dsvoronin:
Anticipating question: "Did you check if we can drop official client now?" I didn't check yet, i want to make sure new client version is working, and then explore oidc auth mechanism in fabric8io

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -51,7 +51,7 @@ androidAnnotations = "androidx.annotation:annotation:1.1.0"
 playServicesMaps = "com.google.android.gms:play-services-maps:17.0.0"
 
 officialKubernetesClient = "io.kubernetes:client-java:8.0.0"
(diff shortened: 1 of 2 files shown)


gradle/libs.versions.toml
```diff
@@ -51,7 +51,7 @@ androidAnnotations = "androidx.annotation:annotation:1.1.0"
 playServicesMaps = "com.google.android.gms:play-services-maps:17.0.0"
 
 officialKubernetesClient = "io.kubernetes:client-java:8.0.0"
-kubernetesClient = "io.fabric8:kubernetes-client:4.9.0"
+kubernetesClient = "io.fabric8:kubernetes-client:5.5.0"
```
eugene-krivobokov, line 54: For the record: https://github.com/fabric8io/kubernetes-client#compatibility-matrix
4.9.0 has the same minimum k8s supported version as 5.5.0

eugene-krivobokov (approved):

PR #1145, opened 2021-07-14
Mbs 11609 extract kubernetes api

dsvoronin:
Extracted KubernetesApi from `:device-provider` to `:kubernetes` module
Move`:kubernetes` module to test-runner 
fix EmulatorsLogsReporting: append vs write


subprojects/common/waiter/src/main/kotlin/com/avito/android/waiter/Wait.kt
```diff
@@ -1,19 +1,17 @@
...
 import com.avito.android.Result
 import com.avito.android.Result.Failure
 import com.avito.android.Result.Success
-import com.avito.logger.Logger
-import kotlinx.coroutines.delay
 import java.util.concurrent.TimeUnit
 
-internal suspend fun <T> waitForCondition(
-    logger: Logger,
+public suspend fun <T> waitForCondition(
```
dsvoronin, line 8: it's is now used in two separate modules, and there is no `coroutines` dependency here

subprojects/test-runner/device-provider/impl/src/main/kotlin/com/avito/android/runner/devices/internal/EmulatorsLogsReporterImpl.kt
```diff
@@ -9,10 +10,11 @@ internal class EmulatorsLogsReporterImpl(
...
 
-    override fun reportEmulatorLogs(emulatorName: Serial, log: String) {
-        getLogFile(emulatorName.value)
-            .apply { parentFile?.mkdirs() }
-            .writeText(log)
+    override fun reportEmulatorLogs(pod: KubePod, emulatorName: Serial, log: String) {
+        val logFile = getLogFile(emulatorName.value)
+        logFile.parentFile?.mkdirs()
+        logFile.appendText("--- Logs of emulator: $pod ---\n")
+        logFile.appendText("$log\n")
```
dsvoronin, line 17: @eugene-krivobokov ^^

subprojects/common/waiter/src/main/kotlin/com/avito/android/waiter/Wait.kt
sergioboi: This action is an example of why `coroutines` are peace of shit
Let's add a TODO to code remove it with FLOW or RX

sergioboi (approved):

sergioboi:
Approve with no watching

PR #1151, opened 2021-07-19
MBS-10151 add support for apps/v1 k8s deployments api

dsvoronin:
legacy mode `useLegacyExtensionsV1Beta` enabled by default

### DIFF
--- a/subprojects/test-runner/device-provider/impl/src/main/kotlin/com/avito/android/runner/devices/internal/kubernetes/ReservationDeploymentFactoryImpl.kt
+++ b/subprojects/test-runner/device-provider/impl/src/main/kotlin/com/avito/android/runner/devices/internal/kubernetes/ReservationDeploymentFactoryImpl.kt
@@ -27,6 +27,7 @@ internal class ReservationDeploymentFactoryImpl(
     private val buildId: String,
     private val buildType: String,
     private val deploymentNameGenerator: DeploymentNameGenerator,
+    private val useLegacyExtensionsV1Beta: Boolean,
     loggerFactory: LoggerFactory
 ) : ReservationDeploymentFactory {
 
@@ -136,7 +137,7 @@ internal class ReservationDeploymentFactoryImpl(
             .plus("deploymentName" to deploymentName)
 
         return newDeployment {
-            apiVersion = "extensions/v1beta1"
+            apiVersion = if (useLegacyExtensionsV1Beta) "extensions/v1beta1" else "apps/v1"
             metadata {
                 name = deploymentName
                 labels = deploymentMatchLabels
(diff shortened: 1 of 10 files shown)


subprojects/test-runner/device-provider/impl/src/main/kotlin/com/avito/android/runner/devices/internal/kubernetes/ReservationDeploymentFactoryImpl.kt
```diff
@@ -136,7 +137,7 @@ internal class ReservationDeploymentFactoryImpl(
             .plus("deploymentName" to deploymentName)
 
         return newDeployment {
-            apiVersion = "extensions/v1beta1"
+            apiVersion = if (useLegacyExtensionsV1Beta) "extensions/v1beta1" else "apps/v1"
```
eugene-krivobokov, line 140: Probably out of the scope of this PR, but it might be better to express a contract in terms of k8s versions because it already defines API. For the client might be easier to think and operate in this way.

This is how I see an ideal solution. Test runner checks k8s version automatically and does fail-fast check: 
"The test runner is supporting only N...M k8s version, but pointed to X version".

If this is not reasonable, we could move this knowledge and responsibility to the client configuration. It may look like a leaky abstraction but will help in troubleshooting.

Example use case to illustrate the problem:

I pointed the test runner to the newest k8s version and faced other backward compatibility problems.
How should I know what version is supported? Can I change this behavior?

Juggling by low-level feature flags for exact k8s features looks not so extensible and convenient as by k8s versions.

eugene-krivobokov (approved):

subprojects/test-runner/device-provider/impl/src/main/kotlin/com/avito/android/runner/devices/internal/kubernetes/ReservationDeploymentFactoryImpl.kt
dsvoronin: i think we should express somewhere in documentation that only `apps/v1` versions are supported
This flag will be remove alongside with the `extensions/v1beta` support.

More extendible client with multiple supported API versions maybe a good idea, but there is only `apps/v1' in the wild for now afaik

eugene-krivobokov: > express somewhere in documentation that only apps/v1 versions are supported

The idea is that this is harder to find. E.g., I want to run tests against GCE. It'd easier to check k8s version than digging into API details.

dsvoronin: i hope we can skip this problem entirely until `apps/v2` will be released
you can't create something as old as 1.8 in GCE right now

image

PR #1156, opened 2021-07-22
MBS-11673 Validate test report dir location

eugene-krivobokov:
### The problem

We had an implicit contract for a test report location.
In AndroidJUnitRunner we get it from AndroidJUnitRunner.
In Gradle we need to use the same path for adb.

### Changes

Added an abstraction for an application directory.

There are two implementations:

- Hardcoded for adb
- From context for AndroidJUnitRunner

Implementation for AndroidJUnitRunner implicitly validates itself against implementation for adb.


subprojects/test-runner/client/src/main/kotlin/com/avito/runner/listener/ArtifactsTestListener.kt
```diff
@@ -131,24 +131,6 @@ internal class ArtifactsTestListener(
...
     }
 
-    private fun pullResults(device: Device, remotePath: Path, destination: Path): Result<File> {
-        val pullResult = device.pullDir(
-            deviceDir = remotePath,
-            hostDir = destination,
-            validator = reportArtifactsPullValidator
-        )
-        return if (device.api >= 30) {
-            val cleaningResult = device.clearDirectory(remotePath)
```
eugene-krivobokov, line 141: Removed responsibility for cleaning. This is what DeviceWorker should do and it does already.

subprojects/test-runner/test-report-artifacts/src/main/kotlin/com/avito/report/TestArtifactsProviderFactory.kt
```diff
@@ -36,19 +29,13 @@ public object TestArtifactsProviderFactory {
...
-        appUnderTestPackage: String,
+    public fun createForAdb(
+        appDirProvider: ApplicationDirProvider,
         name: TestName,
     ): TestArtifactsProvider {
-        val dataPath = if (api >= 30) {
-            "/storage/emulated/0/Android/media/$appUnderTestPackage"
-        } else {
-            "/sdcard/Android/data/$appUnderTestPackage/files"
-        }
```
eugene-krivobokov, line 48: I've hidden this knowledge in a new abstraction - `ApplicationDirProvider`.
This abstraction is used in a Gradle for adb and in AndroidJunitRunner.

subprojects/test-runner/test-inhouse-runner/src/main/kotlin/com/avito/android/internal/RuntimeApplicationDirProvider.kt
```diff
@@ -0,0 +1,39 @@
...
+            appPackage = context.packageName,
+        ).dir
+
+        require(nonRuntimeDir.canonicalPath == runtimeDir.canonicalPath) {
+            "Expected the same dirs: " +
+                "from context: $runtimeDir, " +
+                "without context (for adb): $nonRuntimeDir"
+        }
+    }
+}
```
eugene-krivobokov, line 39: Here we hide knowledge about the output directory and validate it against the same for adb.

subprojects/test-runner/test-inhouse-runner/src/main/kotlin/com/avito/android/runner/TestRunEnvironment.kt
```diff
@@ -150,18 +144,6 @@ internal fun parseReportDestination(argumentsProvider: ArgsProvider): ReportDest
...
-    return if (Build.VERSION.SDK_INT >= 30) {
-        // public dir which is accessible by adb
-        InstrumentationRegistry.getInstrumentation().targetContext.externalMediaDirs[0]
-    } else {
-        ContextCompat.getExternalFilesDirs(
-            InstrumentationRegistry.getInstrumentation().targetContext,
-            null
-        )[0]
-    }
-}
```
eugene-krivobokov, line 163: Moved to `RuntimeApplicationDirProvider`. 
See no need to know it here, because it depends only on Context but not from instrumentation arguments.

subprojects/test-runner/test-report-artifacts/src/main/kotlin/com/avito/report/ReportDirProviderWithCreation.kt
```diff
@@ -1,60 +0,0 @@
-package com.avito.report
-
-import com.avito.android.Result
-import java.io.File
-
```
eugene-krivobokov, line 5: Just extracted classes to independent files.

subprojects/test-runner/test-inhouse-runner/src/main/kotlin/com/avito/android/internal/RuntimeApplicationDirProvider.kt
```diff
@@ -0,0 +1,39 @@
+package com.avito.android.internal
+
+import android.content.Context
+import android.os.Build
+import androidx.core.content.ContextCompat
+import com.avito.report.ApplicationDirProvider
+import com.avito.report.ApplicationDirProviderFactory
+import java.io.File
+
+internal class RuntimeApplicationDirProvider(context: Context) : ApplicationDirProvider {
```
eugene-krivobokov, line 10: Move to test-instrumentation-runner

dsvoronin (approved):

subprojects/test-runner/test-inhouse-runner/src/main/kotlin/com/avito/android/internal/RuntimeApplicationDirProvider.kt
eugene-krivobokov: Can't place it inside `test-instrumentation-runner`.
Now, this module looks like a generic abstraction over AndroidJunitRunner which doesn't know about test runner.
Let it be that way.

dsvoronin: okay, time to extract another part from inhouse runner then ) )

PR #1163, opened 2021-08-04
MBS-11728 Update repository for kotlinx-html-jvm

RuslanMingaliev:


build-logic-settings/dependency-plugin/src/main/kotlin/convention-dependencies.settings.gradle.kts
```diff
@@ -53,13 +53,20 @@ dependencyResolutionManagement {
...
+         * TODO: Remove this block after MBS-11267
+         */
         exclusiveContent {
             forRepository {
                 maven {
                     setUrlOrProxy(
-                        artifactoryUrl = artifactoryUrl,
-                        repositoryName = "KotlinX",
-                        originalRepo = "https://kotlin.bintray.com/kotlinx"
+                        artifactoryUrl = null, // TODO: Add mirroring for this repo
```
sergioboi: @dsvoronin should know

build-logic-settings/dependency-plugin/src/main/kotlin/convention-dependencies.settings.gradle.kts
```diff
@@ -53,13 +53,20 @@ dependencyResolutionManagement {
                 includeGroup("com.google.testing.platform")
             }
         }
+        /**
+         * This repo is used for kotlinx-html:0.7.2 (transitive dependency of current Detekt version)
```
sergioboi, line 57: This repo used kotlinx-html directly from the build-verdict plugin and transitively from detekt

gradle/libs.versions.toml
```diff
@@ -22,7 +22,7 @@ kotlinStdlib = { module = "org.jetbrains.kotlin:kotlin-stdlib", version.ref = "k
 kotlinStdlibJdk7 = { module = "org.jetbrains.kotlin:kotlin-stdlib-jdk7", version.ref = "kotlin" }
 kotlinReflect = { module = "org.jetbrains.kotlin:kotlin-reflect", version.ref = "kotlin" }
 kotlinCompilerEmbeddable = { module = "org.jetbrains.kotlin:kotlin-compiler-embeddable", version.ref = "kotlin" }
-kotlinHtml = "org.jetbrains.kotlinx:kotlinx-html-jvm:0.7.2"
+kotlinHtml = "org.jetbrains.kotlinx:kotlinx-html-jvm:0.7.3"
```
sergioboi, line 25: As I see we've already used 0.7.3 let's make a call one more time and try to force using the 0.7.3 even for old detekt

sergioboi (changes requested):

build-logic-settings/dependency-plugin/src/main/kotlin/convention-dependencies.settings.gradle.kts
dsvoronin: thanks! 
I will add mirror before merging it

build-logic-settings/dependency-plugin/src/main/kotlin/convention-dependencies.settings.gradle.kts
RuslanMingaliev: There is no problem with our build-verdict plugin because it uses correct version of kotlin-html which is available in mavenCentral.

dsvoronin (approved):

PR #1164, opened 2021-08-06
MBS-11730 Add warn logs that plugin is disabled

sergioboi:
it will help to react if you forgot about -Pci=true


sergioboi:
I spent an hour trying to realize where my cd tasks. Maybe I do smthg wrong, but I don't know how to do better

eugene-krivobokov:
> I spent an hour trying to realize where my cd tasks. Maybe I do smthg wrong, but I don't know how to do better

I want to mention the idea of eliminating `ci` property - MBS-7017.
It requires optimizing plugins - impact analysis primarily.

PR #1166, opened 2021-08-13
MBS-11738 Bind uploadCrashlyticsMappingFile to deploy step

sergioboi:
Better watch this PR commit by commit
in test-project I add functionallity to add classpath plugins via DSL
The main logic is in cd module

Good luck and have fun


gradle.properties
```diff
@@ -0,0 +1,15 @@
+kotlin.code.style=official
```
sergioboi, line 1: @dsvoronin review this. Maybe this file isn't needed

subprojects/gradle/cd/src/gradleTest/kotlin/com/avito/ci/BaseCiStepsPluginTest.kt
```diff
@@ -0,0 +1,147 @@
...
+import com.avito.android.plugin.artifactory.artifactoryUserParameterName
+import com.avito.instrumentation.instrumentationPluginId
+import com.avito.test.gradle.TestProjectGenerator
+import com.avito.test.gradle.dependencies.GradleDependency
+import com.avito.test.gradle.module.AndroidAppModule
+import com.avito.test.gradle.module.AndroidLibModule
+import com.avito.test.gradle.plugin.plugins
+import java.io.File
+
+internal abstract class BaseCiStepsPluginTest {
```
sergioboi, line 13: Extract because CiStepsTests has too many test methods.
So I decided to split it

subprojects/gradle/cd/src/gradleTest/kotlin/com/avito/ci/DeployCiStepTest.kt
```diff
@@ -0,0 +1,206 @@
...
+        generateProjectWithConfiguredCiSteps()
+        val result = runTask(":appA:release")
+        result.assertThat().apply {
+            tasksShouldNotBeTriggered(":appA:$deployTaskName")
+            buildSuccessful()
+        }
+    }
+
+    @Test
+    fun `upload proguard to crashlytics enabled - uploadCrashlyticsProguardFile executed`() {
```
sergioboi, line 149: New tests start there

subprojects/gradle/cd/src/main/kotlin/com/avito/ci/steps/deploy/ToGooglePlayDeploysTransformer.kt
```diff
@@ -0,0 +1,69 @@
...
+
+import com.avito.cd.AndroidArtifactType
+import com.avito.cd.CdBuildConfig
+import com.avito.ci.steps.ArtifactsConfiguration
+import com.avito.ci.steps.Output
+import com.avito.upload_to_googleplay.GooglePlayDeploy
+import java.io.File
+import java.util.Locale
+
+internal class ToGooglePlayDeploysTransformer(
```
sergioboi, line 11: Was extracted

subprojects/gradle/cd/src/main/kotlin/com/avito/ci/steps/deploy/UploadCrashlyticsProguardFileTasksProvider.kt
```diff
@@ -0,0 +1,35 @@
+package com.avito.ci.steps.deploy
+
+import com.avito.cd.BuildVariant
+import com.avito.cd.CdBuildConfig
+import org.gradle.api.Project
+import java.util.Locale
+
+internal class UploadCrashlyticsProguardFileTasksProvider {
```
sergioboi, line 8: The new

subprojects/gradle.properties
```diff
@@ -1,6 +1,6 @@
 # Gradle properties are not passed to included builds https://github.com/gradle/gradle/issues/2534
 kotlin.code.style=official
-org.gradle.jvmargs=-Xmx4g -XX:+UseGCOverheadLimit -XX:GCTimeLimit=10 -Dfile.encoding=UTF-8
+org.gradle.jvmargs=-Xmx6g -XX:+UseGCOverheadLimit -XX:GCTimeLimit=10 -Dfile.encoding=UTF-8
```
dsvoronin, line 3: is it clear what task causes OOM?

gradle.properties
dsvoronin: i don't know for sure.
Seems like it's the same problem that was with gradle/wrapper.
Using `./gradlew -p <some project>` starts build within this directory and only reads it's own `gradle.properties` and wrapper settings.

So i don't know how to tell gradle to always use same settings for all included builds inside this repository. 
Probably via symlinks?

Also we can't control IDE run configurations, it is always runs with single project context (e.g. `-p <project>` you run any tasks within)

dsvoronin: [Gradle slack discussion](https://gradle-community.slack.com/archives/CAHSN3LDN/p1628684164052900)

dsvoronin: We can also generate this files for every included build with task like 
https://github.com/avito-tech/avito-android/blob/1f104034705e99206f68f2fe1fe6a5c0bd9263a0/build.gradle.kts#L8-L24

it could be non invasive generation: only rewrite properties that we want to be the same, and left others in place

subprojects/gradle/cd/src/gradleTest/kotlin/com/avito/ci/BaseCiStepsPluginTest.kt
```diff
@@ -0,0 +1,147 @@
...
+        TestProjectGenerator(
+            plugins = plugins {
+                id("com.avito.android.impact")
+            },
+            modules = listOf(
+                AndroidAppModule(
+                    name = "appA",
+                    dependencies = setOf(GradleDependency.Safe.project(":shared")),
+                    plugins = plugins {
+                        classpathPlugin(
```
dsvoronin: i think this is misleading.
We use API that looks like `plugins` block and now injects `buildscript` block inside plugins. which is not possible
It looks wrong here.

I propose to add separate `buildscript` block to Module with it's own API, or just use plugins block with additional settings.gradle code:

```
pluginManagement {
  resolutionStrategy {
        eachPlugin {
            val pluginId = requested.id.id
            when {
                pluginId == "com.google.gms.google-services" -> useModule("com.google.gms:google-services:4.3.3")
                pluginId == "com.google.firebase.crashlytics" -> useModule("com.google.firebase:firebase-crashlytics-gradle:2.7.1")
            }
        }
    }
}
```

i personally prefer to just use second approach, because `apply plugin` is kinda "legacy" approach, and firebase plugins are among several left without support of "plugins" block

gradle.properties
dsvoronin: It could be done separately

subprojects/gradle/cd/src/gradleTest/kotlin/com/avito/ci/BaseCiStepsPluginTest.kt
sergioboi: The current behavior is a black box: you want to add the plugin like a classpath dependency and you don't need to know how this is implemented under the hood. I will try to convince you to believe in it.

The decision to use `pluginManagement` is great. But with the current impl of TestProjectGenerator we have problems.
We add plugins to some `module` but pluginmanagement is in `settings.gradle`. 
If we will configure two places, the DSL will be harder to write and understand.

Separate buildscript block:
pros - like a Gradle 
cons - the implicit intention of why we add that classpath stuff

If I don't convince you I will prefer the buildscript variant

subprojects/gradle.properties
sergioboi: I don't know how to find but could reproduce.
I need drop Gradle caches, build folder and sync the project as a first time

gradle.properties
sergioboi: I think we need a research task to try and make the decision

subprojects/gradle/cd/src/gradleTest/kotlin/com/avito/ci/BaseCiStepsPluginTest.kt
dsvoronin: We can rename our `plugins` block as a middle ground.

```
plugins = pluginConfiguration {
  withId("")
  applyWithBuildscript()
}
```
or something like that

to be more explicit that it's not a 1to1 plugins block definition

subprojects/gradle.properties
dsvoronin: OOM left hprof file as far as i remember
i think we can look at it's contents for clues

gradle.properties
dsvoronin: https://github.com/avito-tech/avito-android/pull/1168
sorry for rushing, but i just sit to play around this idea and seems like solved it

Please take a look

subprojects/gradle/cd/src/gradleTest/kotlin/com/avito/ci/BaseCiStepsPluginTest.kt
sergioboi: I changed part with applyWithBuildscript but decided to do not change a `plugin` part

dsvoronin (approved):

RuslanMingaliev (approved):

PR #1168, opened 2021-08-13
MBS-11759 implement common gradle properties generation

dsvoronin:
Use conf/common-gradle.properties to set properties which is common for all included builds

Then call `./gradlew :generateCommonProperties` to update gradle.properties files

edit: second commit adds same for gradle wrapper tasks


build-logic/gradle/src/main/kotlin/com/avito/android/GenerateCommonProperties.kt
```diff
@@ -0,0 +1,55 @@
...
+    abstract val commonPropertiesFile: RegularFileProperty
+
+    @get:OutputDirectories
+    abstract val projectDirs: SetProperty<Directory>
+
+    @TaskAction
+    fun doWork() {
+        val commonProperties = loadProperties(commonPropertiesFile.get().asFile)
+
+        projectDirs.get().forEach { projectDir ->
```
sergioboi, line 25: ProjectDirs - is Gradle fun?

build-logic/gradle/src/main/kotlin/convention.gradle-properties.gradle.kts
```diff
@@ -0,0 +1,15 @@
...
+tasks.register<GenerateCommonProperties>("generateCommonProperties") {
+    group = "Build setup"
+    description = "Generated common gradle.properties for all included builds"
+
+    commonPropertiesFile.set(layout.projectDirectory.file("conf/common-gradle.properties"))
+
+    projectDirs.set(
+        gradle.includedBuilds
+            .map { it.projectDir }
+            .map { layout.projectDirectory.dir(it.relativeTo(projectDir).path) }
```
sergioboi: Hard to read this line

build-logic/gradle/src/main/kotlin/com/avito/android/GenerateCommonProperties.kt
```diff
@@ -0,0 +1,55 @@
...
+import org.gradle.api.DefaultTask
+import org.gradle.api.file.Directory
+import org.gradle.api.file.RegularFileProperty
+import org.gradle.api.provider.SetProperty
+import org.gradle.api.tasks.InputFile
+import org.gradle.api.tasks.OutputDirectories
+import org.gradle.api.tasks.TaskAction
+import java.io.File
+
+abstract class GenerateCommonProperties : DefaultTask() {
```
sergioboi, line 13: - [ ] Add gradleTests please
- [ ] Add execution on each CI build
- [ ] Should we make task cacheble?

build-logic/gradle/src/main/kotlin/com/avito/android/GenerateCommonProperties.kt
```diff
@@ -0,0 +1,55 @@
...
+            val gradlePropertiesFile = File(projectDir.asFile, "gradle.properties")
+
+            val properties = if (gradlePropertiesFile.exists()) {
+                loadProperties(gradlePropertiesFile)
+            } else {
+                PropertiesConfiguration(gradlePropertiesFile)
+            }
+
+            commonProperties.keys.forEach { key ->
+                properties.setProperty(key, commonProperties.getProperty(key))
```
sergioboi, line 31: Proposal

- [ ] Add the comment `#overriden` on each overridden property
- [x] Add the comment `#from common file` on each property from common file
This will help to understand faster where I should add or update the property

sergioboi (approved):

build-logic/gradle/src/main/kotlin/com/avito/android/GenerateCommonProperties.kt
dsvoronin: nope, it's task property, task output specifically

build-logic/gradle/src/main/kotlin/com/avito/android/GenerateCommonProperties.kt
dsvoronin: Gradle tests should be done for sure, but can't use our TestProjectGenerator directly here
Options are:
 - add it as mavenCentral dependency
 - write test on pure GradleTestKit

[For future self]: don't forget that these tests won't be started on CI, need to wire it additionally

dsvoronin: Execution on every build is tricky.
I have to add additional gradle build for that, because gradle.properties changes will affect build only on second run.
This is the same situation as gradle wrapper task calling. 
So probably it is not a good idea

dsvoronin: Yep, we can make this task cacheable, but it won't be much profitable if task will be executed only manually (see previous comment)

build-logic/gradle/src/main/kotlin/com/avito/android/GenerateCommonProperties.kt
dsvoronin: `#from common file` is a good idea! Will do

Don't understand `#overriden` though

build-logic/gradle/src/main/kotlin/com/avito/android/GenerateCommonProperties.kt
sergioboi: > but can't use our TestProjectGenerator directly

This is a problem. I don't know what is better but prefer `add it as mavenCentral dependency`

> Execution on every build is tricky

But what If I add some properties and forget to regenerate them?
How will we check the correctness and integration of `gradle.properties` that should be generated?
I don't want to think that I shouldn't forget to regenerate something. I want CI to help me with this

build-logic/gradle/src/main/kotlin/com/avito/android/GenerateCommonProperties.kt
sergioboi: Sorry, my eyes was broken ))

build-logic/gradle/src/main/kotlin/com/avito/android/GenerateCommonProperties.kt
sergioboi: `#overriden` it's when for example you have property `org.gradle.xxx`
and the common file has a `org.gradle.xxx=1`
and some project `build.properties` has the same `org.gradle.xxx` prop
As I understand, after generation project's `build.properties` will have  `org.gradle.xxx` with value from common file

build-logic/gradle/src/main/kotlin/convention.gradle-properties.gradle.kts
dsvoronin: refactored this file, should be better

build-logic/gradle/src/main/kotlin/com/avito/android/GenerateCommonProperties.kt
dsvoronin: The only viable solution i see is to check if all properties have correct values from `common` properties file, fail on problems and ask to regenerate it.
Because it should be under version control and we don't have ability to commit and push via CI yet (and maybe for good)

build-logic/gradle/src/main/kotlin/com/avito/android/GenerateCommonProperties.kt
dsvoronin: yep, but it will be with `#from common file` comment now, why `#overriden`?

dsvoronin: image

looks like that. ok?

build-logic/gradle/src/main/kotlin/com/avito/android/GenerateCommonProperties.kt
```diff
@@ -0,0 +1,53 @@
...
+
+        projectDirs.get().forEach { projectDir ->
+
+            val gradlePropertiesFile = File(projectDir.asFile, "gradle.properties")
+
+            val properties = if (gradlePropertiesFile.exists()) {
+                PropertiesConfiguration(gradlePropertiesFile)
+            } else {
+                PropertiesConfiguration(gradlePropertiesFile)
+            }
```
eugene-krivobokov: Why are files in `if` and `else` the same?

```kotlin
PropertiesConfiguration(gradlePropertiesFile)
```

conf/common-gradle.properties
```diff
@@ -0,0 +1,17 @@
+# suppress inspection "UnusedProperty" for whole file
+# Gradle properties are not passed to included builds https://github.com/gradle/gradle/issues/2534
+# See build-logic/gradle for plugin that does common properties distribution
```
eugene-krivobokov, line 3: It's hard to dig into implementation details when I want to understand how to use it.
Example: https://github.com/avito-tech/avito-android/pull/1168/files#r689730969 

It'd be good to describe use cases in one place and refer to them.
Probable places: documentation, this properties file.

build-logic/gradle/src/main/kotlin/com/avito/android/CheckCommonProperties.kt
```diff
@@ -0,0 +1,114 @@
...
+                    val commonValue = commonProperties.getProperty(key)
+                    val value = properties.getProperty(key)
+                    if (value != commonValue) {
+                        inconsistencies.add(
+                            Inconsistency(
+                                projectPath = projectPath,
+                                reason = "different value for property: $key\n" +
+                                    "     expected: $commonValue\n" +
+                                    "     actual  : $value"
+                            )
```
eugene-krivobokov, line 63: Does it mean that we can't override common properties?

It doesn't look like a blocker, just want to clarify use cases.

build-logic/gradle/src/main/kotlin/com/avito/android/GenerateCommonProperties.kt
dsvoronin: Thanks, first implementation was using `java.util.Properties` API , but `commons-configuration` doesn't care

Fixed

build-logic/gradle/src/main/kotlin/com/avito/android/CheckCommonProperties.kt
dsvoronin: Yes, we can't
It shouldn't be a common property in that case

conf/common-gradle.properties
dsvoronin: added 'how to use' section to this file,
it will be better to add plugin docs, but after i extract it to separate plugin, which will be in another PR

PR #1178, opened 2021-08-23
MBS-11465 Do not fail gradle instrumentation task on test runner timeout

RuslanMingaliev:


subprojects/test-runner/client/src/main/kotlin/com/avito/runner/scheduler/runner/TestRunnerImpl.kt
```diff
@@ -40,7 +40,16 @@ internal class TestRunnerImpl(
     private val logger = loggerFactory.create<TestRunner>()
 
     override suspend fun runTests(tests: List<TestCase>): Result<TestRunnerResult> {
-        return withTimeout(executionTimeout.toMillis()) {
+        return withTimeoutOrDefault(
+            timeMillis = executionTimeout.toMillis(),
+            default = Result.Failure(
```
sergioboi, line 45: As I understand this code never be executed

subprojects/test-runner/client/src/main/kotlin/com/avito/runner/scheduler/runner/TestRunnerImpl.kt
```diff
@@ -97,13 +106,13 @@ internal class TestRunnerImpl(
                     )
 
                     testSuiteListener.onTestSuiteFinished()
-                    logger.info("Test run end successfully")
+                    logger.info("Test run finished successfully")
                     Result.Success(result)
                 } catch (e: TimeoutCancellationException) {
-                    logger.critical("Test run end with timeout", e)
+                    logger.critical("Test run finished with timeout", e)
                     Result.Failure(e)
```
sergioboi: Or we won't return this Failure when timeout occurs

subprojects/test-runner/client/src/test/kotlin/com/avito/runner/scheduler/runner/RunnerIntegrationTest.kt
```diff
@@ -579,9 +579,13 @@ internal class RunnerIntegrationTest {
         )
 
         devices.send(device)
-        assertThrows<TimeoutCancellationException> {
-            runner.runTests(tests)
-        }
+        val result = runner.runTests(tests)
+        result.fold({
+            fail("Test run finished successfully, but failure was expected")
```
sergioboi: Expect to see an AssertionError

subprojects/test-runner/client/src/main/kotlin/com/avito/runner/scheduler/runner/TestRunnerImpl.kt
RuslanMingaliev: Test shows the behavior changes

subprojects/test-runner/client/src/main/kotlin/com/avito/runner/scheduler/runner/TestRunnerImpl.kt
sergioboi: The current behavior isn't obvious. We return some Failure from the catch block but ignore it because of coroutine internals.
So I propose do not catch the TimeoutException and move logs to the creation of the default value

RuslanMingaliev: Yes, that's true, but I don't know how to write it better

RuslanMingaliev: > So I propose do not catch the TimeoutException and move logs to the creation of the default value

looks good to me

subprojects/test-runner/client/src/test/kotlin/com/avito/runner/scheduler/runner/RunnerIntegrationTest.kt
RuslanMingaliev: ☑

subprojects/test-runner/client/src/main/kotlin/com/avito/runner/scheduler/runner/TestRunnerImpl.kt
RuslanMingaliev: ☑

subprojects/test-runner/client/src/main/kotlin/com/avito/runner/scheduler/runner/TestRunnerImpl.kt
```diff
@@ -97,13 +106,12 @@ internal class TestRunnerImpl(
...
-                    logger.info("Test run end successfully")
+                    logger.info("Test run finished successfully")
                     Result.Success(result)
-                } catch (e: TimeoutCancellationException) {
-                    logger.critical("Test run end with timeout", e)
-                    Result.Failure(e)
                 } catch (e: Throwable) {
-                    logger.critical("Test run end with error", e)
+                    logger.critical("Test run finished with ${
+                        if (e is TimeoutCancellationException) "timeout" else "error"
```
sergioboi: The commodity is not an option because of poor readability
What Is a problem to make a separate log when Failure for Timeout is created?

sergioboi (approved):

PR #1185, opened 2021-08-26
MBS-11808 Fix ScreenChecks for a PopupWindow page object

eugene-krivobokov:
Added page object for PopupWindow.
It will reduce a boilerplate in clients (e.g. `TooltipElement` in avito repository).

### DIFF
--- a/subprojects/android-test/ui-testing-core-app/src/androidTest/kotlin/com/avito/android/ui/test/dialog/DialogsTest.kt
+++ b/subprojects/android-test/ui-testing-core-app/src/androidTest/kotlin/com/avito/android/ui/test/dialog/DialogsTest.kt
@@ -1,5 +1,6 @@
 package com.avito.android.ui.test.dialog
 
+import androidx.test.espresso.NoMatchingRootException
 import com.avito.android.test.app.core.screenRule
 import com.avito.android.ui.DialogsActivity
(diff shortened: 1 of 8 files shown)


subprojects/android-test/ui-testing-core-app/src/androidTest/kotlin/com/avito/android/ui/test/dialog/DialogsTest.kt
```diff
@@ -36,4 +36,12 @@ class DialogsTest {
             Screen.dialogsScreen.label.checks.isDisplayed()
         }
     }
+
+    @Test
+    fun matcher_with_popup_window__success__popup_is_open() {
```
sergioboi: It would be great to have a negative scenario to prevent false positive

sergioboi (approved):

subprojects/android-test/ui-testing-core-app/src/androidTest/kotlin/com/avito/android/ui/test/dialog/DialogsTest.kt
eugene-krivobokov: Agree. 
Added negative cases for non-default root matchers.

PR #1187, opened 2021-08-27
MBS-11793 Fix flaky Device.rotate() in Android 11

eugene-krivobokov:
Tests: [RotateOrientationTest](https://github.com/avito-tech/avito-android/blob/9d0d9d83e33773e5d0804035b30c2321ee2b24aa/subprojects/android-test/ui-testing-core-app/src/androidTest/kotlin/com/avito/android/ui/test/RotateOrientationTest.kt)

Checked by multiple runs in avito because it's flaky and not reproducible in all tests:

- [Before](http://links.k.avito.ru/2R)
- [After](http://links.k.avito.ru/bU)


subprojects/android-test/ui-testing-core/src/main/kotlin/com/avito/android/test/Device.kt
```diff
@@ -38,7 +40,13 @@ object Device {
      * Changes device orientation between portrait and landscape
      */
     fun rotate() {
-        onView(ViewMatchers.isRoot()).perform(OrientationChangeAction.toggle())
+        if (Build.VERSION.SDK_INT >= 30) {
+            // ViewAction with default root matcher is flaky
+            // It can pick wrong window in case of opened dialog
+            InstrumentationOrientationChangeAction().rotate()
+        } else {
+            onView(ViewMatchers.isRoot()).perform(OrientationChangeAction.toggle())
```
eugene-krivobokov, line 48: I leave the previous implementation until stabilizing the new one.
We may delete it after enabling tests on Android 11.

dsvoronin (approved):

subprojects/android-test/ui-testing-core/src/main/kotlin/com/avito/android/test/action/InstrumentationOrientationChangeAction.kt
```diff
@@ -0,0 +1,29 @@
...
+import android.content.res.Configuration
+import androidx.test.platform.app.InstrumentationRegistry
+import com.avito.android.test.util.getCurrentActivity
+
+internal class InstrumentationOrientationChangeAction {
+
+    fun rotate() {
+        val instrumentation = InstrumentationRegistry.getInstrumentation()
+
+        instrumentation.runOnMainSync {
```
sergioboi, line 14: Could lead to timeouts when there is no Activity

sergioboi (approved):

subprojects/android-test/ui-testing-core/src/main/kotlin/com/avito/android/test/action/InstrumentationOrientationChangeAction.kt
eugene-krivobokov: Without Activity, it fails fastly with an error "No activities in RESUMED state found".
The behavior is the same as before. It also fails for this reason.

PR #1188, opened 2021-08-27
MBS-11774 Fix Percent.kt

sergioboi:
fix equals with BigDecimal and precision
fix NaN and Infinity Percent.kt creation

### DIFF
--- a/subprojects/common/math/src/main/kotlin/com/avito/math/Percent.kt
+++ b/subprojects/common/math/src/main/kotlin/com/avito/math/Percent.kt
@@ -5,10 +5,10 @@ import java.math.MathContext
 import java.text.DecimalFormat
 
 @Suppress("MagicNumber")
-public fun Int.percentOf(sum: Int): Percent = BigDecimalTwoDigitsPrecisionPercent.create(toFloat() / sum * 100)
+public fun Int.percentOf(sum: Int): Percent = FourDigitsPercent.create(toFloat() / sum * 100)
 
 @Suppress("MagicNumber")
-public fun Long.percentOf(sum: Long): Percent = BigDecimalTwoDigitsPrecisionPercent.create(toFloat() / sum * 100)
+public fun Long.percentOf(sum: Long): Percent = FourDigitsPercent.create(toFloat() / sum * 100)
 
 /**
  * convert 0.0 - 1.0 representation to [Percent]
@@ -16,7 +16,7 @@ public fun Long.percentOf(sum: Long): Percent = BigDecimalTwoDigitsPrecisionPerc
 @Suppress("MagicNumber")
 public fun Double.fromZeroToOnePercent(): Percent {
     require(this in 0.0..1.0) { "Trying to convert $this to percents; should be in [0.0..1.0] range" }
-    return BigDecimalTwoDigitsPrecisionPercent.create(this * 100)
+    return FourDigitsPercent.create(this * 100)
 }
 
 /**
@@ -25,32 +25,32 @@ public fun Double.fromZeroToOnePercent(): Percent {
 @Suppress("MagicNumber")
 public fun Float.fromZeroToOnePercent(): Percent {
     require(this in 0.0..1.0) { "Trying to convert $this to percents; should be in [0.0..1.0] range" }
-    return BigDecimalTwoDigitsPrecisionPercent.create(this * 100)
+    return FourDigitsPercent.create(this * 100)
 }
 
 public fun Int.fromZeroToHundredPercent(): Percent {
     require(this in 0..100) { "Trying to convert $this to percents; should be in [0..100] range" }
-    return BigDecimalTwoDigitsPrecisionPercent.create(this)
+    return FourDigitsPercent.create(this)
 }
 
 public interface Percent {
 
-    public fun roundToTwoDigitsPrecision(): String
+    public fun formatAsFourDigitsPrecision(): String
 
     public fun roundToInt(): Int
 
     public fun roundToLong(): Long
 
     public companion object {
-        public val ZERO: Percent = BigDecimalTwoDigitsPrecisionPercent.create(0)
+        public val ZERO: Percent = FourDigitsPercent.create(0)
     }
 }
 
-internal data class BigDecimalTwoDigitsPrecisionPercent private constructor(val value: BigDecimal) : Percent {
+internal data class FourDigitsPercent private constructor(val value: BigDecimal) : Percent {
 
     private val formatted: String by lazy { DecimalFormat("0.##'%'").format(value) }
 
-    override fun roundToTwoDigitsPrecision(): String = formatted
+    override fun formatAsFourDigitsPrecision(): String = formatted
 
     override fun roundToInt(): Int = value.toInt()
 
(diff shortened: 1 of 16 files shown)


subprojects/common/math/src/main/kotlin/com/avito/math/Percent.kt
```diff
@@ -23,32 +25,44 @@ public fun Double.fromZeroToOnePercent(): Percent {
...
 
-    public fun toInt(): Int
+    public fun roundToInt(): Int
 
-    public fun toLong(): Long
+    public fun roundToLong(): Long
 }
 
-internal data class NumberPercent(val value: Number) : Percent {
+internal data class BigDecimalTwoDigitsPrecisionPercent private constructor(val value: BigDecimal) : Percent {
```
sergioboi: With Number equals doesn't work

subprojects/common/math/src/main/kotlin/com/avito/math/Percent.kt
```diff
@@ -23,32 +25,44 @@ public fun Double.fromZeroToOnePercent(): Percent {
...
-    override fun toLong(): Long = value.toLong()
+    override fun roundToLong(): Long = value.toLong()
 
-    override fun toString(): String = twoDigitsString()
+    override fun toString(): String = formatted
+
+    companion object {
+        fun create(number: Number): Percent {
+            val mathContext = MathContext(4)
+            return when (number) {
```
sergioboi, line 64: This is the only way to make `equals` works as expected

subprojects/common/math/src/main/kotlin/com/avito/math/Percent.kt
```diff
@@ -23,32 +25,44 @@ public fun Double.fromZeroToOnePercent(): Percent {
...
+    return BigDecimalTwoDigitsPrecisionPercent.create(this)
 }
 
 public interface Percent {
 
-    public fun twoDigitsString(): String
+    public fun roundToTwoDigitsPrecision(): String
 
-    public fun toInt(): Int
+    public fun roundToInt(): Int
```
sergioboi, line 40: prefer to realize that under the hood your percents will be rounded

subprojects/gradle/test-summary/src/main/kotlin/com/avito/test/summary/model/CrossDeviceSuite.kt
```diff
@@ -25,25 +26,34 @@ public data class CrossDeviceSuite(val crossDeviceRuns: List<CrossDeviceRunTest>
...
 
-    val percentFailedOnSomeDevicesOfAutomated: Percent = failedOnSomeDevicesCount.percentOf(automatedCount)
+    val percentLostOnSomeDevicesOfAutomated: Percent
+    val percentSuccessOfAutomated: Percent
+    val percentSkippedOnAllDevicesOfAutomated: Percent
+    val percentFailedOnAllDevicesOfAutomated: Percent
+    val percentFailedOnSomeDevicesOfAutomated: Percent
+
+    init {
+        if (automatedCount == 0) {
```
sergioboi, line 42: Found out the bug.
Before `NumberPercent.twoDigits` will return "NaN" if `value` is NaN Number
After creating Percent from NaN will throw an exception

subprojects/common/math/src/main/kotlin/com/avito/math/Percent.kt
```diff
@@ -23,32 +25,48 @@ public fun Double.fromZeroToOnePercent(): Percent {
...
 public fun Int.fromZeroToHundredPercent(): Percent {
     require(this in 0..100) { "Trying to convert $this to percents; should be in [0..100] range" }
-    return NumberPercent(this)
+    return BigDecimalTwoDigitsPrecisionPercent.create(this)
 }
 
 public interface Percent {
 
-    public fun twoDigitsString(): String
+    public fun roundToTwoDigitsPrecision(): String
```
eugene-krivobokov: Methods `roundTo...` are not simmetrical in return types:

rountToInt() --> Int
roundToTwoDigitsPrecision() --> String

Maybe this should be `format...`

subprojects/common/math/src/main/kotlin/com/avito/math/Percent.kt
eugene-krivobokov: Proposal for renaming:
`BigDecimalTwoDigitsPrecisionPercent` --> `TwoDigitsPrecisionPercent` or `TwoDigitsPercent`

Because from the client perspective BigDecimal is implementation details that are already hidden by API.

subprojects/gradle/test-summary/src/main/kotlin/com/avito/test/summary/model/CrossDeviceSuite.kt
```diff
@@ -25,25 +26,34 @@ public data class CrossDeviceSuite(val crossDeviceRuns: List<CrossDeviceRunTest>
...
-
     val failedOnSomeDevicesCount: Int
         get() = consistentData.count { it.status is CrossDeviceStatus.FailedOnSomeDevices }
 
-    val percentFailedOnSomeDevicesOfAutomated: Percent = failedOnSomeDevicesCount.percentOf(automatedCount)
+    val percentLostOnSomeDevicesOfAutomated: Percent
+    val percentSuccessOfAutomated: Percent
+    val percentSkippedOnAllDevicesOfAutomated: Percent
+    val percentFailedOnAllDevicesOfAutomated: Percent
+    val percentFailedOnSomeDevicesOfAutomated: Percent
```
eugene-krivobokov, line 39: It looks like the Hungarian notation. I think it's excessive.

subprojects/gradle/test-summary/src/main/kotlin/com/avito/test/summary/model/CrossDeviceSuite.kt
```diff
@@ -25,25 +26,34 @@ public data class CrossDeviceSuite(val crossDeviceRuns: List<CrossDeviceRunTest>
...
+
+    init {
+        if (automatedCount == 0) {
+            val zeroPercents = 0F.fromZeroToOnePercent()
+            percentSuccessOfAutomated = zeroPercents
```
eugene-krivobokov, line 44: Can `percentSuccessOfAutomated` and other `percent...` have non-zero values in this case?
If not it may be good to check this precondition explicitly.
Otherwise, it's non-obvious why N is 0% from 0.

eugene-krivobokov (approved):

subprojects/common/math/src/main/kotlin/com/avito/math/Percent.kt
sergioboi: To be honest, the correct name will be FourDigitsPercent

subprojects/common/math/src/main/kotlin/com/avito/math/Percent.kt
sergioboi: Agree

subprojects/gradle/test-summary/src/main/kotlin/com/avito/test/summary/model/CrossDeviceSuite.kt
sergioboi: Agree

subprojects/gradle/test-summary/src/main/kotlin/com/avito/test/summary/model/CrossDeviceSuite.kt
sergioboi: As I understand it's normal case when you have zero automated tests and two manual tests.

But it's incorrect to count manual to automated. Better is count manual comaparing with total. So, I could rewrite percents logic

PR #1189, opened 2021-08-27
MBS-11822 fix local report link

dsvoronin:
look at deviceName, not buildId to determine if it's a local run

### DIFF
--- a/subprojects/test-runner/test-inhouse-runner/src/main/kotlin/com/avito/android/runner/TestRunEnvironment.kt
+++ b/subprojects/test-runner/test-inhouse-runner/src/main/kotlin/com/avito/android/runner/TestRunEnvironment.kt
@@ -120,8 +120,8 @@ fun parseEnvironment(
 }
 
 internal fun parseReportDestination(argumentsProvider: ArgsProvider): ReportDestination {
-    val buildId = argumentsProvider.getMandatoryArgument("teamcityBuildId").toInt()
-    return if (buildId == TestRunEnvironment.LOCAL_STUDIO_RUN_ID) {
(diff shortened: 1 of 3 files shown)


subprojects/test-runner/test-inhouse-runner/src/main/kotlin/com/avito/android/runner/TestRunEnvironment.kt
```diff
@@ -120,8 +120,8 @@ fun parseEnvironment(
 }
 
 internal fun parseReportDestination(argumentsProvider: ArgsProvider): ReportDestination {
-    val buildId = argumentsProvider.getMandatoryArgument("teamcityBuildId").toInt()
-    return if (buildId == TestRunEnvironment.LOCAL_STUDIO_RUN_ID) {
+    val deviceName = argumentsProvider.getMandatoryArgument("deviceName")
+    return if (deviceName.equals("local", ignoreCase = true)) {
```
eugene-krivobokov, line 124: `deviceName=local` implies sending a report from the test runner.
It's a non-obvious relationship.
How about `avito.build`?

I'm ok with it as a fix, but for the future, I'd like to see a more straightforward solution.

eugene-krivobokov (approved):

subprojects/test-runner/test-inhouse-runner/src/main/kotlin/com/avito/android/runner/TestRunEnvironment.kt
dsvoronin: i'm okay with some build environment property

PR #1198, opened 2021-09-05
MBS-9085 [Configuration Cache] remove logger factory methods access to project 

dsvoronin:


subprojects/gradle/module-dependencies-graph/src/main/kotlin/com/avito/module/dependencies/ModuleDependenciesGraphPlugin.kt
```diff
@@ -11,8 +12,24 @@ public class ModuleDependenciesGraphPlugin : Plugin<Project> {
...
+                )
+            )
+        }
         target.tasks.register("collectAppsMetrics", CollectAppsMetricsTask::class.java) {
+            it.loggerFactory.set(
+                GradleLoggerFactory.fromProject(
+                    project = target,
+                    pluginName = "ModuleDependenciesGraphPlugin",
+                    taskName = "CollectAppsMetricsTask"
+                )
```
eugene-krivobokov, line 30: I propose to avoid hardcode by supporting this frequent use case in factory:

```kotlin
GradleLoggerFactory.fromProject(project, plugin, task) // <-- no strings here
```

eugene-krivobokov (approved):

subprojects/gradle/module-dependencies-graph/src/main/kotlin/com/avito/module/dependencies/ModuleDependenciesGraphPlugin.kt
dsvoronin: Yep, fixed

PR #1199, opened 2021-09-05
MBS-9085 [Configuration cache] sentry support

dsvoronin:
+ remove useless UnstableApiUsage suppress annotations


eugene-krivobokov (approved):

subprojects/logger/gradle-logger/src/main/kotlin/com/avito/logger/GradleLoggerFactory.kt
```diff
@@ -146,7 +146,6 @@ public class GradleLoggerFactory(
             verboseMode = getVerbosity(project)?.let { VerboseMode(it, doPrintStackTrace(project)) }
         )
 
-        @Suppress("UnstableApiUsage")
```
eugene-krivobokov, line 149: How did you find all obsolete suppressed cases?

dsvoronin: Just removed all of them )

PR #1204, opened 2021-09-10
MBS-11769 add umbrella tasks

dsvoronin:
samples instrumentation tasks was in separate teamcity build, now it's a part of :checkAll / :build tasks, so executed in CI single build

move documentation lint before gradle build

todo:
 - remove all `-p subprojects`
 - compileAll for included build is now stub task


build.gradle.kts
```diff
@@ -15,10 +16,55 @@ tasks.register<Exec>("installGitHooks") {
...
+    dependsOn(
+        gradle.includedBuilds
+            .filter { !it.name.startsWith("build-logic") }
+            .map { it.task(":compileAll") }
+    )
+}
+
+val assembleAll = tasks.register("assembleAll") {
+    group = taskGroup
+    description = "Compile all code of all components"
```
eugene-krivobokov:
```suggestion
    description = "Assemble all components"
```

build.gradle.kts
```diff
@@ -15,10 +16,55 @@ tasks.register<Exec>("installGitHooks") {
...
+        gradle.includedBuilds
+            .filter { !it.name.startsWith("build-logic") }
+            .map { it.task(":assembleAll") }
+    )
+}
+
+tasks.register("build") {
+    group = taskGroup
+    description = "Run all tests, build (without publishing)"
+    dependsOn(assembleAll, compileAll, checkAll)
```
eugene-krivobokov:
```suggestion
    dependsOn(assembleAll, checkAll)
```
Do we have cases when assemble doesn't include compile?

build.gradle.kts
```diff
@@ -15,10 +16,55 @@ tasks.register<Exec>("installGitHooks") {
...
+    dependsOn(
+        gradle.includedBuilds
+            .filter { !it.name.startsWith("build-logic") }
+            .map { it.task(":assembleAll") }
+    )
+}
+
+tasks.register("build") {
+    group = taskGroup
+    description = "Run all tests, build (without publishing)"
```
eugene-krivobokov:
```suggestion
    description = "Build and run all tests (without publishing)"
```
To reflect real execution order

build.gradle.kts
```diff
@@ -15,10 +16,55 @@ tasks.register<Exec>("installGitHooks") {
...
+
+// Register lifecycle tasks in this umbrella build.
+// A user/CI usually only needs these.
+val checkAll = tasks.register("checkAll") {
+    group = taskGroup
+    description = "Run all tests and create code coverage report"
+
+    dependsOn(
+        gradle.includedBuilds
+            .filter { !it.name.startsWith("build-logic") }
```
eugene-krivobokov: Can we hide this duplicated knowledge about "build-logic"?
Maybe reusable function or make fake empty tasks inside build-logic.

build-logic/gradle/src/main/kotlin/convention.lifecycle.gradle.kts
```diff
@@ -0,0 +1,43 @@
+val taskGroup = "Avito android build"
```
eugene-krivobokov:
```suggestion
val taskGroup = "Avito Android build"
```
Just to be consistent as in [official guidelines](https://developer.android.com/distribute/marketing-tools/brand-guidelines#android_in_text)

build-logic/gradle/src/main/kotlin/convention.lifecycle.gradle.kts
```diff
@@ -0,0 +1,43 @@
...
+
+tasks.named<TaskReportTask>("tasks") {
+    displayGroup = taskGroup
```
eugene-krivobokov: How could I see tasks from others groups?

build-logic/gradle/src/main/kotlin/convention.lifecycle.gradle.kts
```diff
@@ -0,0 +1,43 @@
...
+}
+
+tasks.register("assembleAll") {
+    group = taskGroup
+    description = "Compile all code of the '${project.name}' component"
```
eugene-krivobokov:
```suggestion
    description = "Compile all code of the '${project.path}' component"
```
Do we need to worry about probable duplicates here?

build-logic/gradle/src/main/kotlin/convention.lifecycle.gradle.kts
```diff
@@ -0,0 +1,43 @@
...
+    group = taskGroup
+    description = "Compile all code and run all tests of the '${project.name}' component"
+    dependsOn(
+        subprojects
+            .filter { !it.isJustADir() }
+            .map { "${it.path}:check" }
+    )
+}
+
+fun Project.isJustADir(): Boolean {
```
eugene-krivobokov:
```suggestion
fun Project.isPlainDir(): Boolean
```
```suggestion
fun Project.isModule(): Boolean
```
Just as options

build-logic/gradle/src/main/kotlin/convention.lifecycle.gradle.kts
dsvoronin: In description probably not, but i will use path anyway

build-logic/gradle/src/main/kotlin/convention.lifecycle.gradle.kts
dsvoronin: I don't get why it's here 😆 
Just copy pasted from an example

removed

build.gradle.kts
dsvoronin: compileAll also includes compiling unit and android tests, which should be a part of checkAll

build.gradle.kts
dsvoronin: i made tasks in build-logic projects

PR #1205, opened 2021-09-11
MBS-11661 Fix waitForLauncher in tests against Android 11 targetSdk apps

eugene-krivobokov:


subprojects/android-test/ui-testing-core/src/main/kotlin/com/avito/android/test/Device.kt
```diff
@@ -82,19 +83,30 @@ object Device {
     }
 
     fun waitForLauncher(timeout: Long = UITestConfig.activityLaunchTimeoutMilliseconds) {
```
eugene-krivobokov, line 86: In app with targetSdk 30 in Android 11 it was broken - always timeout.

The reason is in incorrect `launcherPackageName`:
```kotlin
public String getLauncherPackageName() {
        Intent intent = new Intent(Intent.ACTION_MAIN);
        intent.addCategory(Intent.CATEGORY_HOME);
        PackageManager pm = mInstrumentation.getContext().getPackageManager();
        ResolveInfo resolveInfo = pm.resolveActivity(intent, PackageManager.MATCH_DEFAULT_ONLY); <---
        return resolveInfo.activityInfo.packageName;
}
```
It returns fallback intent to `com.android.settings`, but the real launcher is hidden:

```
system_process I/AppsFilter: interaction: PackageSetting{9b6897d com.avito.android..../10121} -> PackageSetting{44ad072 com.android.launcher3/10109} BLOCKED
```

I've commented why other options are worse for a target application.

subprojects/test-runner/test-inhouse-runner/src/main/kotlin/com/avito/android/runner/PlayServices.kt
```diff
@@ -1,30 +1,14 @@
...
 
 import android.content.Context
-import android.content.pm.PackageManager
-import com.avito.logger.Logger
 import com.google.android.gms.common.ConnectionResult
 import com.google.android.gms.common.GoogleApiAvailability
 
-internal fun Context.checkPlayServices(logger: Logger) {
-    logger.debug(
-        "Required play services version: $playServicesMetaDataVersion, on device: $playServicesOnDeviceVersion"
```
eugene-krivobokov, line 11: 1. Don't need this log when everything is ok.
In case of error, it's also useless. We can't choose the version, only update for the latest.

2. Logger is broken in infra tests here due to MBS-11782.

sergioboi (approved):

subprojects/android-test/ui-testing-core/src/main/kotlin/com/avito/android/test/UITestConfig.kt
```diff
@@ -53,6 +54,13 @@ object UITestConfig {
         AssertionError::class.java
     )
 
+    /**
+     * Pattern for launcher package in Android 11+ devices.
+     * Hardcoded values due to package visibility restrictions:
+     * https://developer.android.com/training/package-visibility/declaring
+     */
+    var deviceLauncherPackage: Pattern = Pattern.compile(".+launcher.*")
```
eugene-krivobokov, line 62: I've extracted this hardcode to config.
It may be useful to fix it in local builds before fixing in the test library.

PR #1207, opened 2021-09-13
MBS-11882 Fix ToastRule in Android 11 target

eugene-krivobokov:
Breaking change in toasts for apps with targetSdk 30 in Android 11:
No view for toasts created by `Toast.makeText`.

https://developer.android.com/about/versions/11/behavior-changes-11#toasts


subprojects/android-test/toast-rule/src/main/kotlin/com/avito/android/rule/internal/ProxyToastChecks.kt
```diff
@@ -0,0 +1,110 @@
...
+import org.hamcrest.Matchers.empty
+import org.hamcrest.Matchers.not
+import org.hamcrest.TypeSafeMatcher
+
+internal class ProxyToastChecks(
+    private val proxyToast: MockProxyToast,
+    private val textAccessor: ToastTextAccessor,
+) : ToastChecks {
+
+    private val longToastDuration = 3500L
```
dsvoronin: Any chance to get value from android resources?

subprojects/android-test/toast-rule/src/main/kotlin/com/avito/android/rule/internal/ProxyToastChecks.kt
```diff
@@ -0,0 +1,110 @@
...
+
+    override fun toastDisplayedWithText(text: String) {
+        internalToastDisplayedWithText(text)
+    }
+
+    override fun toastDisplayedWithText(@StringRes resId: Int) {
+        val toast = firstToastOrThrow()
+        val context = toast.getFieldByReflection("mContext") as Context
```
sergioboi: Are you sure we should use reflection here?
Could we simply use context from `Instrumentation`?

subprojects/android-test/toast-rule/src/main/kotlin/com/avito/android/rule/internal/ProxyToastChecks.kt
```diff
@@ -0,0 +1,110 @@
...
+            MatcherAssert.assertThat(
+                "There are no toasts shown",
+                proxyToast.shownToasts,
+                not(empty())
+            )
+            proxyToast.shownToasts.first()
+        }
+    }
+
+    private fun withView(viewMatcher: Matcher<View>): Matcher<Toast> = object : TypeSafeMatcher<Toast>() {
```
sergioboi: Why do you think that we don't need a class?

subprojects/android-test/toast-rule/src/main/kotlin/com/avito/android/rule/internal/ToastTextAccessor.kt
```diff
@@ -0,0 +1,59 @@
...
+internal class InAppToastTextAccessor : ToastTextAccessor {
+
+    override fun text(toast: Toast): String? {
+        @Suppress("DEPRECATION")
+        val view = requireNotNull(toast.view) {
+            "Toast.view must be present in $toast"
+        }
+        val messageView: TextView? = view.findViewById(android.R.id.message)
+        val isCustomView = messageView == null
+        return if (isCustomView) null else messageView!!.text.toString()
```
sergioboi: force unwrap should be skipped

subprojects/android-test/toast-rule/src/main/kotlin/com/avito/android/rule/internal/ToastTextAccessor.kt
```diff
@@ -0,0 +1,59 @@
...
+    }
+}
+
+internal class SystemUIToastTextAccessor : ToastTextAccessor {
```
sergioboi, line 60: I need a comment about how Toast draws in 30 and upper APIs

subprojects/android-test/toast-rule/src/main/kotlin/com/avito/android/rule/internal/ToastTextAccessor.kt
```diff
@@ -0,0 +1,59 @@
...
+
+    override fun text(toast: Toast): String? {
```
sergioboi, line 62: Let's discuss a behavior.  We are trying to express the complex state via nullabillity

sergioboi (approved):

subprojects/android-test/toast-rule/src/main/kotlin/com/avito/android/rule/internal/ProxyToastChecks.kt
eugene-krivobokov: Found it in a hidden class. 
We could access it only through reflection, but it will be version-specific.
I think it's too much for this knowledge.

subprojects/android-test/toast-rule/src/main/kotlin/com/avito/android/rule/internal/ProxyToastChecks.kt
eugene-krivobokov: Thanks, didn't think of it. 
Fixed

To be correct, application context is not a complete replacement. There is a case of configuration-specific resources. 
But it looks like a corner case. We can live without it.

subprojects/android-test/toast-rule/src/main/kotlin/com/avito/android/rule/internal/ProxyToastChecks.kt
eugene-krivobokov: Fixed, introduced a class

subprojects/android-test/toast-rule/src/main/kotlin/com/avito/android/rule/internal/ToastTextAccessor.kt
eugene-krivobokov: Fixed

subprojects/android-test/toast-rule/src/main/kotlin/com/avito/android/rule/internal/ToastTextAccessor.kt
eugene-krivobokov: Added

subprojects/android-test/toast-rule/src/main/kotlin/com/avito/android/rule/internal/ToastTextAccessor.kt
eugene-krivobokov: Added an explicit check for the existence of the field.
Besides that, it looks ok. The field is nullable in the case of a custom view.

PR #1208, opened 2021-09-13
MBS-11769 optimize gradle tests run

dsvoronin:
## limit gradle workers to 4 locally

Why 4? Empirically doesn't cause hangs on my pc, but still have some parallelism for build.
As far as i know we have values between 12 and 32 in our team (6 cpu + hyperthreading on mac's minimum)
Could tune this value higher, but should be done with mac

## more memory for gradle daemons in tests to avoid OOM metaspace

Previous tweak affected only test process, but OOM was in gradle daemon, spawned for test

### DIFF
--- a/build-logic-settings/gradle.properties
+++ b/build-logic-settings/gradle.properties
@@ -20,3 +20,6 @@ org.gradle.caching=true
 # from common properties
 # https://github.com/gradle/gradle/issues/4823
 org.gradle.configureondemand=false
+# from common properties
+# Limited locally, because unlimited workers will hang pc on parallel gradleTest execution
+org.gradle.workers.max=4
--- a/samples/gradle.properties
+++ b/samples/gradle.properties
@@ -24,3 +24,6 @@ avito.logging.verbosity=WARNING
 avito.build=local
 #
 android.useAndroidX=true
+# from common properties
+# Limited locally, because unlimited workers will hang pc on parallel gradleTest execution
+org.gradle.workers.max=4
--- a/build-logic/gradle.properties
+++ b/build-logic/gradle.properties
@@ -20,3 +20,6 @@ org.gradle.caching=true
 # from common properties
 # https://github.com/gradle/gradle/issues/4823
 org.gradle.configureondemand=false
+# from common properties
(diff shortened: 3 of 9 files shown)


build-logic-settings/gradle.properties
```diff
@@ -20,3 +20,6 @@ org.gradle.caching=true
 # from common properties
 # https://github.com/gradle/gradle/issues/4823
 org.gradle.configureondemand=false
+# from common properties
+# Limited locally, because unlimited workers will hang pc on parallel gradleTest execution
+org.gradle.workers.max=4
```
sergioboi, line 25: FYI i override this in my local gradle.properties
And I don't sure that you set the optimal value
I want to replace that or add a comment on how to choose the correct number for my local machine

samples/gradle.properties
```diff
@@ -24,3 +24,6 @@ avito.logging.verbosity=WARNING
 avito.build=local
 #
 android.useAndroidX=true
+# from common properties
```
sergioboi, line 27: as I remember we discuss to add `generated when a property generated

dsvoronin: https://github.com/avito-tech/avito-android/pull/1168#discussion_r689426112

build-logic-settings/gradle.properties
dsvoronin: yep, anyone can override it locally, my goal was to set a viable default
default value wasn't set, and is max cpu by default; which leads to pc hang on gradleTest task, because of amount of gradle processes spawned in parallel

I don't know how to choose a proper value yet

Will re-test on macbook

sergioboi (approved):

build-logic-settings/gradle.properties
dsvoronin: macbook with 4 workers:
./gradlew -p subprojects gradleTest
BUILD SUCCESSFUL in 9m 25s

without noticeable ide lags while building

eugene-krivobokov:
Will you do the same in the avito project?

eugene-krivobokov:
> Will you do the same in the avito project?

MBS-11929

PR #1212, opened 2021-09-15
MBS-11872 Fix generics for custom ScreenChecks

eugene-krivobokov:


subprojects/android-test/ui-testing-core/src/main/kotlin/com/avito/android/test/page_object/DialogScreen.kt
```diff
@@ -28,10 +28,10 @@ abstract class DialogScreen(
 
     val rootElement = element<ViewElement>()
 
-    open class DialogScreenChecks(
-        screen: DialogScreen,
+    open class DialogScreenChecks<T : DialogScreen>(
```
eugene-krivobokov, line 31: Couldn't extend this class

PR #1218, opened 2021-09-15
MBS-11914 update radiography

dsvoronin:
https://github.com/square/radiography/blob/main/CHANGELOG.md#version-241

bugfix: avoid main thread check crashes

Stop throwing when not on main thread. While Curtains APIs should be used from the main thread, there are a number edge cases as well as bugs in consuming SDKs / apps and the Android Framework SDKs that can trigger calls from the wrong thread, and in many cases there isn't much developers can do, so Curtains is relaxing the constraint and doing a best effort approach.


eugene-krivobokov (approved):

eugene-krivobokov:
Did it work?
It seems broken: MBS-12007

PR #1220, opened 2021-09-16
MBS-11916 fix some dependencies was not downloaded through proxy

dsvoronin:
Checked by running with `--refresh-dependencies` and looking to scan / network page

Problem was combined:
not only i flushed all the caches
jcenter and gradle plugins portal based on it's repos degraded: https://status.gradle.com/


subprojects/build.gradle.kts
```diff
@@ -2,7 +2,7 @@ plugins {
     id("convention.lifecycle")
     // accessing version catalog here is blocked by IDE false-postive error
     // https://youtrack.jetbrains.com/issue/KTIJ-19369
-    id("com.autonomousapps.dependency-analysis") version "0.74.0"
+    id("com.autonomousapps.dependency-analysis") version "0.77.0"
```
eugene-krivobokov, line 5: Is it a necessary update for the fix?

eugene-krivobokov (approved):

subprojects/settings.gradle.kts
```diff
@@ -1,5 +1,76 @@
 enableFeaturePreview("VERSION_CATALOGS")
-enableFeaturePreview("TYPESAFE_PROJECT_ACCESSORS")
+
+pluginManagement {
+    includeBuild("../build-logic-settings")
+
+    val artifactoryUrl: String? by settings
+
+    repositories {
+        maven {
```
sergioboi, line 9: Could we make an extension function `mavenWithProxy` that will have 3 parameters: title, URL, proxy URL
To reduce duplications

sergioboi (approved):

subprojects/build.gradle.kts
dsvoronin: Plugin was failing with weird stacktrace until i updated it

subprojects/settings.gradle.kts
dsvoronin: looks like not in pluginManagement

dsvoronin:
https://github.com/avito-tech/avito-android/pull/1224

PR #1233, opened 2021-09-21
MBS-10164 Clarify error message for uniqueRClasses build check

eugene-krivobokov:


subprojects/gradle/build-checks/src/main/kotlin/com/avito/android/build_checks/internal/unique_r/UniqueRClassesTask.kt
```diff
@@ -69,9 +70,12 @@ internal abstract class UniqueRClassesTask @Inject constructor(
...
                     AndroidAppChecksExtension::uniqueRClasses,
                     """
-                    Application ${project.path} has modules with the same package: $duplicates.
+                    Application ${project.path} has dependencies with the same package in AndroidManifest.xml: $duplicates.
                     It leads to unexpected resource overriding.
-                    Please, make packages unique.
+                    
+                    How to fix: 
+                    Make packages in manifests unique.
+                    If this is impossible because of clashes between external dependencies, ignore them in '${AndroidAppChecksExtension::uniqueRClasses.name}.${UniqueRClasses::allowedNonUniquePackageNames.name}'
```
eugene-krivobokov, line 79: Example:

```
Application :app has dependencies with the same package in AndroidManifest.xml: [lib.package].
It leads to unexpected resource overriding.

How to fix: 
Make packages in manifests unique.
If this is impossible because of clashes between external dependencies, ignore them in 'uniqueRClasses.allowedNonUniquePackageNames' 

This check can be disabled in an extension:
buildChecks {
    uniqueRClasses {
        enabled = false
    }
}
See https://avito-tech.github.io/avito-android/projects/BuildChecks (search 'uniqueRClasses')
```

PR #1236, opened 2021-09-27
MBS-9085 Add make commands for cleaning build cache

eugene-krivobokov:


Makefile
```diff
@@ -122,9 +122,15 @@ __check_defined = \
     $(if $(value $1),, \
       $(error Undefined argument: $1$(if $2, ($2))))
 
-clean:
+clean_build:
```
eugene-krivobokov, line 125: Renamed to be consistent with internal repository

PR #1237, opened 2021-09-27
MBS-11985 Update dependency-analysis plugin

eugene-krivobokov:


subprojects/gradle/cd/src/gradleTest/kotlin/com/avito/ci/steps/DependencyAnalysisStepTest.kt
```diff
@@ -19,7 +19,7 @@ internal class DependencyAnalysisStepTest {
         TestProjectGenerator(
             plugins = plugins {
                 id("com.avito.android.cd")
-                id("com.autonomousapps.dependency-analysis").version("0.74.0")
+                id("com.autonomousapps.dependency-analysis").version("0.78.0")
```
eugene-krivobokov, line 22: Tech debt, but I want to remove this build step in MBS-12005

PR #1238, opened 2021-09-29
MBS-11997 remove warnings spam on ciSteps local applying

dsvoronin:
reason: non-actionable
how to see these messages locally: use verbose logging https://avito-tech.github.io/avito-android/contributing/Logging/#verbose-mode

### DIFF
--- a/subprojects/gradle/cd/src/main/kotlin/com/avito/ci/CiStepsPlugin.kt
+++ b/subprojects/gradle/cd/src/main/kotlin/com/avito/ci/CiStepsPlugin.kt
@@ -27,7 +27,7 @@ public class CiStepsPlugin : Plugin<Project> {
         project.extensions.add("builds", buildContainer)
 
         if (project.buildEnvironment !is BuildEnvironment.CI) {
-            logger.warn("The plugin is applied but it's disabled. Add -Pci=true to enable the plugin")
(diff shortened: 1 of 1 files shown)


subprojects/gradle/cd/src/main/kotlin/com/avito/ci/CiStepsPlugin.kt
```diff
@@ -27,7 +27,7 @@ public class CiStepsPlugin : Plugin<Project> {
         project.extensions.add("builds", buildContainer)
 
         if (project.buildEnvironment !is BuildEnvironment.CI) {
-            logger.warn("The plugin is applied but it's disabled. Add -Pci=true to enable the plugin")
+            logger.info("The plugin is applied but it's disabled. Add -Pci=true to enable the plugin")
```
eugene-krivobokov, line 30: I agree that this is not a warning.
Anyway, It will be a continuous toggling.
We rise logging levels after each hard troubleshooting when we found that something was just turned off. After some time we lower the same level.
It's a signal to fix root causes.

dsvoronin: yep, different behavior because of buildEnvironment check is a root cause in my opinion

PR #1239, opened 2021-09-29
Remove obsolete checks

eugene-krivobokov:


ci/docker/python-profiler/README.md
```diff
@@ -1 +0,0 @@
-Some scripts to parse and send gradle-profiler results are in python
```
eugene-krivobokov, line 1: Don't use it for a long time
:rooster:

PR #1243, opened 2021-09-29
MBS-9085 remove unused checks

dsvoronin:


subprojects/gradle/build-checks/src/gradleTest/kotlin/com/avito/android/build_checks/ConfigurationTest.kt
```diff
@@ -47,11 +47,7 @@ internal class ConfigurationTest {
                 javaVersion { enabled = false }
                 androidSdk { enabled = false }
                 macOSLocalhost { enabled = false }
-                dynamicDependencies { enabled = false }
-                gradleDaemon { enabled = false }
-                moduleTypes { enabled = false }
                 gradleProperties { enabled = false }
-                incrementalKapt { enabled = false }
```
eugene-krivobokov, line 54: We need to remove them from documentation also

PR #1249, opened 2021-10-08
MA-2182 Rework ownership plugin

RuslanMingaliev:


subprojects/gradle/code-ownership/src/main/kotlin/com/avito/android/CodeOwnershipPlugin.kt
```diff
@@ -14,17 +15,23 @@ public class CodeOwnershipPlugin : Plugin<Project> {
 
         val enabled = target.getBooleanProperty("avito.moduleOwnershipValidationEnabled", false)
 
+        val forceConfiguration =
```
eugene-krivobokov: A proposal:
To rename "force configuration" to "force ownership" \ "strict ownership" to reflect an intention rather than the configuration phase.

subprojects/gradle/code-ownership/src/main/kotlin/com/avito/android/CodeOwnershipPlugin.kt
```diff
@@ -14,17 +15,23 @@ public class CodeOwnershipPlugin : Plugin<Project> {
...
+            target.getBooleanProperty("avito.moduleOwnershipForceConfiguration", false)
```
eugene-krivobokov: Let's use common conventions for naming toggles: https://avito-tech.github.io/avito-android/contributing/GradlePlugins/#feature-toggles

`avito.ownership.X`

subprojects/gradle/code-ownership/src/main/kotlin/com/avito/android/CheckProjectDependenciesOwnershipTask.kt
```diff
@@ -1,93 +1,13 @@
 package com.avito.android
 
-import Visibility
-import com.avito.impact.configuration.internalModule
 import org.gradle.api.DefaultTask
 import org.gradle.api.tasks.TaskAction
 
+@Deprecated("Modules visibility restriction is deprecated. Left for backward compatibility")
```
eugene-krivobokov: Explain please what's the problem here.

Does any client know about this task? 
I thought that they know only about extension and shouldn't use tasks directly.

subprojects/gradle/code-ownership/src/main/kotlin/com/avito/android/CodeOwnershipPlugin.kt
```diff
@@ -14,17 +15,23 @@ public class CodeOwnershipPlugin : Plugin<Project> {
...
+                setupLibrary(subproject, enabled, forceConfiguration)
             }
         }
     }
 
-    private fun setupLibrary(project: Project, enabled: Boolean) {
+    private fun setupLibrary(project: Project, enabled: Boolean, forceConfiguration: Boolean) {
         val codeOwnershipExtension = project.extensions.create<CodeOwnershipExtension>("ownership")
 
         val ownershipTask =
```
midery: Should we register this task if it has no implementation?

subprojects/gradle/code-ownership/src/main/kotlin/com/avito/android/CheckProjectDependenciesOwnershipTask.kt
RuslanMingaliev: My bad, I was certain that this task is used in Avito, and its removal could lead to a broken CI. But it doesn't.

subprojects/gradle/code-ownership/src/main/kotlin/com/avito/android/CodeOwnershipPlugin.kt
RuslanMingaliev: See the conversation above

midery (approved):

subprojects/gradle/code-ownership/src/main/kotlin/com/avito/android/CodeOwnershipPlugin.kt
```diff
@@ -12,31 +13,33 @@ public class CodeOwnershipPlugin : Plugin<Project> {
     override fun apply(target: Project) {
         check(target.isRoot()) { "Code ownership plugin must be applied to the root project" }
 
-        val enabled = target.getBooleanProperty("avito.moduleOwnershipValidationEnabled", false)
+        val strictOwnership = target.getBooleanProperty("avito.ownership.strictOwnership", false)
 
         target.subprojects { subproject ->
```
eugene-krivobokov, line 18: We have to avoid iterating through subprojects. 
It gives a performance penalty while configuring a project.

This is the client's responsibility to apply the plugin to all modules. It can use convention plugins.
Please, schedule a fix for this issue.

eugene-krivobokov: MBS-12327

PR #1250, opened 2021-10-12
MBS-11968 Fix TestKey already has been received

sergioboi:


subprojects/test-runner/service/src/main/kotlin/com/avito/runner/service/DeviceWorkerPoolImpl.kt
```diff
@@ -48,9 +48,10 @@ internal class DeviceWorkerPoolImpl(
                             testListener = testListener,
                             deviceListener = CompositeDeviceListener(
                                 listOf(
+                                    // TODO order impact the result
```
sergioboi, line 51: The actual fix

sergioboi: But we don't have logs enough for being sure

subprojects/test-runner/service/src/main/kotlin/com/avito/runner/service/DeviceWorkerPoolImpl.kt
```diff
@@ -75,11 +76,11 @@ internal class DeviceWorkerPoolImpl(
...
 
                             is DeviceWorkerMessage.FailedIntentionProcessing -> {
-                                logger.debug(
+                                logger.warn(
                                     "Received worker failed message during executing intention:" +
-                                        " ${message.intention}. Rescheduling..."
+                                        " ${message.intention}. Rescheduling...", message.t
                                 )
-
+                                // TODO make rescheduling as a common with new execution number
```
eugene-krivobokov, line 83: Let's discuss this on call.
I've also suspected rescheduling of failed intentions as the root cause but couldn't find any proofs.
Added some details to MBS-11968 description.

PR #1253, opened 2021-10-13
MBS-11239

sergioboi:
The small refactoring. Look commit messages


subprojects/logger/logger/src/testFixtures/kotlin/com/avito/logger/PrintlnLogger.kt
```diff
@@ -1,8 +1,8 @@
 package com.avito.logger
 
-class StubLogger(private val tag: String) : Logger {
+public class PrintlnLogger(private val tag: String) : Logger {
```
eugene-krivobokov, line 3: Another possible naming is `StdoutLogger` to show the intention rather than implementation.

subprojects/gradle/git/src/main/kotlin/com/avito/git/GitImpl.kt
```diff
@@ -48,7 +43,6 @@ internal class GitImpl(
 
     private fun git(command: String): Result<String> =
         processRunner.run(command = "git $command", timeout = Duration.ofSeconds(10))
-            .onFailure { error -> logger.warn("git error running: '$command'", error) }
```
eugene-krivobokov, line 51: We will lose information about git errors. Why so?

eugene-krivobokov: Is it the responsibility of clients now? I mean, to get it from Result

subprojects/logger/logger/src/testFixtures/kotlin/com/avito/logger/PrintlnLogger.kt
```diff
@@ -1,8 +1,8 @@
...
 
-    private val isRunFromIde = true
+    private val isRunFromIde = System.getProperty("isInvokedFromIde") == "true"
 
     override fun debug(msg: String) {
         if (isRunFromIde) {
```
eugene-krivobokov: I'd like to avoid if's in all methods.
We could hide this switch in a logger factory and create a real empty stub implementation if we don't need logs.

subprojects/logger/logger/src/testFixtures/kotlin/com/avito/logger/PrintlnLogger.kt
```diff
@@ -1,8 +1,8 @@
 package com.avito.logger
 
-class StubLogger(private val tag: String) : Logger {
+public class PrintlnLogger(private val tag: String) : Logger {
 
-    private val isRunFromIde = true
+    private val isRunFromIde = System.getProperty("isInvokedFromIde") == "true"
```
eugene-krivobokov: This change disables logging in CI.
Is it an intentional change?

subprojects/logger/logger/src/testFixtures/kotlin/com/avito/logger/PrintlnLogger.kt
sergioboi: Yes but I don't see the difference so won't change the name

subprojects/gradle/git/src/main/kotlin/com/avito/git/GitImpl.kt
sergioboi: I don't know why we need this information.
The API provides Result for clients and they must decide what to do with errors

subprojects/logger/logger/src/testFixtures/kotlin/com/avito/logger/PrintlnLogger.kt
sergioboi: Agree

subprojects/logger/logger/src/testFixtures/kotlin/com/avito/logger/PrintlnLogger.kt
sergioboi: I simply return code to the previous state because `private val isRunFromIde = true` is the non sence

sergioboi: And as I understand this logger is used only in tests

eugene-krivobokov: Ok, I see

subprojects/gradle/git/src/main/kotlin/com/avito/git/GitImpl.kt
eugene-krivobokov: Agree

eugene-krivobokov (approved):

Makefile
```diff
@@ -308,3 +308,6 @@ deploy_avito_cache_node:
 delete_avito_cache_node:
 	cd ./ci/k8s/gradle-remote-cache && \
 	kubectl delete -f avito-project.yaml
+
+write_locks:
+	./gradlew resolveAndLockAll --write-locks
```
dsvoronin: Add `--no-configuration-cache --quiet > /dev/null` as in avito project:

`--no-configuration-cache` - needed here if cache enabled locally, because dependencies task doesn't support it
 `--quiet` and `> /dev/null` - for speedup, see avito PR

dsvoronin (approved):

PR #1255, opened 2021-10-13
MBS-12154 remove samples project

dsvoronin:
there are multiple issues with this composite builds configurations:
 - some builscript classpath problems causes AGP issues
 - brakes build caching somehow (not investigated enough to find the reason)
 - IDE sync is super long because all subprojects should be build on configuration phase

We lost the abiltiy to quickly iterate, running instrumentation plugin, but these problems are more than that; should be fixed some other way


dsvoronin:
Avito integration compatibility PR waiting for review

sergioboi (approved):

sergioboi:
Without samples we should try invest to integration tests, ABI changes checker

PR #1259, opened 2021-10-26
MA-2182 Add owners function with vararg

RuslanMingaliev:


subprojects/gradle/code-ownership/src/gradleTest/kotlin/com/avito/android/StrictOwnershipTest.kt
```diff
@@ -49,7 +49,7 @@ internal class StrictOwnershipTest {
                                     |def speed = new Owner() { }
                                     |
                                     |ownership {
-                                    |    owners = [speed]
+                                    |    owners(speed)
                                     |}
```
eugene-krivobokov, line 53: The target client uses it in kts.
Even if the implementation is interchangeable, it would be more practical to test more often usecase.

subprojects/gradle/code-ownership/src/main/kotlin/com/avito/android/CodeOwnershipExtension.kt
```diff
@@ -24,11 +29,11 @@ public open class CodeOwnershipExtension(
     private fun throwInvalidOwnershipSettingsException(projectPath: String) {
         throw IllegalStateException(
             """
-                |Owners must be set for $projectPath
-                |Configure ownership settings in $projectPath `build.gradle` file. For example: 
+                |Owners must be set for the $projectPath project.
+                |Configure the ownership extension for $projectPath in the `build.gradle.kts` file. For example:
```
eugene-krivobokov:
```suggestion
                |Configure the ownership extension for $projectPath in the `build.gradle[.kts]` file: 
```
OR
```suggestion
                |Configure the ownership extension for $projectPath in the buildscript: 
```
It's an assumption about the type of file. We don't know it here.

eugene-krivobokov (approved):

subprojects/gradle/code-ownership/src/gradleTest/kotlin/com/avito/android/StrictOwnershipTest.kt
RuslanMingaliev: Agree, rewrote the tests

eugene-krivobokov: Now the error message is correct only for kts.
If the client uses groovy DSL, it will be confusing.

There are two options:
- Provide the same API for Groovy DSL
- Show another error message

PR #1261, opened 2021-10-27
MBS-12251 [Instrumentation test plugin] use new AGP variants API

dsvoronin:
Gradle plugin refactoring:
- use new variant API for AGP
- use "configurators" to set groups of task properties
- add LocalRunInteractor to extract part of the plugin that works only to set correct args for local runs (should be probably removed in future)

~~- [ ] [fastCheck build](http://links.k.avito.ru/6Y)~~
- [x] [fastCheck build](http://links.k.avito.ru/HY)


subprojects/gradle/cd/src/gradleTest/kotlin/com/avito/ci/steps/UploadCdBuildResultIntegrationTest.kt
```diff
@@ -166,7 +167,9 @@ class RealTest {
             ).tryParseRev("HEAD").getOrThrow()
         )
 
-        val runId = "runId"
+        val commit = projectDir.git("rev-parse HEAD").trim()
+
+        val runId = "$commit.teamcity-BT"
```
dsvoronin, line 172: behavior change: `runId` set from `build.gradle` config `android.defaultConfig.testInstrumentationArgs` will be ignored
reason: we build complex RunId to be correct for our report engine, overriding here can be problematic

subprojects/test-runner/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/internal/AndroidVariantConfiguratorFactory.kt
```diff
@@ -0,0 +1,30 @@
+package com.avito.instrumentation.internal
+
+import com.android.build.api.variant.ApplicationVariant
+import com.android.build.api.variant.LibraryVariant
+import com.android.build.api.variant.Variant
+
+internal class AndroidVariantConfiguratorFactory {
+
+    fun createConfigurator(variant: Variant): AndroidVariantConfigurator<*>? {
```
sergioboi, line 9: Strange create that could return null

subprojects/test-runner/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/internal/BuildEnvResolver.kt
```diff
@@ -3,13 +3,13 @@ package com.avito.instrumentation.internal
 import com.avito.utils.gradle.envArgs
 import org.gradle.api.Project
 
-internal object BuildEnvResolver {
+internal class BuildEnvResolver(private val project: Project) {
```
sergioboi, line 6: I'm not sure why we need a project reference there

subprojects/test-runner/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/internal/ConfiguratorsFactory.kt
```diff
@@ -0,0 +1,112 @@
...
+    val localRunInteractor: LocalRunInteractor = LocalRunInteractor(
+        loggerFactory = loggerFactory,
+        argsTester = argsTester,
+        dslInteractor = androidDslInteractor
+    )
+
+    fun createTaskConfigurators(
+        configuration: InstrumentationConfiguration,
+        variant: Variant
+    ): List<InstrumentationTaskConfigurator>? {
```
sergioboi, line 69: emptylist?

subprojects/test-runner/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/internal/GitResolver.kt
```diff
@@ -4,17 +4,17 @@ import com.avito.git.gitState
 import org.gradle.api.Project
 import org.gradle.api.provider.Provider
 
-internal object GitResolver {
+internal class GitResolver(private val project: Project) {
```
sergioboi, line 7: The same, too wide scope to save whole Project there

subprojects/test-runner/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/internal/InstrumentationTaskConfigurator.kt
```diff
@@ -0,0 +1,8 @@
+package com.avito.instrumentation.internal
+
+import com.avito.instrumentation.InstrumentationTestsTask
+
+internal interface InstrumentationTaskConfigurator {
+
+    fun configure(task: InstrumentationTestsTask)
```
sergioboi, line 7:
```suggestion
    fun configure(task: Provider<InstrumentationTestsTask>)
```
Would be more Gradle way

sergioboi: Or you guarantee laziness at the upper level?

subprojects/test-runner/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/internal/ReportResolver.kt
```diff
@@ -3,34 +3,35 @@ package com.avito.instrumentation.internal
...
 import com.avito.instrumentation.configuration.ReportViewer
 
-internal class ReportResolver(private val runIdResolver: RunIdResolver) {
+internal class ReportResolver(
+    private val extension: InstrumentationTestsPluginExtension,
+    private val runIdResolver: RunIdResolver
+) {
 
-    fun getReportApiUrl(extension: InstrumentationTestsPluginExtension): String {
+    fun getReportApiUrl(): String {
```
sergioboi, line 11:
```suggestion
    val reportApiUrl: String
```
Why we don't use the property?

subprojects/test-runner/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/internal/AndroidVariantConfigurator.kt
```diff
@@ -0,0 +1,8 @@
+package com.avito.instrumentation.internal
+
+import com.android.build.api.variant.Variant
+
+internal interface AndroidVariantConfigurator<T : Variant> : InstrumentationTaskConfigurator {
+
+    val variant: T
```
sergioboi, line 7: I don't get why we need the property there
Looks like an abstract class with constructor argument

subprojects/test-runner/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/internal/AndroidVariantConfiguratorFactory.kt
dsvoronin: yep, i'm thinking of improvement, 
see here https://github.com/avito-tech/avito-android/pull/1261/files#diff-a9f5557634124eb2a0ba2c4457392e08c1c8ead18a370ff37347d311470e210cR48

subprojects/test-runner/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/internal/BuildEnvResolver.kt
dsvoronin: You mean to pass `envArgs`, or even `build` directly?

subprojects/test-runner/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/internal/ConfiguratorsFactory.kt
dsvoronin: should work, but problem lies here https://github.com/avito-tech/avito-android/pull/1261#discussion_r738319406

subprojects/test-runner/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/internal/GitResolver.kt
dsvoronin: ok, i'll pass `gitState` directly

subprojects/test-runner/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/internal/InstrumentationTaskConfigurator.kt
dsvoronin: I couldn't find a way to lazily configure task at the moment of registration, like here:
https://github.com/avito-tech/avito-android/pull/1261/files#diff-a9f5557634124eb2a0ba2c4457392e08c1c8ead18a370ff37347d311470e210cR57

It wants Action<InstrumentationTestsTask> as parameter, which i use but only to iterate through `Configurators`
Not sure if passing knowledge that it is a configuration is worth it.

subprojects/test-runner/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/internal/ReportResolver.kt
dsvoronin: i think we can

subprojects/test-runner/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/internal/AndroidVariantConfigurator.kt
dsvoronin: abstract class should work, though about more resposibility for this configurator, like more interfaces applied, but ended up single inheritance

subprojects/test-runner/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/internal/BuildEnvResolver.kt
dsvoronin: can't in fact: envArgs should be lazy, it's a debt: https://github.com/avito-tech/avito-android/blob/71c71f8eb633b4e4bfc81984fe3f62b967b3806a/subprojects/gradle/gradle-extensions/src/main/kotlin/com/avito/kotlin/dsl/ProjectExtensions.kt#L92

subprojects/test-runner/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/internal/ReportResolver.kt
dsvoronin: ReportViewer model is not a properties, just var's https://github.com/avito-tech/avito-android/blob/71c71f8eb633b4e4bfc81984fe3f62b967b3806a/subprojects/test-runner/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/configuration/InstrumentationTestsReportExtension.kt#L21

so it should be queried lazily, properties with getters will work, but not with eager initializers

PR #1263, opened 2021-10-28
MBS-12153 [Instrumentation plugin] support configuration cache

dsvoronin:
Previous test only checked if configuration cache works with instrumentation plugin applied.
New one checks if its working when instrumentaionTask is actually executing


subprojects/test-runner/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/InstrumentationTestsTask.kt
```diff
@@ -144,6 +142,7 @@ public abstract class InstrumentationTestsTask @Inject constructor(
     @get:Internal
     override val verdict: SpannedString
         get() {
+            val gson: Gson = GsonBuilder().setPrettyPrinting().create()
```
dsvoronin, line 143: Configuration cache tried to serialize private gson property.
I think it's ok to create it here every time.

PR #1264, opened 2021-10-28
MBS-12056 [Instrumentation plugin] support flavors

dsvoronin:
plugin will produce tasks with flavors if present:
for example: `instrumentationDemoUi`, where `demo` - flavor name and `ui` - instrumentation configuration

afaik tests can be run for only one buildType, specified by `testBuildType` property, so no need to use variants here

multidimensional flavors also supported, but i'm not sure how to handle which dimension goes first, so i create tasks for all combinations

cc @materkey

### DIFF
--- a/subprojects/gradle/cd/src/main/kotlin/com/avito/ci/steps/TestSummaryPluginBuildStep.kt
+++ b/subprojects/gradle/cd/src/main/kotlin/com/avito/ci/steps/TestSummaryPluginBuildStep.kt
@@ -42,7 +42,8 @@ public abstract class TestSummaryPluginBuildStep(context: String, name: String) 
 
         if (useImpactAnalysis && !project.internalModule.isModified()) return
 
-        val instrumentationTask = project.tasks.instrumentationTask(configuration)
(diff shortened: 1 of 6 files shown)


subprojects/gradle/cd/src/main/kotlin/com/avito/ci/steps/TestSummaryPluginBuildStep.kt
```diff
@@ -42,7 +42,8 @@ public abstract class TestSummaryPluginBuildStep(context: String, name: String)
 
         if (useImpactAnalysis && !project.internalModule.isModified()) return
 
-        val instrumentationTask = project.tasks.instrumentationTask(configuration)
+        // flavors not supported here
```
sergioboi, line 45: And what does it mean? What should I consider looking at this comment?

dsvoronin: it's a reminder that CiPlugin doesn't know about flavors

PR #1265, opened 2021-10-29
MBS-11834 [Instrumentation plugin] add instrumentation.environment k8s config

dsvoronin:
instead of implicit gradle parameters
"default" config added for backward compatibility


subprojects/test-runner/instrumentation-tests/src/gradleTest/kotlin/com/avito/instrumentation/StubConfig.kt
```diff
@@ -65,7 +66,20 @@ internal val kotlinStubConfig = """
   |            }
   |        }
   |    }
-  |}    
+  |    
+  |    environments {
+  |       register<com.avito.instrumentation.configuration.KubernetesViaContext>("k8sContext") {
```
sergioboi, line 71: It will be simpler for users If they can write
```kotlin
environments {
    kubernetesViaContext("name") { }
    ... etc
}
```

subprojects/test-runner/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/InstrumentationTestsPlugin.kt
```diff
@@ -103,6 +132,38 @@ public class InstrumentationTestsPlugin : Plugin<Project> {
...
+
+        extension.environmentsContainer.register<LocalAdb>("local")
+
         return extension
     }
+
+    /**
+     * todo remove (registered for backward compatibility)
+     */
+    private fun registerDefaultEnvironment(
```
sergioboi, line 154: When do you plan to remove `default` env?

subprojects/test-runner/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/internal/EnvironmentConfigurator.kt
```diff
@@ -0,0 +1,43 @@
...
+
+internal class EnvironmentConfigurator(
+    private val environment: ExecutionEnvironment
+) : InstrumentationTaskConfigurator {
+
+    override fun configure(task: InstrumentationTestsTask) {
+        when (environment) {
+            is KubernetesViaCredentials ->
+                task.kubernetesCredentials.set(
+                    KubernetesCredentials.Service(
```
sergioboi, line 18: Looks like we want to reuse those models

subprojects/test-runner/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/InstrumentationTestsPlugin.kt
dsvoronin: Right after CISteps removal, because almost every build uses instrumentation tasks, which uses default environment for now; so in a couple of days

subprojects/test-runner/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/internal/EnvironmentConfigurator.kt
dsvoronin: I'm not sure that we need to reuse models for gradle public API,
if it will be covered by proposed dsl abstraction: https://github.com/avito-tech/avito-android/pull/1265#discussion_r748098479 then yes

subprojects/test-runner/instrumentation-tests/src/gradleTest/kotlin/com/avito/instrumentation/StubConfig.kt
dsvoronin: looks good, yes

PR #1267, opened 2021-11-01
MBS-12273 setup gradle enterprise

dsvoronin:
local setup has changed, docs updated

enterprise properties names are borrowed from internal project, should be renamed there to include `avito.` prefix


build-logic-settings/cache-plugin/src/main/kotlin/convention-cache.settings.gradle.kts
```diff
@@ -1,18 +1,34 @@
...
+    if (useEnterprise) {
+        remote<HttpBuildCache> {
+            setUrl("${mandatoryStringProperty("avito.gradle.enterprise.url")}/cache/")
+            isEnabled = true
+            isPush = booleanProperty("avito.gradle.buildCache.remote.push", false)
+            isAllowUntrustedServer = true
+            isAllowInsecureProtocol = true
+        }
+    } else {
+        gradle.buildFinished {
```
dsvoronin, line 28: not supported for configuration cache, but acts good: first build prints visible warning (after the build) and following ones will not print anything

looks like a good balance for me without creating ugly build service workarounds

Makefile
```diff
@@ -265,38 +265,5 @@ benchmark_fast_check:
 benchmark_gradle_test:
 	gradle-profiler --benchmark --scenario-file gradle/performance.scenarios gradleTest
 
-## Gradle cache node
-GRADLE_CACHE_NODE_TAG=9.9
-
-# publish to internal repo to avoid rate limits problems
-internal_publish_gradle_cache_node_image:
```
eugene-krivobokov: Why do we remove it?
Have we made a final decision about Enterprise?

build-logic-settings/settings.gradle.kts
```diff
@@ -49,24 +49,47 @@ dependencyResolutionManagement {
...
-        isEnabled = avitoGithubRemoteCacheHost.orNull != null
-        isPush = avitoGithubRemoteCachePush.toBoolean()
-        isAllowUntrustedServer = true
-        isAllowInsecureProtocol = true
+    local {
+        isEnabled = true
+        isPush = true
+        removeUnusedEntriesAfterDays = 30
+    }
+    if (useEnterprise) {
```
eugene-krivobokov: This decision should be made by "is remote cache enabled" flag.
Enterprise is only one implementation to conform HttpBuildCache contract.

build-logic-settings/settings.gradle.kts
```diff
@@ -49,24 +49,47 @@ dependencyResolutionManagement {
...
 
 buildCache {
-    remote<HttpBuildCache> {
-        setUrl("http://${avitoGithubRemoteCacheHost.orNull}/cache/")
-        isEnabled = avitoGithubRemoteCacheHost.orNull != null
-        isPush = avitoGithubRemoteCachePush.toBoolean()
-        isAllowUntrustedServer = true
-        isAllowInsecureProtocol = true
+    local {
+        isEnabled = true
```
eugene-krivobokov: Even inside Docker container? Have we measured it?

docs/content/contributing/internal/RemoteCache.md
```diff
@@ -1,46 +1,13 @@
...
-    - DOCKER_REGISTRY env set
-    - GRADLE_CACHE_NODE_HOST env set
-    - [kubernetes credentials configured](http://links.k.avito.ru/kubectl)
-    - Write access to `gradle-remote-cache` namespace in cluster (see MBS-5444)
-
-???+ warning
-    Persistence volume not set, deleting a pod will lead to losing cache.
-
-`make deploy_gradle_cache_node`
+Add gradle properties to your `$GRADLE_USER_HOME/gradle.properties`:
```
eugene-krivobokov: `$GRADLE_USER_HOME` - to be precise, this env is optional

build-logic-settings/cache-plugin/src/main/kotlin/convention-cache.settings.gradle.kts
```diff
@@ -1,18 +1,34 @@
...
-@Suppress("UnstableApiUsage")
-val avitoGithubRemoteCacheHost: Provider<String> = settings.providers
-    .environmentVariable("GRADLE_CACHE_NODE_HOST")
-    .forUseAtConfigurationTime()
+import com.avito.booleanProperty
+import com.avito.mandatoryStringProperty
 
-val avitoGithubRemoteCachePush: String =
-    extra.properties.getOrDefault("avitoGithub.gradle.buildCache.remote.push", "false").toString()
+// See also duplicated settings in parent project
```
eugene-krivobokov: Let's describe here reasons from https://github.com/gradle/gradle/issues/18511

build-logic-settings/cache-plugin/src/main/kotlin/convention-cache.settings.gradle.kts
eugene-krivobokov: > ugly build service workarounds

What are you talking about?

eugene-krivobokov: > first build prints visible warning (after the build) and following ones will not print anything

My point is no one reads or should read build logs.
I propose to use logs only as side effects and never rely on them in preventing problems.
It's too easy to miss a one-time warning and no one will see the problem.

If we strongly don't want to make this property mandatory we could use tags\values in build scan.

dsvoronin: > What are you talking about?

https://github.com/runningcode/gradle-doctor/blob/e6cd0299da6e537b05e07ece8ba6813ccccd387b/doctor-plugin/src/main/java/com/osacky/doctor/BuildFinishService.kt

dsvoronin: > If we strongly don't want to make this property mandatory

New users should set it to false explicitly?

> My point is no one reads or should read build logs.

Bold message in the end on build is hard to miss?

> we could use tags\values in build scan.

it's easy to miss also, but i see no harm adding a tag also

Makefile
dsvoronin: Nope, we can remove it later, yes

build-logic-settings/settings.gradle.kts
dsvoronin: You're right, we don't need in container

docs/content/contributing/internal/RemoteCache.md
dsvoronin: Do you know a better way to express that it's not always a `~/.gradle`?

build-logic-settings/settings.gradle.kts
dsvoronin: Removed this flag entirely;
also removed backup to cache node.

My reasoning is: when we will decide about enterprise it will stay as it is or be removed in favor of separate node; but maintaining both ways is in my opinion an overkill for this project

build-logic-settings/cache-plugin/src/main/kotlin/convention-cache.settings.gradle.kts
eugene-krivobokov: > Bold message in the end on build is hard to miss?

The point is about logs in general. I see how to make a solid process around tags in scans or around other events.
With logs _only_, I have only a hope that a developer will notice it. Hope is not a strategy.

dsvoronin: I agree with you; let's setup tags after we will make a decision about enterprise

eugene-krivobokov: >> If we strongly don't want to make this property mandatory

> New users should set it to false explicitly?

Yes, they should. This is inconvenient for new external contributors but more robust than warning in logs.
I don't argue for this solution, just wanted to mention this decision.

build-logic-settings/settings.gradle.kts
eugene-krivobokov: Ok, I estimate supporting differently but don't want to discuss details by text.

docs/content/contributing/internal/RemoteCache.md
eugene-krivobokov: [`GRADLE_USER_HOME`](https://docs.gradle.org/current/userguide/build_environment.html#sec:gradle_environment_variables)

build-logic-settings/cache-plugin/src/main/kotlin/convention-cache.settings.gradle.kts
```diff
@@ -1,18 +1,35 @@
...
-val avitoGithubRemoteCacheHost: Provider<String> = settings.providers
-    .environmentVariable("GRADLE_CACHE_NODE_HOST")
-    .forUseAtConfigurationTime()
+import com.avito.booleanProperty
+import com.avito.stringProperty
 
-val avitoGithubRemoteCachePush: String =
-    extra.properties.getOrDefault("avitoGithub.gradle.buildCache.remote.push", "false").toString()
+// Duplicated in settings in parent project
+// see https://github.com/gradle/gradle/issues/18511
```
eugene-krivobokov, line 8:
```suggestion
// Can't use only one build cache configuration in the root project.
// Included builds inside `pluginManagement` block can't reuse it, because they are evaluated earlier than the build cache configuration in the root project.
```

The issue will be closed as "won't fix", it can't be fixed.
It'd be more convenient to have details without reading the issue.

eugene-krivobokov (approved):

PR #1270, opened 2021-11-02
MBS-12172 Add Robolectric support

RuslanMingaliev:


eugene-krivobokov (approved):

subprojects/android-test/ui-testing-core-app/src/sharedTest/kotlin/com/avito/android/ui/test/ButtonsOverRecyclerTest.kt
```diff
@@ -1,13 +1,16 @@
...
+import androidx.test.ext.junit.runners.AndroidJUnit4
 import com.avito.android.test.app.core.screenRule
 import com.avito.android.ui.ButtonsOverRecyclerActivity
 import org.hamcrest.Matchers.equalTo
 import org.hamcrest.Matchers.greaterThan
 import org.junit.Rule
 import org.junit.Test
+import org.junit.runner.RunWith
 
+@RunWith(AndroidJUnit4::class)
```
dsvoronin, line 13: Why do we need those? and what's the default runner?

build-logic/android/src/main/kotlin/convention.android-robolectric.gradle.kts
```diff
@@ -0,0 +1,37 @@
+import com.android.build.gradle.BaseExtension
+import com.avito.android.withVersionCatalog
+
+plugins {
+    id("convention.android-base")
+}
+
+configure<BaseExtension> {
+    val sharedTestSourcesFolder = "src/sharedTest/kotlin"
```
sergioboi, line 9: So tests should be placed there?

RuslanMingaliev: You can run tests that are placed there both with an emulator and without it.

dsvoronin: Did you check that these tests still runs on emulators on PR?

subprojects/android-test/ui-testing-core-app/src/sharedTest/kotlin/com/avito/android/ui/test/ButtonsOverRecyclerTest.kt
RuslanMingaliev: Espresso tests must run under instrumentation. `InstrumentationRegistry.getInstrumentation()` with throw an exception otherwise. [By default](https://cs.android.com/androidx/android-test/+/master:ext/junit/java/androidx/test/ext/junit/runners/AndroidJUnit4.java;l=59), `RobolectricTestRunner` is used for unit tests and `AndroidJUnit4ClassRunner` for Android tests on the emulator (if it is not overridden by `defaultConfig.testInstrumentationRunner`).

dsvoronin: Could it be set globally for sharedTest somehow?
It's easy to forget to add one, and i bet an error will be hard to debug

RuslanMingaliev: Didn't find a way to do that. 

You can face the problem in 2 cases: 
1. When creating a new test
2. When moving the existing test to the shared folder

In both cases the error message "`java.lang.IllegalStateException: No instrumentation registered! Must run under a registering instrumentation.`" seems understandable and clear.

dsvoronin: > seems understandable and clear

I can't see how it's connected with run annotation

build-logic/android/src/main/kotlin/convention.android-robolectric.gradle.kts
RuslanMingaliev: Tests on emulators are broken now. I've created a bug, see MBS-12302.

PR #1276, opened 2021-11-12
MBS-11239 Add unique dir for each build execution

sergioboi:


subprojects/logger/gradle-logger/src/main/kotlin/com/avito/logger/LegacyGradleLoggerConfigurator.kt
```diff
@@ -15,7 +16,8 @@ internal class LegacyGradleLoggerConfigurator(
             val verbosity = getVerbosity() ?: LogLevel.INFO
             printlnHandler.set(GradleLoggerExtension.PrintlnMode(verbosity, doPrintStackTrace()))
             fileHandler.set(LogLevel.INFO)
-            fileHandlerRootDir.set(project.layout.buildDirectory.dir("logs"))
+            val instantString = Instant.now().toString()
```
sergioboi: This file will be deleted so duplication is not crucial

subprojects/logger/gradle-logger/src/main/kotlin/com/avito/logger/GradleLoggerPlugin.kt
```diff
@@ -85,11 +86,12 @@ public class GradleLoggerPlugin : Plugin<Project> {
                 LoggerService::javaClass.name,
                 LoggerService::class.java
             ) {
+                val instantString = Instant.now().toString()
```
eugene-krivobokov: Result string contains `:` symbol - `2021-11-12T14:36:34.262Z`.
In older HFS in macOs it converted to dir delimiter `/`.
Also not sure about Windows.
https://en.wikipedia.org/wiki/Filename#Reserved_characters_and_words

I'd like to avoid such symbols just to be more confident.

eugene-krivobokov (approved):

subprojects/logger/gradle-logger/src/main/kotlin/com/avito/logger/GradleLoggerPlugin.kt
sergioboi: Fixed the format now is yyyy-mm-dd-hh-mm-ss

PR #1277, opened 2021-11-16
MBS-12298 qapps plugin rework

dsvoronin:
two new tasks: qappsUploadUnsignedVariant/qappsUploadSignedVariant instead of qappsUploadVariant

signed is directly depend on signService task
Unsigned on AGP


eugene-krivobokov (approved):

subprojects/gradle/qapps/src/main/kotlin/com/avito/plugin/QAppsPlugin.kt
```diff
@@ -16,23 +18,72 @@ public class QAppsPlugin : Plugin<Project> {
...
-                        GradleLoggerPlugin.getLoggerFactory(this)
-                    )
+                        val apkProvider = signTaskProvider.flatMap { it.signedDirProperty }
+
+                        apkDirectory.set(apkProvider)
+
+                        // todo remove, somehow implicit dependency not working
+                        dependsOn(signTaskProvider)
+                    } else {
+                        logger.info(
```
sergioboi, line 58: looks like a warning

PR #1281, opened 2021-11-17
MBS-12332 signer plugin rework

dsvoronin:
register tasks only for configured build variants

more tests on new logic and docs will be in next rounds; too much for a single PR


subprojects/delivery/qapps/src/main/kotlin/com/avito/plugin/QAppsPlugin.kt
```diff
@@ -46,7 +46,7 @@ public class QAppsPlugin : Plugin<Project> {
                     configure(extension, variant)
 
                     if (project.plugins.hasPlugin(SignServicePlugin::class.java)) {
-                        val signTaskProvider = project.tasks.signedApkTaskProvider(variant.name)
+                        val signTaskProvider = project.tasks.legacySignedApkTaskProvider(variant.name)
```
sergioboi, line 49: Did you think to move this logic to the client side? Or this isn't possible?

subprojects/delivery/signer/src/main/kotlin/com/avito/plugin/AbstractSignTask.kt
```diff
@@ -0,0 +1,104 @@
...
+            )
+            .build()
+
+        val logger = loggerFactory.create<LegacySignArtifactTask>()
+
+        workerExecutor.inMemoryWork {
+            SignViaServiceAction(
+                serviceUrl = serviceUrl,
+                httpClient = httpClient,
+                token = tokenProperty.get(),
```
sergioboi, line 77: Propose to unwrap directly in the task action method
Is the error message is clear?

subprojects/delivery/signer/src/main/kotlin/com/avito/plugin/SignBundleTask.kt
```diff
@@ -1,42 +1,18 @@
...
-     * transform API is broken for ArtifactType.Bundle, hacking here to provide expected result
-     * https://issuetracker.google.com/issues/174678813
-     */
-    private fun resultLocation(): File {
-        return signedFile().path
-            .replaceAfterLast("/", unsignedFile().name) // replace bugged `out` file name with original one
-            .replace("/$name", "") // remove `signBundleViaServiceVariant` from path, bugged
-            .let { File(it) }
-    }
+    override fun unsignedFile(): File = bundleFile.get().asFile
```
sergioboi, line 19: As I understand correctly, this code will inline value when the task is creating
So, I propose to use get() = notation

subprojects/delivery/signer/src/main/kotlin/com/avito/plugin/SignServicePlugin.kt
```diff
@@ -42,65 +44,83 @@ public class SignServicePlugin : Plugin<Project> {
 
                 val urlResolver = UrlResolver(extension)
 
+                if (extension.apkSignTokens.containsKey(variant.name)) {
+                    target.tasks.register<SignApkTask>(signApkTaskName(variant.name)) {
+                        description = "Signs ${variant.name} apk with in-house service"
+                        apkDirectory.set(variant.artifacts.get(SingleArtifact.APK))
+                        tokenProperty.set(extension.apkSignTokens[variant.name]!!)
```
sergioboi, line 51: We will have bad error message there

sergioboi: Sorry I don't get that it's impossible

sergioboi (approved):

subprojects/delivery/qapps/src/main/kotlin/com/avito/plugin/QAppsPlugin.kt
dsvoronin: all "legacy" things will be removed after migration

subprojects/delivery/signer/src/main/kotlin/com/avito/plugin/AbstractSignTask.kt
dsvoronin: Good catch!
That's the lack of tests showing up
will cover that scenario

subprojects/delivery/signer/src/main/kotlin/com/avito/plugin/LegacySignBundleTask.kt
```diff
@@ -0,0 +1,42 @@
...
+
+    override fun hackForArtifactsApi() {
+        signedFile().copyTo(resultLocation(), overwrite = true)
+    }
+
+    /**
+     * transform API is broken for ArtifactType.Bundle, hacking here to provide expected result
+     * https://issuetracker.google.com/issues/174678813
+     */
+    private fun resultLocation(): File {
```
dsvoronin, line 36: todo: this problem is still here for new implementation, should check this

subprojects/delivery/signer/src/main/kotlin/com/avito/plugin/SignBundleTask.kt
dsvoronin: yep, you're right if it was a val, but not for fun's

subprojects/delivery/signer/src/main/kotlin/com/avito/plugin/SignServicePlugin.kt
dsvoronin: i with i can express this better with kotlin.
Maybe some unwrap private function will increase readability

PR #1283, opened 2021-11-18
MBS-12292 Robolectric 4.7.1

RuslanMingaliev:


subprojects/android-test/ui-testing-core-app/src/sharedTest/resources/com/avito/android/ui/robolectric.properties
```diff
@@ -1,4 +1,5 @@
 sdk=22,29
+qualifiers=w320dp-h480dp
```
RuslanMingaliev, line 2: The same as used in the instrumentation tests

eugene-krivobokov (approved):

PR #1284, opened 2021-11-18
MBS-12266 Support using testFixtures inside TestKit project

eugene-krivobokov:
### The problem

I need to test a case when a user extends classes from a plugin inside a precompiled script plugin. 
These custom classes are used in build scripts in modules with the plugin.

### Constraints

I can't declare these classes in build scripts directly because they won't be accessible to all modules.

I couldn't simulate this case directly with a precompiled script plugin in generated project.
The plugin classpath (`GradleRunner.withPluginClasspath`) is not accessible inside included builds (where precompiled script plugins live).
My question in Gradle slack: [link](https://gradle-community.slack.com/archives/CA745PZHN/p1636981349087300)

### Changes

I've supported a workaround:

- Declare classes inside testFixtures
- Add them to plugin classpath

Screenshot from 2021-11-18 21-29-30

This is a preliminary PR, usages will be in the next PR.

UPD: Example - #1286


subprojects/gradle/test-project/src/main/kotlin/com/avito/test/gradle/GradleTestKit.kt
```diff
@@ -42,43 +43,38 @@ public fun gradlew(
...
+
+            if (useModuleClasspath) {
+                withPluginClasspath()
             }
 
+            if (useTestFixturesClasspath) {
+                withPluginClasspath(
+                    pluginClasspath + testFixturesClasspath()
+                )
+            }
```
eugene-krivobokov, line 75: We add `testFixtures` here

subprojects/gradle/test-project/src/main/kotlin/com/avito/test/gradle/module/Module.kt
```diff
@@ -6,9 +6,13 @@ import com.avito.test.gradle.plugin.PluginsSpec
 
 public interface Module : Generator {
     public val name: String
+    public val imports: List<String>
```
eugene-krivobokov, line 9: Imports are valid in any Gradle project even without plugins

sergioboi (approved):

PR #1285, opened 2021-11-22
MBS-12266 Share module dependencies logic

eugene-krivobokov:
I need to reuse the logic of finding dependencies in a module.
Implementation in `module-dependencies-graph` plugin looks good enough to reuse.
I've extracted it to a shared module.


subprojects/gradle/impact-shared/build.gradle.kts
```diff
@@ -5,7 +5,7 @@ plugins {
 }
 
 dependencies {
-    api(projects.subprojects.gradle.moduleDependenciesGraph)
```
eugene-krivobokov, line 8: We consumed here a plugin that is worse.

subprojects/gradle/module-dependencies-shared/src/main/kotlin/com/avito/module/dependencies/ModuleConfigurationDependencies.kt
```diff
@@ -4,20 +4,21 @@ import com.avito.module.configurations.ConfigurationType
 import org.gradle.api.Project
 import org.gradle.api.artifacts.ProjectDependency
 
-public fun Project.dependenciesOnProjects(
-    types: Set<Class<out ConfigurationType>>
-): Set<ProjectDependency> {
+public fun Project.directDependenciesOnProjects(
+    types: Set<ConfigurationType>
+): Map<ConfigurationType, Set<Project>> {
```
eugene-krivobokov, line 9: We lost a mapping before. 
This is a client's responsibility to decide what to do with this information.

subprojects/gradle/module-dependencies-shared/src/main/kotlin/com/avito/module/configurations/ConfigurationType.kt
```diff
@@ -2,31 +2,31 @@ package com.avito.module.configurations
 
 import org.gradle.api.artifacts.Configuration
 
-public sealed class ConfigurationType {
+public enum class ConfigurationType {
```
eugene-krivobokov, line 5: Don't see the benefits of using a sealed class here.
Clients have to convert all the time to class.

dsvoronin (approved):

settings.gradle.kts
```diff
@@ -38,6 +38,7 @@ include(":subprojects:gradle:graphite-config")
 include(":subprojects:gradle:impact")
 include(":subprojects:gradle:impact-shared")
 include(":subprojects:gradle:module-dependencies-graph")
+include(":subprojects:gradle:module-dependencies-shared")
```
sergioboi, line 41: I'm not confident in that name. It would be good If you could replace `shared` with something meaningful

eugene-krivobokov: Agree, I'll try to find more obvious name

eugene-krivobokov: #1298

PR #1291, opened 2021-11-26
MBS-12266 Usability fixes in modules types plugin

eugene-krivobokov:


subprojects/gradle/module-types/src/main/kotlin/com/avito/android/module_type/DependencyMatcher.kt
```diff
@@ -4,7 +4,23 @@ import com.avito.module.configurations.ConfigurationType
 
 public interface DependencyMatcher {
 
-    public fun match(from: ModuleWithType, to: ModuleWithType, configuration: ConfigurationType): Boolean
+    @Deprecated(
+        "Use matches function instead",
+        ReplaceWith("matches")
+    )
+    public fun match(from: ModuleWithType, to: ModuleWithType, configuration: ConfigurationType): Boolean =
```
eugene-krivobokov, line 11: from / to is more ambiguous. It implies direction that is not obvious.

dsvoronin (approved):

midery (approved):

subprojects/gradle/code-ownership/src/gradleTest/kotlin/com/avito/android/ExportOwnershipInfoTest.kt
```diff
@@ -27,9 +27,6 @@ internal class ExportOwnershipInfoTest {
                 AndroidAppModule(
                     "app",
                     imports = listOf("import com.avito.android.model.Owner"),
-                    plugins = plugins {
-                        id("com.avito.android.module-types")
```
eugene-krivobokov, line 31: @midery , @RuslanMingaliev , cc
It wasn't really used here.

PR #1296, opened 2021-11-30
MBS-11239 Add AndroidLogcatLoggingHandler.kt

sergioboi:


subprojects/logger/logger/src/main/kotlin/com/avito/logger/metadata/LoggerMetadata.kt
```diff
@@ -0,0 +1,6 @@
+package com.avito.logger.metadata
+
+public interface LoggerMetadata {
+    public val asMessagePrefix: String
```
sergioboi, line 4: The change I would like you to look at
I delete LoggingFormatter and inline adding logger info to the message to concrete LoggingHandlers
i.e PrintlnLoggingHandler adds the `messagePrefix` but Sentry doesn't

PR #1304, opened 2021-12-15
MBS-12444 Fix missing configurations in module dependencies

eugene-krivobokov:


subprojects/gradle/module-dependencies/src/main/kotlin/com/avito/module/dependencies/ModuleConfigurationDependencies.kt
```diff
@@ -2,23 +2,28 @@ package com.avito.module.dependencies
...
+
+    return configurations.groupBy { ConfigurationType.of(it) }
+        .filterKeys { configurationType -> types.contains(configurationType) }
+        .mapValues { (_, configurations) ->
+            configurations
+                .flatMap { directProjectDependencies(project, it) }
                 .toSet()
         }
-        .filterValues { it.isNotEmpty() }
-        .mapKeys { ConfigurationType.of(it.key) }
```
eugene-krivobokov, line 20: Were losing configurations here because they have the same type in map

PR #1307, opened 2022-01-12
MBS-12124 create new nupokati plugin for all release stuff

dsvoronin:
Goal of this review is to check if current behavior of release didn't change (on top of what tests says).
New plugin has some stuff duplicated to have less dependencies.
New plugin work is in progress

### DIFF
--- a/subprojects/delivery/nupokati/build.gradle.kts
+++ b/subprojects/delivery/nupokati/build.gradle.kts
@@ -1,0 +1,35 @@
+plugins {
+    id("convention.kotlin-jvm")
+    id("convention.publish-gradle-plugin")
+    id("convention.gradle-testing")
+    id("convention.test-fixtures")
+}
+
+dependencies {
+    implementation(libs.gson)
+    implementation(libs.okhttp)
+    implementation(libs.androidGradle)
+
+    implementation(projects.subprojects.common.httpClient)
+    implementation(projects.subprojects.common.okhttp)
+    implementation(projects.subprojects.common.problem)
+    implementation(projects.subprojects.delivery.uploadToGoogleplay)
+    implementation(projects.subprojects.gradle.git)
+    implementation(projects.subprojects.gradle.gradleExtensions)
+    implementation(projects.subprojects.gradle.statsdConfig)
+    implementation(projects.subprojects.logger.slf4jGradleLogger)
+    implementation(projects.subprojects.testRunner.instrumentationTests)
+    implementation(projects.subprojects.testRunner.reportViewer)
+
+    gradleTestImplementation(projects.subprojects.gradle.testProject)
+}
+
+gradlePlugin {
+    plugins {
+        create("nupokati") {
+            id = "com.avito.android.nupokati"
+            implementationClass = "com.avito.android.NupokatiPlugin"
+            displayName = "Nupokati service plugin"
+        }
+    }
+}
--- a/subprojects/delivery/nupokati/src/main/kotlin/com/avito/android/ArtifactoryExtension.kt
+++ b/subprojects/delivery/nupokati/src/main/kotlin/com/avito/android/ArtifactoryExtension.kt
@@ -1,0 +1,24 @@
+package com.avito.android
+
+import org.gradle.api.provider.Property
+
+public abstract class ArtifactoryExtension {
+
+    public abstract val baseUrl: Property<String>
+
+    public abstract val login: Property<String>
+
+    public abstract val password: Property<String>
+
+    /**
+     * This parameters used to create path in artifactory repository
+     * and differs from one that comes from json cd build config
+     * There is no real reason of that
+     *
+     *  todo merge, take one from json
+     */
+    public abstract val backupRepository: Property<String>
+    public abstract val projectName: Property<String>
+    public abstract val projectType: Property<String>
+    public abstract val version: Property<String>
+}
--- a/subprojects/delivery/nupokati/src/main/kotlin/com/avito/android/google_play/DeployBundleToGooglePlayTask.kt
+++ b/subprojects/delivery/nupokati/src/main/kotlin/com/avito/android/google_play/DeployBundleToGooglePlayTask.kt
@@ -1,0 +1,51 @@
+package com.avito.android.google_play
+
+import com.avito.android.model.CdBuildConfig
+import com.avito.upload_to_googleplay.GooglePlayDeploy
+import com.avito.upload_to_googleplay.createGooglePlayDeployer
+import org.gradle.api.DefaultTask
+import org.gradle.api.file.RegularFile
+import org.gradle.api.model.ObjectFactory
+import org.gradle.api.provider.Property
+import org.gradle.api.tasks.Input
+import org.gradle.api.tasks.InputFile
+import org.gradle.api.tasks.TaskAction
+import javax.inject.Inject
+
+public abstract class DeployBundleToGooglePlayTask @Inject constructor(objects: ObjectFactory) : DefaultTask() {
+
+    @get:InputFile
+    public val googlePlayKeyFile: Property<RegularFile> = objects.fileProperty()
+
+    @get:InputFile
+    public val bundle: Property<RegularFile> = objects.fileProperty()
+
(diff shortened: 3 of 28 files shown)


subprojects/delivery/nupokati/src/main/kotlin/com/avito/android/provider/CdBuildConfigTransformer.kt
```diff
@@ -0,0 +1,37 @@
...
+import com.google.gson.Gson
+import com.google.gson.GsonBuilder
+import org.gradle.api.Transformer
+import org.gradle.api.file.ProjectLayout
+import java.io.File
+
+internal class CdBuildConfigTransformer(
+    private val rootProjectLayout: ProjectLayout,
+    private val validator: CdBuildConfigValidator,
+) : Transformer<CdBuildConfig, String> {
```
eugene-krivobokov, line 14: `Transformer<OUT,​IN>` - RTL 😞

subprojects/delivery/nupokati/src/main/kotlin/com/avito/android/provider/CdBuildConfigTransformer.kt
```diff
@@ -0,0 +1,37 @@
...
+
+    private fun deserializeToCdBuildConfig(configFile: File): CdBuildConfig {
+        require(configFile.exists()) { "Can't find cd config file in $configFile" }
+        return uploadCdGson.fromJson(configFile.reader(), CdBuildConfig::class.java)
+    }
+}
+
+internal val uploadCdGson: Gson by lazy {
+    GsonBuilder()
+        .setFieldNamingPolicy(FieldNamingPolicy.LOWER_CASE_WITH_UNDERSCORES)
```
eugene-krivobokov, line 33: Does it mean that we have values without `@SerializedName`?
I'd prefer to describe it explicitly and don't rely on any naming policy. It's too easy to break.

subprojects/delivery/nupokati/src/main/kotlin/com/avito/android/google_play/DeployBundleToGooglePlayTask.kt
```diff
@@ -0,0 +1,51 @@
...
+    public abstract val track: Property<CdBuildConfig.Deployment.Track>
+
+    @get:Input
+    public abstract val applicationId: Property<String>
+
+    @TaskAction
+    public fun doWork() {
+        val deployer = createGooglePlayDeployer(
+            jsonKey = googlePlayKeyFile.get().asFile,
+            logger = logger
```
eugene-krivobokov, line 36: Are we intentionally passing task logger here? Is it important?
If I'm not mistaken, any class can obtain it by itself:
```kotlin
org.slf4j.LoggerFactory.getLogger(class)
```

subprojects/delivery/nupokati/src/main/kotlin/com/avito/android/model/CdBuildConfig.kt
```diff
@@ -0,0 +1,45 @@
...
+        ) : Deployment()
+
+        /**
+         * Deploy artifacts to QApps.
+         * Uses UploadToQapps build step to find them.
+         */
+        public data class Qapps(
+            /**
+             * Send artifacts as release versions.
+             * Non-release artifacts stored limited time.
```
eugene-krivobokov:
```suggestion
             * Non-release artifacts are stored limited time.
```

subprojects/delivery/nupokati/src/main/kotlin/com/avito/android/model/CdBuildConfig.kt
```diff
@@ -0,0 +1,45 @@
...
+             */
+            val isRelease: Boolean
+        ) : Deployment()
+
+        public data class Unknown(val type: String) : Deployment()
+
+        public enum class Track {
```
eugene-krivobokov, line 38: Shouldn't this model be inside Google Play model?
Otherwise, it looks like it can be used inside any deployment.

subprojects/delivery/nupokati/src/main/kotlin/com/avito/android/provider/CdBuildConfigTransformer.kt
dsvoronin: I don't know why they preferred it that way

subprojects/delivery/nupokati/src/main/kotlin/com/avito/android/provider/CdBuildConfigTransformer.kt
dsvoronin: Yep, i just copied what we have already
It's really less obvious, but i see a hidden profit: it will be a little easier to migrate from gson that way

subprojects/delivery/nupokati/src/main/kotlin/com/avito/android/google_play/DeployBundleToGooglePlayTask.kt
dsvoronin: GooglePlayDeployer is planned to be removed this Q and implemented on service side
I'd skip tuning here

subprojects/delivery/nupokati/src/main/kotlin/com/avito/android/model/CdBuildConfig.kt
dsvoronin: Thanks
also `for a limited time` is better, right?

subprojects/delivery/nupokati/src/main/kotlin/com/avito/android/model/CdBuildConfig.kt
dsvoronin: I'd skip it for same reasons: https://github.com/avito-tech/avito-android/pull/1307#discussion_r784019736

subprojects/delivery/nupokati/src/main/kotlin/com/avito/android/google_play/DeployBundleToGooglePlayTask.kt
eugene-krivobokov: Ok

subprojects/delivery/nupokati/src/main/kotlin/com/avito/android/model/CdBuildConfig.kt
eugene-krivobokov: yes

subprojects/delivery/nupokati/build.gradle.kts
```diff
@@ -0,0 +1,35 @@
...
+    implementation(projects.subprojects.logger.slf4jGradleLogger)
+    implementation(projects.subprojects.testRunner.instrumentationTests)
+    implementation(projects.subprojects.testRunner.reportViewer)
+
+    gradleTestImplementation(projects.subprojects.gradle.testProject)
+}
+
+gradlePlugin {
+    plugins {
+        create("artifactory") {
```
dsvoronin: ><

dsvoronin: fixed

eugene-krivobokov (approved):

subprojects/delivery/nupokati/src/main/kotlin/com/avito/android/ArtifactoryExtension.kt
```diff
@@ -0,0 +1,24 @@
...
+    public abstract val login: Property<String>
+
+    public abstract val password: Property<String>
+
+    /**
+     * This parameters used to create path in artifactory repository
+     * and differs from one that comes from json cd build config
+     * There is no real reason of that
+     *
+     *  todo merge, take one from json
```
dsvoronin, line 18: That should be fixed without any changes in service contract, but it's a risky one, because we will start to write files in another directory/repository in artifactory.
There is a possibility that someone already used that paths, but a small one, because it's designed to be used only by service itself

I will do a change, on next phase
All old artifacts can't be moved without changing result json's accordingly, so maybe i should not bother

PR #1308, opened 2022-01-14
MBS-12172 Rename isRealRun function

RuslanMingaliev:


dsvoronin:
I remember how @sboishtyan changed it `isFakeOrchestratorRun` -> `isRealRun`

sergioboi:
Hakuna matata

subprojects/test-runner/test-instrumentation-runner/src/main/kotlin/com/avito/android/runner/OrchestratorDelegate.kt
```diff
@@ -5,6 +5,13 @@ import androidx.test.internal.runner.RunnerArgsAccessor
 
 interface OrchestratorDelegate {
 
+    @Deprecated(
```
sergioboi, line 8: @RuslanMingaliev Do you really need this change?

RuslanMingaliev: don't think so

PR #1310, opened 2022-01-19
MBS-12598 Add emcee QueueApi.kt

sergioboi:


subprojects/emcee/queue-api/src/main/kotlin/com/avito/emcee/queue/Device.kt
```diff
@@ -0,0 +1,15 @@
...
+import com.squareup.moshi.Json
+
+public data class Device(
+    /**
+     * Should be deleted
+     */
+    @Json(name = "deviceType")
+    val type: String,
+    /**
+     * Representing a emulator api version i.e. 21,22,29,30 etc.
```
eugene-krivobokov:
```suggestion
     * Representing an emulator API version: 21,22,29,30 etc.
```

subprojects/emcee/queue-api/src/main/kotlin/com/avito/emcee/queue/Device.kt
```diff
@@ -0,0 +1,15 @@
...
+     */
```
eugene-krivobokov, line 13: Let's use `@Json` for all json fields.
It's too easy to break a contract while refactoring.

subprojects/emcee/queue-api/src/main/kotlin/com/avito/emcee/queue/Job.kt
```diff
@@ -0,0 +1,28 @@
+package com.avito.emcee.queue
+
+import com.squareup.moshi.Json
+
+public data class Job(
+    @Json(name = "jobId")
+    val id: String,
+    /**
+     * Uses to prioritize inside the group. Higher will schedule faster
```
eugene-krivobokov, line 9: Are all integer values valid?

subprojects/emcee/queue-api/src/main/kotlin/com/avito/emcee/queue/TestTimeoutConfiguration.kt
```diff
@@ -0,0 +1,6 @@
+package com.avito.emcee.queue
+
+public data class TestTimeoutConfiguration(
+    val singleTestMaximumDuration: Int,
+    val testRunnerMaximumSilenceDuration: Int
```
eugene-krivobokov, line 5:
```suggestion
    val testRunnerMaximumSilenceDurationSec/Ms/...: Int
```
Let's clarify units if we can't use better types here and convert them by moshi.

eugene-krivobokov (approved):

subprojects/emcee/queue-api/src/main/kotlin/com/avito/emcee/queue/TestTimeoutConfiguration.kt
sergioboi: This part will be changed

subprojects/emcee/queue-api/src/main/kotlin/com/avito/emcee/queue/Job.kt
sergioboi: @beefon this question is for you

subprojects/emcee/queue-api/src/main/kotlin/com/avito/emcee/queue/Device.kt
sergioboi: I think it's an overhead. I think we will rarely change those models. And they are placed in a separate module

eugene-krivobokov: The motivation is don't even think about it while changing any code

dsvoronin (approved):

subprojects/emcee/queue-api/src/main/kotlin/com/avito/emcee/queue/Job.kt
beefon: Expected to be in range of `[0, 999]`

subprojects/emcee/queue-api/src/main/kotlin/com/avito/emcee/queue/TestExecutionBehavior.kt
```diff
@@ -0,0 +1,9 @@
+package com.avito.emcee.queue
+
+import com.squareup.moshi.Json
+
+public data class TestExecutionBehavior(
+    val environment: Map<String, String>,
+    @Json(name = "numberOfRetries")
+    val retries: Int
```
beefon, line 8: There are some new fields out there
https://github.com/avito-tech/Emcee/blob/master/Sources/RunnerModels/TestExecutionBehavior.swift

If you don't need them, I guess you just don't parse them.

PR #1311, opened 2022-01-20
MBS-12124 nupokati plugin part 2

dsvoronin:
integration test
more assertions needed (next PR)


subprojects/common/test-okhttp/src/main/kotlin/com/avito/test/http/MockDispatcher.kt
```diff
@@ -46,7 +49,13 @@ public class MockDispatcher(
         val matchedMock = synchronized(mocks) {
             mocks.findLast { it.requestMatcher.invoke(requestData) }
         }
-        val response = matchedMock?.response ?: unmockedResponse
+        val response = if (matchedMock?.response != null) {
+            logger.info("Request matched: [$requestData], answering: ${matchedMock.response}")
```
eugene-krivobokov, line 53: It looks like a debug level

PR #1312, opened 2022-01-21
MBS-10660 Delete checking plugin id equals to project.name

sergioboi:


build-logic/publication/src/main/kotlin/convention.publish-gradle-plugin.gradle.kts
```diff
@@ -13,21 +13,6 @@ gradlePlugin {
 
 val pluginPrefix = group.toString()
 
-val moduleNamingExceptions = setOf(
-    "legacy-signer", // will be removed in favor of new signer plugin
-)
-
-fun checkModuleToPluginNameConsistency(pluginName: String, pluginDeclaration: PluginDeclaration) {
```
sergioboi, line 20: Discussed with @eugene-krivobokov 
As we understand `artifactId` is overridden by pluginName. So artifact will be published with right id
And we don't understand the sense of this check

eugene-krivobokov: To clarify the idea.
plugin id is the source of truth. It's used to make maven artifact id. Clients know about it and do the same inverted logic.

dsvoronin (approved):

PR #1315, opened 2022-01-26
MBS-12124 pass artifacts in CdBuildResult

dsvoronin:


gradle/libs.versions.toml
```diff
@@ -17,7 +17,7 @@ minSdk = "21"
 junit5 = "5.8.0"
 junit5Platform = "1.8.0"
 hamcrest = "2.2"
-moshi = "1.13.0"
+moshi = "1.12.0"
```
dsvoronin, line 20: version downgraded, because of kotlin 1.6 dependency, which is unsupported in avito yet
seems like 1.12.0 has all features needed except for KSP, which is not yet used

PR #1317, opened 2022-01-27
MBS-11520 update AGP to 7.1

dsvoronin:


build-logic/publication/src/main/kotlin/convention.publish-android-library.gradle.kts
```diff
@@ -17,13 +17,14 @@ configure<LibraryExtension> {
...
     var registeredVariants = 0
 
+    // todo use new publishing: https://developer.android.com/studio/releases/gradle-plugin#build-variant-publishing
     libraryVariants
         .matching {
             allVariantNames += it.name
             it.name == publishExtension.variant.get()
         }
         .whenObjectAdded {
-            publishing {
```
dsvoronin, line 26: android library now have it's own publishing extension: https://developer.android.com/studio/releases/gradle-plugin#build-variant-publishing

subprojects/delivery/sign-service/src/main/kotlin/com/avito/android/signer/SignServicePlugin.kt
```diff
@@ -68,20 +62,4 @@ public class SignServicePlugin : Plugin<Project> {
             }
         }
     }
-
-    private fun checkSigningConfig(variant: ApplicationVariant) {
-        require(variant.signingConfig == null) {
```
dsvoronin, line 73: signingConfig can't be null anymore, despite comment in javadoc (it's a mistake)

gradle/libs.versions.toml
```diff
@@ -1,8 +1,8 @@
 [versions]
```
eugene-krivobokov, line 1: We need also to update r8

eugene-krivobokov (approved):

gradle/libs.versions.toml
dsvoronin: removed a dependency, couldn't find usages

PR #1322, opened 2022-01-31
MBS-10301 [test-okhttp] add nothingCaptured check

dsvoronin:


subprojects/common/test-okhttp/src/main/kotlin/com/avito/test/http/RequestCapturer.kt
```diff
@@ -24,10 +24,22 @@ public class RequestCapturer(
 
         public fun singleRequestCaptured(): RequestChecks = waitForAssertion {
             synchronized(this@RequestCapturer) {
-                assertThat("", requests.size == 1)
+                assertThat(
+                    "Single request should be captured, Currently matched: $requests",
+                    requests.size == 1
```
eugene-krivobokov, line 29: Could `hasSize(1)` matcher give better error message?

Kutashov: Btw, should this single size check be here? 
Sometimes there are multiple requests at the same time, that have similar paths, and I have to do some workarounds to bypass this check.

Kutashov (approved):

subprojects/common/test-okhttp/src/main/kotlin/com/avito/test/http/RequestCapturer.kt
eugene-krivobokov: I have an alternative proposal.

Why don't we give a list of all requested requests?
The client can easily check collection by any condition.
Trying to express all use cases without accessing requests bloats API.

Kutashov: Yeah, I would partially agree with that. 
RequestChecks interface seems to be trivial and might be eliminated. 
But FormUrlEncodedBodyChecks should be preserved as it hides raw data internals.

dsvoronin: I'd merge this PR, because discussion is out of it's scope

Btw, do you want to make such change? It would be be a nice task to do

Kutashov: @dsvoronin I'm not quite sure, this's something crucial to do now. 
But I'll bear in mind this task, we just need to figure out a way to do this

PR #1325, opened 2022-02-02
MBS-12634 emcee plugin extension

dsvoronin:


subprojects/emcee/action/src/main/kotlin/com/avito/emcee/EmceeTestActionFactory.kt
```diff
@@ -8,8 +8,8 @@ import com.avito.emcee.queue.QueueApi
 import com.avito.emcee.queue.QueueApi.Companion.create
 import retrofit2.Retrofit
 
-public class EmceeTestActionFactory {
-    private val queueApi: QueueApi = Retrofit.Builder().create()
+public class EmceeTestActionFactory(emceeQueueBaseUrl: String) {
```
eugene-krivobokov, line 11: We could use URL here to validate input

subprojects/emcee/gradle-plugin/src/main/kotlin/com/avito/emcee/internal/ApplicationVariantConfigurator.kt
```diff
@@ -0,0 +1,29 @@
...
+    override fun configure(task: EmceeTestTask) {
+        val androidTestVariant = variant.androidTest
+
+        requireNotNull(androidTestVariant) {
+            "Variant '$variant' is used to run instrumentation tests, " +
+                "but androidTest configuration is disabled.\n" +
+                "Please enable it in config to run tests.\n" +
+                "See: https://developer.android.com/reference/tools/gradle-api/7.0" +
+                "/com/android/build/api/variant/ApplicationVariant" +
+                "#androidTest:com.android.build.api.component.AndroidTest"
```
eugene-krivobokov, line 19: Multiline string instead of concatenation would be more conventional

subprojects/emcee/action/src/main/kotlin/com/avito/emcee/EmceeTestActionFactory.kt
dsvoronin: Well, it will be validated on lower level, during service construction
This layer doesn't have direct access to okhttp classes, i decided not to broaden it's visibility just for that

eugene-krivobokov: Ok, I meant `java.net.URL` rather than okhttp classes.

subprojects/emcee/action/build.gradle.kts
```diff
@@ -1,6 +1,7 @@
 plugins {
     id("convention.kotlin-jvm")
     id("convention.publish-kotlin-library")
+    kotlin("kapt") // TODO replace with ksp
```
sergioboi, line 4: I don't get why we need the `kapt` here

subprojects/emcee/action/src/main/kotlin/com/avito/emcee/EmceeTestActionConfig.kt
```diff
@@ -15,4 +21,16 @@ public class EmceeTestActionConfig(
     public val devices: List<Device>,
     public val apk: File,
     public val testApk: File,
-)
+) {
+
+    public companion object {
+
+        @OptIn(ExperimentalStdlibApi::class)
+        public fun createMoshiAdapter(): JsonAdapter<EmceeTestActionConfig> {
```
sergioboi, line 29: If it's public then this module should provide `moshi` as the `api`

subprojects/emcee/action/build.gradle.kts
sergioboi: I get

subprojects/emcee/action/src/main/kotlin/com/avito/emcee/EmceeTestActionConfig.kt
sergioboi: I fixed that in my PR

subprojects/emcee/gradle-plugin/src/main/kotlin/com/avito/emcee/EmceeExtension.kt
```diff
@@ -0,0 +1,25 @@
...
+
+import org.gradle.api.file.DirectoryProperty
+import org.gradle.api.provider.ListProperty
+import org.gradle.api.provider.Property
+import org.gradle.api.tasks.Nested
+import java.time.Duration
+
+public interface EmceeExtension {
+
+    @get:Nested
```
sergioboi, line 11: Why do we need this annotation here?

subprojects/emcee/gradle-plugin/src/main/kotlin/com/avito/emcee/EmceeTestTask.kt
```diff
@@ -29,41 +31,55 @@ public abstract class EmceeTestTask : DefaultTask() {
...
+
+    @get:InputDirectory
+    public abstract val testApk: DirectoryProperty
+
+    @get:Input
+    public abstract val configTestMode: Property<Boolean>
 
-    @get:InputFile
-    public abstract val testApk: Property<RegularFile>
+    @get:OutputDirectory
```
sergioboi, line 49: This will be used for caching but we used this prop only in testing.
I propose to mark property `Internal`

dsvoronin: `@Internal` until we decide how to store report as output?

subprojects/emcee/gradle-plugin/src/main/kotlin/com/avito/emcee/EmceeExtension.kt
dsvoronin: My understanding is to tell real EmceeExtension instance creator that it should also take nested bean into account, i don't know for what exactly. Maybe to see it's annotations. 
But we only use annotations in tasks to mark proeprties. Maybe it's not needed here, i'll try to remove, we have an integration test already

dsvoronin: Yep, it's not working without `@Nested`:

```
Caused by: org.gradle.internal.instantiation.ClassGenerationException: Could not generate a decorated class for 
...
Caused by: java.lang.IllegalArgumentException: Cannot have abstract method EmceeExtension.getJob()
```

subprojects/emcee/gradle-plugin/src/main/kotlin/com/avito/emcee/EmceeTestTask.kt
sergioboi: Until it behaves as an output with Gradle output semantics

subprojects/emcee/gradle-plugin/src/main/kotlin/com/avito/emcee/EmceeExtension.kt
sergioboi: Thank you for trying

PR #1333, opened 2022-02-09
MBS-12659 Add Worker Http client

sergioboi:


gradle/libs.versions.toml
```diff
@@ -119,7 +119,7 @@ detektFormatting = { module = "io.gitlab.arturbosch.detekt:detekt-formatting", v
 # https://github.com/robolectric/robolectric/releases
 robolectric = "org.robolectric:robolectric:4.7.1"
 
-moshiKotlin = { module = "com.squareup.moshi:moshi-kotlin", version.ref = "moshi" }
```
sergioboi, line 122: We don't need the `moshiKotlin`. It is needed only for reflection without code generation

settings.gradle.kts
```diff
@@ -138,6 +138,8 @@ include(":subprojects:delivery:upload-cd-build-result")
 include(":subprojects:delivery:upload-to-googleplay")
 
 include(":subprojects:emcee:gradle-plugin")
-include(":subprojects:emcee:queue-backend-api")
+include(":subprojects:emcee:queue-client-api")
```
sergioboi, line 141: The module resctructure.
Modules with `-models` contain only models
Modules with `-api` contain models and retrofit service
`client` related for queue client - gradle-plugin

subprojects/emcee/gradle-plugin/build.gradle.kts
```diff
@@ -7,11 +7,12 @@ plugins {
 dependencies {
     implementation(projects.subprojects.emcee.action)
     implementation(libs.androidGradle)
-    implementation(libs.moshiKotlin)
-    implementation(libs.kotlinReflect) {
-        because("moshi 1.12.0 depend on 1.4.x kotlin, and 1.13 on 1.6.x, we use 1.5")
+    implementation(libs.moshiAdapters) {
```
sergioboi: I don't where this module took these deps before

subprojects/emcee/queue-worker-api/src/main/kotlin/com/avito/emcee/worker/GetBucketResponse.kt
```diff
@@ -0,0 +1,31 @@
...
+import com.avito.emcee.queue.BuildArtifacts
+import com.avito.emcee.queue.Device
+import com.avito.emcee.queue.TestEntry
+import com.avito.emcee.queue.TestExecutionBehavior
+import com.avito.emcee.queue.TestTimeoutConfiguration
+import com.squareup.moshi.Json
+import com.squareup.moshi.JsonClass
+import dev.zacsweers.moshix.sealed.annotations.TypeLabel
+
+@JsonClass(generateAdapter = true, generator = "sealed:caseId")
```
sergioboi, line 12: @dsvoronin copy-pasting your code

dsvoronin: https://github.com/ZacSweers/MoshiX/tree/main/moshi-sealed
i'm using this approach

gradle/libs.versions.toml
dsvoronin: good find. i didn't noticed that
also, not the best artifact name ) should be something with reflection in it i think

subprojects/emcee/action/build.gradle.kts
```diff
@@ -5,16 +5,15 @@ plugins {
 }
 
 dependencies {
-    api(projects.subprojects.emcee.shared)
-    implementation(projects.subprojects.emcee.queueBackendApi)
+    api(projects.subprojects.emcee.queueClientModels)
+    api(libs.moshiAdapters) {
+        "used because EmceeConfig serialized to JSON for testing. Replace as soon as possible"
```
dsvoronin, line 10: but replaced how? do you have an idea?

dsvoronin (approved):

subprojects/emcee/action/build.gradle.kts
sergioboi: No. But I want to leave an intention. If I could I would replace

PR #1335, opened 2022-02-13
MBS-12582 Rewrite bootstrapping docker image builder to Kotlin [1/3]

eugene-krivobokov:
Rewrote image that builds other Docker images.
Now it has no Python.

In the next PRs:
- Support building of emulators
- Rebuild all images to check migration to a new builder


ci/docker/local_build.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
```
eugene-krivobokov, line 1: Added new script to local build image.

ci/docker/image-builder/src/main/kotlin/ru/avito/image_builder/Main.kt
```diff
@@ -0,0 +1,121 @@
...
+
+    @OptIn(ExperimentalCli::class)
+    @JvmStatic
+    public fun main(args: Array<String>) {
+        val parser = ArgParser(
+            programName = "image-builder",
+        )
+
+        parser.subcommands(
+            BuildImage("build", "Build image"),
```
eugene-krivobokov, line 108: Added new entrypoint mostly for local builds.

Previously, it was an implicit behavior: you can build an image locally without Docker registry credentials and publishing will fail.

Now it's more straightforward. You can't run publish without credentials.

ci/docker/image-builder/src/main/kotlin/ru/avito/image_builder/internal/ImageBuildProperties.kt
```diff
@@ -0,0 +1,42 @@
+package ru.avito.image_builder.internal
+
+import java.io.File
+import java.util.Properties
+
+/**
+ * Properties file to store associated attributes for image
+ */
```
eugene-krivobokov, line 8: The same idea as before.
It's used to store metadata in the code rather than passing all by arguments.

```
./publish.sh <directory with Dockerfile>
```

Changes:
- Changed format from YAML to properties. 
Yaml is excessive and requires external dependencies to parse.

ci/docker/documentation/docker-build.properties
```diff
@@ -0,0 +1 @@
+image-name:android/documentation
```
eugene-krivobokov: https://github.com/avito-tech/avito-android/pull/1335/files#r805346137

ci/docker/image-builder/build.gradle.kts
```diff
@@ -0,0 +1,53 @@
+import org.jetbrains.kotlin.gradle.dsl.KotlinJvmProjectExtension
+
```
eugene-krivobokov, line 2: Here is a module with a CLI app that builds Docker images. 

As before, it's completely independent of any code in the repository.
Otherwise, we should rebuild all images on any change in the repository to be sure.

ci/docker/image-builder/build.gradle.kts
```diff
@@ -0,0 +1,53 @@
...
+}
+
+dependencies {
+    implementation(platform("org.jetbrains.kotlin:kotlin-bom"))
+    implementation("org.jetbrains.kotlinx:kotlinx-cli:0.3.3")
+
+    testImplementation("org.junit.jupiter:junit-jupiter:5.8.0")
+
+    integTestImplementation("org.junit.jupiter:junit-jupiter:5.8.0")
+}
```
eugene-krivobokov, line 20: As minimum external dependencies as possible to have fewer points of change over time.

ci/docker/image-builder/src/main/kotlin/ru/avito/image_builder/internal/RegistryLogin.kt
```diff
@@ -0,0 +1,26 @@
...
+
+    fun login() {
+        val result = docker.login(
+            credentials.username,
+            credentials.password,
+            credentials.registry
+        )
+        check(result.isSuccess) {
+            "Login to registry by user ${credentials.username} failed: ${result.exceptionOrNull()}"
+        }
```
eugene-krivobokov, line 21: https://github.com/avito-tech/avito-android/pull/1335#discussion_r805345515

ci/docker/image-builder/src/main/kotlin/ru/avito/image_builder/internal/docker/CliDocker.kt
```diff
@@ -0,0 +1,65 @@
+package ru.avito.image_builder.internal.docker
+
+import ru.avito.image_builder.internal.process.CommandExecutor
+import java.time.Duration
+
+internal open class CliDocker(
+    private val executor: CommandExecutor = CommandExecutor()
+) : Docker {
+
```
eugene-krivobokov, line 9: Why not use any JVM client library?

I've tried to use https://github.com/docker-java/docker-java and found this approach overcomplicated because of many additional layers of abstraction. It takes too much time to figure out how to use them.
This is why I've decided to use CLI API. It's stable and simple enough.

ci/docker/build.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+
```
eugene-krivobokov, line 2: Added new script to local build image.

docs/content/ci/Containers.md
```diff
@@ -8,11 +8,6 @@ avito-disclaimer.md
 
 Все образы расположены в `ci/docker`.
 
-## Android SDK image
```
eugene-krivobokov, line 11: There is no such image. It was deleted.

ci/docker/image-builder/Dockerfile
```diff
@@ -0,0 +1,34 @@
+FROM openjdk:11 as BUILD
```
dsvoronin, line 1: we should provide an ability to use "inhouse" image here

dsvoronin (approved):

ci/docker/image-builder/Dockerfile
eugene-krivobokov: Agree.

I see only one option - copy this base image to the inhouse registry and build upon it.

I'd like to propose a simplification. Let's use _only_ inhouse registry in the whole build chain:

```
openjdk (dockerhub) --> inhouse openjdk --> image-builder

ubuntu (dockerhub) --> inhouse ubuntu --> android-builder\...
```

Rationale:

- Switching between registries while building seems more complex. We need to ensure every time that images in different registries are identical and updated simultaneously.
- Uploading to DockerHub complicates the whole process and is not automated yet.
- Last, but not least: I don't see for whom we support DockerHub. Images are by design reproducible and everyone could build them by themselves.

eugene-krivobokov: I'd like to discuss it in MBS-12767 because I don't want to make half-measures only for image-builder.
We have the same issue in other images (emulator, android-builder, ...).

eugene-krivobokov:
I've merged to fix android-builder image building.
Feel free to comment. I'll fix it in the next PR.

ci/docker/build.sh
sergioboi: Don't get what `to local build image` means

ci/docker/image-builder/.dockerignore
```diff
@@ -0,0 +1,6 @@
+# Include only app sources for build
```
sergioboi, line 1: I need explanation for what that file exists

ci/docker/image-builder/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+
+RUN mkdir /app
+COPY . /app
+WORKDIR /app
+
+# Skip integration tests because they use Docker.
+# This is impossible while building the image.
+# We can't mount /var/run/docker.sock.
+RUN ./gradlew build -x integrationTest
```
sergioboi, line 10: I don't get this line.
I prefer to add a description to that file. What and how it builds

ci/docker/image-builder/README.md
```diff
@@ -0,0 +1,4 @@
+Docker image to build other images.
```
sergioboi, line 1: I'm not sure that Readme better then comment inside dockerfile

ci/docker/image-builder/src/integTest/java/ru/avito/image_builder/internal/CliDockerTest.kt
```diff
@@ -0,0 +1,86 @@
+package ru.avito.image_builder.internal
```
sergioboi, line 1: I don't get what are you testing there

eugene-krivobokov: Tests check Docker client implementation.
It uses CLI API. The weak point is passing arguments and parsing output.
Thus, I want to be sure at least in basic scenarios.

ci/docker/image-builder/Dockerfile
eugene-krivobokov: This command runs a Gradle task to build the app.
This project is [a conventional CLI app](https://docs.gradle.org/current/userguide/application_plugin.html)

> to that file

I don't get it. What file?

ci/docker/image-builder/.dockerignore
eugene-krivobokov: https://docs.docker.com/engine/reference/builder/#dockerignore-file
This is a standard config to reduce the amount of copied files while building an image.

ci/docker/build.sh
eugene-krivobokov: I mean this scenario:

1. I've made changes in image X and want to check them locally.
2. I run `./build.sh X` and get this X image.
3. Repeat 1-2 if needed.
4. Push changes

ci/docker/image-builder/src/main/kotlin/ru/avito/image_builder/internal/ImageBuildProperties.kt
eugene-krivobokov: I'm thinking about removing this feature.

Motivation:
- These configs are lying already. E.g. `android-builder` has different names in the in-house registry and Docker hub. It means, that these configs are not the source of truth and make more confusion. Maybe, it can be fixed and restricted, but I don't see why image content defines image name in all registries. The only store this huge binary.
- Having simple and straightforward logic is more important for me than explicitly storing one additional argument in CI scripts.

eugene-krivobokov: #1341 
Now it's simpler

ci/docker/image-builder/Dockerfile
eugene-krivobokov: I've added an explanation of working with the project to README.
This project is shipped in a container but before we work with it like with any other project.

https://github.com/avito-tech/avito-android/pull/1341/files#diff-46c7cb3f9fc9566533c7dd6923e8338086828b6f91a8ad244b96fe7090ba585c

ci/docker/image-builder/README.md
eugene-krivobokov: It depends on the scenario.
Information inside Dockerfile doesn't help with scenarios when I'd like to run this project without building the image.

- Moved info about dependencies into buildscript.
- Added info about working with the project

#1341

sergioboi: So add to Docker file `see details in README file`

ci/docker/image-builder/.dockerignore
sergioboi: Thank you. I will take it into account

ci/docker/image-builder/README.md
eugene-krivobokov: Got it. Will add in the next PR

PR #1344, opened 2022-02-23
Fix graddle security issue

Chromico:
This fixes this security issue: https://security.gradle.com/advisory/2021-11


eugene-krivobokov:
As I see, the advisory is about the server-side of Enterprise that works in cloud in our case.
Client-side (the plugin) works while building and here is no surface of attack. 
Also, the changelog for the plugin itself shows only minor fixes: https://plugins.gradle.org/plugin/com.gradle.enterprise/3.7.2

Anyway, these improvements are helpful. Thanks for the PR!

PR #1345, opened 2022-02-24
MBS-12648 Add RV support for the Robolectric tests

RuslanMingaliev:


subprojects/test-runner/robolectric-inhouse-runner/src/main/kotlin/com/avito/robolectric/runner/description/RobolectricDescriptionMetaDataParser.kt
```diff
@@ -0,0 +1,23 @@
...
+
+        return DescriptionMetaData(
+            environment = "Robolectric-${result.groupValues[2]}",
+            testName = result.groupValues[1],
+            className = result.groupValues[3]
+        )
+    }
+
+    private companion object {
+        private val REGEX = Regex("([\\s\\S]*)\\[(.*)]\\((.*)\\)")
```
eugene-krivobokov, line 21: Using `.*` between two delimiters is greedy and will capture text between multiple pairs.
E.g. regex `\[.+\]` in `[a] text between [b] text between [c]` string will match the entire string as only one match.

If you need to capture text only between delimiters, you should exclude delimiter by `[^delimiter]` instead of using any symbol - `.`.
For the previous example, `\[[^\]]+\]` will give three matches - `[a]`, `[b]`, `[c]`.

If we don't control the count of pairs ("[22]" as I see), it might be more robust to find them all and filter manually unneeded.

subprojects/test-runner/test-report-android/src/main/kotlin/com/avito/android/test/report/troubleshooting/TroubleshooterFactory.kt
```diff
@@ -0,0 +1,13 @@
+package com.avito.android.test.report.troubleshooting
+
+import com.avito.android.test.report.troubleshooting.dump.MainLooperQueueDumper
+import com.avito.android.test.report.troubleshooting.dump.ViewHierarchyDumper
+
+object TroubleshooterFactory {
```
eugene-krivobokov, line 6: Why do we need this factory?
It adds nothing to the builder.

eugene-krivobokov:
Could you please visualize changes in modules dependencies and their responsibility?
It's hard to get it from text diff.

subprojects/test-runner/test-report-jvm/src/main/kotlin/com/avito/test/report/arguments/PropertiesArgsProvider.kt
```diff
@@ -0,0 +1,24 @@
+package com.avito.test.report.arguments
+
+import com.avito.android.test.report.arguments.ArgsProvider
+import java.io.Serializable
+import java.util.Properties
+
+public class PropertiesArgsProvider(private val properties: Properties) : ArgsProvider {
+
+    override fun getOptionalArgument(name: String): String? {
+        return properties.getProperty(name, "").let { it.ifBlank { null } }
```
eugene-krivobokov, line 10: It will return null if the property is an empty string.
We should check the existence of property and don't make such unsafe assumptions.

subprojects/test-runner/test-report/src/main/kotlin/com/avito/android/test/report/InternalReport.kt
```diff
@@ -1,18 +1,16 @@
...
 import com.avito.android.test.report.model.TestMetadata
 
-interface InternalReport : ReportTestLifecycle<StepResult> {
+public interface InternalReport : ReportTestLifecycle<StepResult> {
 
-    @VisibleForTesting
-    val currentState: ReportState
+    public val currentState: ReportState
 
     // delete init
```
eugene-krivobokov, line 10: What about this comment? Is it still valid?

subprojects/test-runner/test-report/src/main/kotlin/com/avito/android/test/report/arguments/ArgsProvider.kt
```diff
@@ -0,0 +1,17 @@
...
+     * Blank strings also handled and will return null
+     */
+    public fun getOptionalArgument(name: String): String?
+
+    public fun getMandatoryArgument(name: String): String
+
+    public fun <T : Serializable> getMandatorySerializableArgument(name: String): T
+
+    public fun <T : Serializable> getOptionalSerializableArgument(name: String): T?
+}
```
eugene-krivobokov, line 17: I found `getOptionalX(): T? / getMandatoryX(): T` naming pattern too verbose and propose to use another one.

- `getX(): T / getXorNull(): T?`
- Or `getXorThrow(): T / getX(): T?` - this option is good because of `orThrow` important details

subprojects/test-runner/test-report/src/main/kotlin/com/avito/android/util/ReflectionExt.kt
```diff
@@ -1,10 +1,10 @@
 package com.avito.android.util
 
-internal inline fun <reified T> Any.getFieldValue(field: String): T = javaClass.getDeclaredField(field)
+public inline fun <reified T> Any.getFieldValue(field: String): T = javaClass.getDeclaredField(field)
```
eugene-krivobokov, line 3: It doesn't look like `test-report` public API.

subprojects/test-runner/transport/src/main/kotlin/com/avito/android/transport/LocalRunTransport.kt
```diff
@@ -38,17 +37,13 @@ internal class LocalRunTransport(
     private val reportLinksGenerator: ReportLinksGenerator = ReportViewerLinksGeneratorImpl(
         reportViewerUrl = reportViewerUrl,
         reportCoordinates = reportCoordinates,
-        reportViewerQuery = ReportViewerQuery.createForAndroidRuntime()
+        reportViewerQuery = reportViewerQuery
     )
 
     override fun sendReport(state: Started) {
-        if (Looper.myLooper() == Looper.getMainLooper()) {
```
eugene-krivobokov, line 45: When will we check this contract?

docs/content/blog/2019/12/open-source-introduction.md
```diff
@@ -85,7 +85,7 @@ And even if nobody uses the project code in its original form, developers will b
 ### Learn by teaching
 
 Just moving the code into open source is not enough. 
-Practices, approaches, methods of troubleshooting, and making decisions — this is what is essential. 
+Practices, approaches, methods of troubleshooting, and making decisions — this is what is essential.
```
dsvoronin: additional space is probably needed to express newline, check pls

eugene-krivobokov: Only two spaces at the end give a line break.

dsvoronin (approved):

PR #1348, opened 2022-02-24
MBS-12691 Enable by default stable features

eugene-krivobokov:


subprojects/assemble/build-trace/src/main/kotlin/com/avito/android/build_trace/BuildTracePlugin.kt
```diff
@@ -21,7 +21,7 @@ public abstract class BuildTracePlugin : Plugin<Project> {
     }
 
     private fun registerListeners(project: Project, extension: BuildTraceExtension) {
-        val isEnabled = extension.enabled.getOrElse(false)
+        val isEnabled = extension.enabled.getOrElse(true)
```
eugene-krivobokov, line 24: The most common case is to apply and use plugins.
It's inconvenient to have it disabled by default.

subprojects/assemble/build-checks/src/main/kotlin/com/avito/android/build_checks/RootProjectChecksExtension.kt
```diff
@@ -73,7 +73,7 @@ public open class RootProjectChecksExtension : BuildChecksExtension() {
         }
 
         public open class PreventKotlinDaemonFallback : RootProjectCheck() {
-            override var enabled: Boolean = false
+            override var enabled: Boolean = true
```
eugene-krivobokov, line 76: Stable enough

PR #1352, opened 2022-07-06
Ma 2495 deeplink manifest generator plugin

midery:
**Problem**: 
1. Information about deeplink publicity is stored not in a code, but in android manifest. This leads to several problems:
   *  Harder to parse: if we need to extract information about deeplinks publicity we need to interfere at the process of merging android app manifest, and manually parse it.
   * Harder to validate: Manifest in avito is stored in single module with a reference to an activity, which is responsible for opening deeplinks, so we need to write all the validation logic in this module, or in root application module. If deeplinks are not stored in such module, and are divided within features, we may need to write extra logic to collect them from source code and merge all links between modules, which could be quite complicated. 
   * Dead code: manifest is not connected anyhow to a deeplink model from source code: if deeplink is removed from source code, manifest would not be updated as well. This would cause situtation, in which application will open such deeplink, but the page will be blank and no actions will be performed, and this would cause poor user experience. 
2. In avito, we are already collecting information about deeplinks by annotation processing, and do not to spread this information to another places. There should be just one source of truth. 

**Solution**: 
We've implemented deeplink generator gradle plugin, that should be applied to each module, which contains public deeplinks. This plugin generates AndroidManifest entries for all public deeplinks in a module:

```kotlin
deeplinkGenerator {
    
    //Screen class, that is used to open deeplinks. Could be supplied one time, by convention plugin.
    activityIntentFilterClass = "com.avito.deeplink_generator.SomeActivity" 

    //Default deeplink scheme. Could be supplied one time, by convention plugin
    defaultScheme = "ru.avito"
    
    //Public deeplinks for a library module. Should be supplied by a module itself.
    publicDeeplinks(
        "1/feed",
        "1/profile
    )
}
```

There is no way to validate deeplinks during manifest processing stage, because manifest processing for each module happens before this module's source code compilation, thus, manifest will be ready when we don't have any information about deeplinks from code. 
So, we've added `ValidatePublicDeeplinksTask` to validate that all deeplinks from `publicDeeplinks` section are also existing in code, and vice versa. This task should run after source code compilation, and should be connected to a build process manually by plugin users.


subprojects/gradle/deeplink-generator/src/gradleTest/kotlin/com/avito/deeplink_generator/ConfigurationCacheCompatibilityTest.kt
```diff
@@ -0,0 +1,65 @@
...
+            modules = listOf(
+                AndroidAppModule(
+                    name = "app",
+                    enableKotlinAndroidPlugin = false,
+                    dependencies = setOf(
+                        GradleDependency.Safe(
+                            GradleDependency.Safe.CONFIGURATION.IMPLEMENTATION,
+                            GradleDependency.Safe.Coordinate.Project(Path.path(":feed"))
+                        )
+                    ),
```
eugene-krivobokov, line 25: this could be simplified to 

```kotlin
dependencies = setOf(
    project(":feed")
)
```

subprojects/gradle/deeplink-generator/src/gradleTest/kotlin/com/avito/deeplink_generator/MergeDeeplinkManifestTest.kt
```diff
@@ -0,0 +1,213 @@
...
+        TestProjectGenerator(
+            name = "rootapp",
+            modules = listOf(
+                AndroidAppModule(
+                    name = "app",
+                    enableKotlinAndroidPlugin = false,
+                    dependencies = setOf(
+                        Safe(Safe.CONFIGURATION.IMPLEMENTATION, Safe.Coordinate.Project(Path.path(":feed"))),
+                        Safe(Safe.CONFIGURATION.IMPLEMENTATION, Safe.Coordinate.Project(Path.path(":profile"))),
+                    )
```
eugene-krivobokov, line 120: https://github.com/avito-tech/avito-android/pull/1352/files#r915595892

subprojects/gradle/deeplink-generator/src/gradleTest/kotlin/com/avito/deeplink_generator/ValidatePublicDeeplinksTest.kt
```diff
@@ -0,0 +1,142 @@
...
+import com.avito.test.gradle.plugin.plugins
+import org.gradle.testkit.runner.TaskOutcome
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.io.TempDir
+import java.io.File
+
+internal class ValidatePublicDeeplinksTest {
+
+    @Test
+    fun `assemble single library - validate same links - success`(@TempDir projectDir: File) {
```
eugene-krivobokov: https://english.stackexchange.com/a/235095

subprojects/gradle/deeplink-generator/src/main/kotlin/com/avito/deeplink_generator/DeeplinkGeneratorExtension.kt
```diff
@@ -0,0 +1,17 @@
...
+import com.avito.deeplink_generator.model.Deeplink
+
+public open class DeeplinkGeneratorExtension {
+
+    public var publicDeeplinks: Set<Deeplink> = emptySet()
+
+    public var defaultScheme: String = ""
+    public var activityIntentFilterClass: String = ""
+    public var validationCodeFixHint: String = ""
+
```
eugene-krivobokov: Instead of empty values it would be more conventionally to use Properties.
They can easily distinguish not set values and add default ones if needed.

subprojects/gradle/deeplink-generator/src/main/kotlin/com/avito/deeplink_generator/ValidatePublicDeeplinksTask.kt
```diff
@@ -0,0 +1,79 @@
...
+    private fun extractPublicDeeplinksFromCode(): Set<Deeplink> {
+        val publicDeeplinksFromCode = publicDeeplinksFromCode.orNull
+        val publicDeeplinksFromCodeFile = publicDeeplinksFromCodeFile.orNull?.asFile
+
+        return when {
+            publicDeeplinksFromCodeFile != null ->
+                DeeplinkFileParser.parse(publicDeeplinksFromCodeFile, defaultScheme.get())
+            publicDeeplinksFromCode != null -> publicDeeplinksFromCode
+            else -> error(
+                "You must supply either `publicDeeplinksFromCode` or `publicDeeplinksFromCodeFile` to run this task."
```
eugene-krivobokov: > You must supply ... to run this task

If I see this message, I won't understand how to do it.
Let's mention here the extension.

subprojects/gradle/deeplink-generator/src/main/kotlin/com/avito/deeplink_generator/internal/merge/MergeDeeplinkManifestTask.kt
```diff
@@ -0,0 +1,154 @@
...
+
+        private const val PUBLIC_DEEPLINK_MANIFEST_HEADER = """
+            <manifest xmlns:android="http://schemas.android.com/apk/res/android"
+                xmlns:tools="http://schemas.android.com/tools">
+            
+                <application>
+                    <activity
+                        android:name="%s"
+                        tools:node="merge">
+        """
```
eugene-krivobokov: `trimIndent()` ?

subprojects/gradle/deeplink-generator/src/main/kotlin/com/avito/deeplink_generator/internal/validator/PublicDeeplinksValidator.kt
```diff
@@ -0,0 +1,56 @@
...
+                validDeeplinks.add(link)
+            } else {
+                codeOnlyDeeplinks.add(link)
+            }
+        }
+
+        if (codeOnlyDeeplinks.isNotEmpty()) {
+            error(
+                """
+                    Deeplinks are marked as public in code, but not in buildScript.
```
eugene-krivobokov: `buildScript` -> `build script`

I haven't seen it as a term \ proper noun.

subprojects/gradle/deeplink-generator/src/main/kotlin/com/avito/deeplink_generator/model/Deeplink.kt
```diff
@@ -0,0 +1,34 @@
+package com.avito.deeplink_generator.model
+
+import java.io.Serializable
+
+public class Deeplink(
```
eugene-krivobokov, line 5: Will this model be used in clients' applications code?

midery: It may be used in client build scripts, as a type safe way to store public links information.

subprojects/gradle/deeplink-generator/src/gradleTest/kotlin/com/avito/deeplink_generator/ValidatePublicDeeplinksTest.kt
midery: Thanks for a hint, fixed

subprojects/gradle/deeplink-generator/src/gradleTest/kotlin/com/avito/deeplink_generator/ConfigurationCacheCompatibilityTest.kt
midery: Cool dsl, haven't saw it on a first look. Thanks!

subprojects/gradle/deeplink-generator/src/main/kotlin/com/avito/deeplink_generator/DeeplinkGeneratorExtension.kt
midery: Got it, I'll use Properties

subprojects/gradle/deeplink-generator/src/main/kotlin/com/avito/deeplink_generator/ValidatePublicDeeplinksTask.kt
midery: We're configuring inputs for this task not in extension, but in the task itself. 

publicDeeplinksFromCode / publicDeeplinksFromCodeFile may not exist when the extension is created. I'll try to rethink the way the links are supplied. Thanks for the feedback.

subprojects/gradle/deeplink-generator/src/gradleTest/kotlin/com/avito/deeplink_generator/MergeDeeplinkManifestTest.kt
```diff
@@ -0,0 +1,212 @@
...
+        gradlew(projectDir, ":feed:assembleRelease")
+            .assertThat()
+            .taskWithOutcome(":feed:mergeReleasePublicDeeplinkManifest", TaskOutcome.SUCCESS)
+        gradlew(projectDir, ":feed:assembleRelease")
+            .assertThat()
+            .taskWithOutcome(":feed:mergeReleasePublicDeeplinkManifest", TaskOutcome.UP_TO_DATE)
+    }
+
+    @Test
+    fun `assemble single library without deeplinks - manifest task not triggered`(@TempDir projectDir: File) {
```
eugene-krivobokov: `manifest task not triggered` -> `manifest with deeplinks is not generated` ?

task seems implementation details

subprojects/gradle/deeplink-generator/src/main/kotlin/com/avito/deeplink_generator/internal/merge/MergeDeeplinkManifestTask.kt
```diff
@@ -0,0 +1,154 @@
...
+    abstract val outputManifest: RegularFileProperty
+
+    @TaskAction
+    fun merge() {
+        val inputManifestFile = inputManifest.asFile.get()
+        val outputManifestFile = outputManifest.asFile.get()
+        val publicDeeplinkManifest = createPublicDeeplinkManifest(outputManifestFile.parentFile)
+        publicDeeplinkManifest.deleteOnExit()
+
+        mergeManifests(
```
eugene-krivobokov, line 54: Is it an official way to use an internal implementation here? Don't we have a public API?

eugene-krivobokov (approved):

subprojects/gradle/deeplink-generator/src/main/kotlin/com/avito/deeplink_generator/internal/merge/MergeDeeplinkManifestTask.kt
midery: Well, I didn't found any recommendations on how to merge two manifest files. 
An official example shows us that we could modify manifest https://developer.android.com/studio/build/extend-agp#supported-operations , but only as a plain text document. 

`mergeManifests` function is public, documented (a little bit) and used in gradle tasks to merge both application and library manifest. I believe, we can use it in our task as well.

subprojects/gradle/deeplink-generator/src/gradleTest/kotlin/com/avito/deeplink_generator/MergeDeeplinkManifestTest.kt
midery: Thanks, fixed the function name

RuslanMingaliev (approved):

PR #1359, opened 2022-07-12
MBSA-648 Collect JVM metrics for builds

eugene-krivobokov:


subprojects/assemble/build-metrics/src/gradleTest/kotlin/com/avito/android/plugin/build_metrics/jvm/JvmMetricsTest.kt
```diff
@@ -0,0 +1,61 @@
...
+import com.avito.android.stats.TimeMetric
+import com.avito.test.gradle.TestProjectGenerator
+import com.avito.test.gradle.plugin.plugins
+import com.google.common.truth.IterableSubject
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.io.TempDir
+import java.io.File
+
+internal class JvmMetricsTest {
```
eugene-krivobokov, line 14: The simplest integration test

subprojects/assemble/build-metrics/src/main/kotlin/com/avito/android/plugin/build_metrics/internal/jvm/Jcmd.kt
```diff
@@ -0,0 +1,111 @@
...
+import java.time.Duration
+
+internal class Jcmd(
+    private val processRunner: ProcessRunner,
+    javaHome: File = javaHome()
+) {
+
+    private val jcmd = File(javaHome, "bin/jcmd").path
+
+    fun gcHeapInfo(pid: Long): Result<HeapInfo> {
```
eugene-krivobokov, line 16: Here we gather information about heap and some of non-heap areas

subprojects/assemble/build-metrics/src/main/kotlin/com/avito/android/plugin/build_metrics/internal/jvm/Jps.kt
```diff
@@ -0,0 +1,38 @@
...
+import java.time.Duration
+
+internal class Jps(
+    private val processRunner: ProcessRunner,
+    javaHome: File = javaHome()
+) {
+
+    private val jps = File(javaHome, "bin/jps").path
+
+    fun run(): Result<Set<LocalVm.Unknown>> {
```
eugene-krivobokov, line 16: Here we look for Java processes

subprojects/assemble/build-metrics/src/main/kotlin/com/avito/android/plugin/build_metrics/internal/jvm/VmResolver.kt
```diff
@@ -0,0 +1,70 @@
...
+import java.util.stream.Collectors
+
+internal class VmResolver(
+    private val jps: Jps
+) {
+
+    /**
+     * Resolves Gradle processes with their descendants
+     */
+    fun localVMs(): Result<Set<LocalVm>> {
```
eugene-krivobokov, line 17: Here we look for all JVMs spawned by Gradle processes

subprojects/assemble/build-metrics/src/main/kotlin/com/avito/android/plugin/build_metrics/internal/teamcity/CollectTeamcityMetricsTask.kt
```diff
@@ -13,21 +12,21 @@ import org.gradle.api.tasks.TaskAction
 import org.gradle.workers.WorkerExecutor
 import javax.inject.Inject
 
-public abstract class CollectTeamcityMetricsTask @Inject constructor(
+internal abstract class CollectTeamcityMetricsTask @Inject constructor(
```
dsvoronin, line 15: Tasks are usually public, why do you think narrowing the access is needed here?

eugene-krivobokov: The contract is not to use tasks directly.

- If someone consumes this module as a dependency in another plugin
- If someone uses this class in buildscripts

Even if this modifier doesn't forbid usage, it can be a reminder like "internal" package.

subprojects/assemble/build-metrics/src/main/kotlin/com/avito/android/plugin/build_metrics/internal/jvm/Jcmd.kt
```diff
@@ -0,0 +1,111 @@
...
+        val result = processRunner.run("$jcmd $pid GC.heap_info", Duration.ofSeconds(5))
+        if (result.isFailure()) {
+            return Result.Failure((result as Result.Failure).throwable)
+        }
+        return parseHeapInfo(result.getOrThrow())
+    }
+
+    private fun parseHeapInfo(output: String): Result<HeapInfo> {
```
dsvoronin: i'd extract these private methods to to some `HeapInfoParser` for easier testing.
Same pattern for `jps`

subprojects/assemble/build-metrics/src/main/kotlin/com/avito/android/plugin/build_metrics/internal/jvm/Jps.kt
```diff
@@ -0,0 +1,38 @@
...
+        return processRunner.run("$jps -l", Duration.ofSeconds(5))
```
dsvoronin: Probably it's better to have timeouts extracted to constructor param with default value.
Here and in `jcmd`

subprojects/assemble/build-metrics/src/main/kotlin/com/avito/android/plugin/build_metrics/internal/jvm/Jps.kt
```diff
@@ -0,0 +1,38 @@
...
+        }
+        return when (parts.size) {
+            // If the target JVM is started with a custom launcher
+            1 -> LocalVm.Unknown(id = parts[0].toLong(), name = "Unknown")
+            else -> LocalVm.Unknown(id = parts[0].toLong(), name = parts[1])
+        }
+    }
+}
+
+internal fun javaHome(): File = File(System.getProperty("java.home"))
```
dsvoronin, line 39: Is it set by gradle itself? Like passed `org.gradle.java.home`?

subprojects/assemble/build-metrics/src/main/kotlin/com/avito/android/plugin/build_metrics/internal/jvm/JvmMetricsSender.kt
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    private fun processName(vm: LocalVm): SeriesName {
+        val name = when (vm) {
+            is GradleDaemon -> "gradle_daemon"
+            is GradleWorker -> "gradle_worker"
+            is KotlinDaemon -> "kotlin_daemon"
+            is Unknown -> vm.name
+                .takeLast(64)
+                .removePrefix(".")
```
dsvoronin, line 44: Could it be `.` symbols not only as prefix?

dsvoronin (approved):

subprojects/assemble/build-metrics/src/main/kotlin/com/avito/android/plugin/build_metrics/internal/jvm/JvmMetricsSender.kt
eugene-krivobokov: Here is a corner case with very long class name or jar name: 
`very.long.package.ClassName` -> `very_long_package_ClassName`

I remove dot to avoid useless `_` in front of series name.
Here may be more smart heuristic to extract class name. 
If we see this metric, we probably should support this process explicitly rather than complex heuristics.

subprojects/assemble/build-metrics/src/main/kotlin/com/avito/android/plugin/build_metrics/internal/jvm/Jps.kt
eugene-krivobokov: Yes

subprojects/assemble/build-metrics/src/main/kotlin/com/avito/android/plugin/build_metrics/internal/jvm/Jps.kt
eugene-krivobokov: Done, but I think that this is more like premature flexibility.
If no one is not changing it, it's not needed yet.

subprojects/assemble/build-metrics/src/main/kotlin/com/avito/android/plugin/build_metrics/internal/jvm/Jcmd.kt
eugene-krivobokov: Extracted parsing for jcmd, but mostly for readability. Here are no problems with testing.

Left jps the same - it has only one command by the nature.

subprojects/assemble/build-metrics/src/main/kotlin/com/avito/android/plugin/build_metrics/internal/teamcity/CollectTeamcityMetricsTask.kt
dsvoronin: i see, thanks for the explanation

subprojects/assemble/build-metrics/src/main/kotlin/com/avito/android/plugin/build_metrics/internal/jvm/Jcmd.kt
dsvoronin: yep, i saw test code, but i believe that testing functions that doesn't involve `ProcessRunner` leads to more tests longevity

eugene-krivobokov: I'd like to explain motivation in favor of another approach.

`ProcessRunner` in those tests is a stub. This abstraction has a single and reasonable responsibility. This is why stubbing it is ok.

Extracting only parsing into a new abstraction have led to more complex and anemic design: more abstractions, more connections caused not by responsibility but only tests. 
As for me, litmus test here is responsibility. It gives more reusability and reduces accidental complexity.

I think it's better to extract commands in this case. They still will use ProcessRunner, but they will reflect meaningful and more reusable abstraction.

PR #1360, opened 2022-07-12
MBSA-586 release docs update

dsvoronin:


eugene-krivobokov (approved):

dsvoronin:
I encountered another flaky test here: `RunnerIntegrationTest.tests execution timeout - run succeed - when timeout is not exceeded`.

Fails in very stable manner with Idea junit runner "until failure"

Will try to handle it

PR #1361, opened 2022-07-12
MBSA-670 fix test runner with stuck intention router

dsvoronin:
test was timed out when first intention was send before observation started

### DIFF
--- a/subprojects/test-runner/client/src/test/kotlin/com/avito/runner/scheduler/runner/RunnerIntegrationTest.kt
+++ b/subprojects/test-runner/client/src/test/kotlin/com/avito/runner/scheduler/runner/RunnerIntegrationTest.kt
@@ -579,12 +579,15 @@ internal class RunnerIntegrationTest {
 
         devices.send(device)
         val result = runner.runTests(tests)
-        result.fold({
-            throw AssertionError("Test run finished successfully, but failure was expected")
-        }, {
-            assertThat(it).isInstanceOf(IllegalStateException::class.java)
-            assertThat(it).hasMessageThat().startsWith("Test run finished with timeout")
-        })
+        result.fold(
+            onSuccess = {
+                throw AssertionError("Test run finished successfully, but failure was expected")
+            },
+            onFailure = {
+                assertThat(it).isInstanceOf(IllegalStateException::class.java)
+                assertThat(it).hasMessageThat().startsWith("Test run finished with timeout")
+            }
+        )
 
         assertThat(devicesProvider.isReleased).isTrue()
         state.assertIsCancelled()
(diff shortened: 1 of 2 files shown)


subprojects/test-runner/service/src/main/kotlin/com/avito/runner/service/IntentionsRouter.kt
```diff
@@ -32,6 +39,11 @@ public class IntentionsRouter(
 
         logger.debug("sending intention with id: $intentionId [$intention]")
 
+        while (!isObserving.get()) {
```
dsvoronin: I see it as an ugly solution, but at least it works.

IDEA test runner with `repeat until failure` gave me stable fails on 2..300 first runs, and 3000+ is green after this fix introduced.

i have ~70 logs: `Intention router is not being observed yet, waiting...` on 3000 runs

Do you have an idea why it is stuck that way? And maybe there is a better solution?

subprojects/test-runner/client/src/test/kotlin/com/avito/runner/scheduler/runner/RunnerIntegrationTest.kt
```diff
@@ -600,7 +603,7 @@ internal class RunnerIntegrationTest {
         val runner = provideRunner(
             targets = targets,
             devicesProvider = devicesProvider,
-            executionTimeout = Duration.ofSeconds(5),
+            executionTimeout = Duration.ofSeconds(1),
```
dsvoronin, line 606: Conditions are even tighter :)

subprojects/test-runner/service/src/main/kotlin/com/avito/runner/service/IntentionsRouter.kt
```diff
@@ -6,12 +6,16 @@ import com.avito.runner.service.model.intention.Intention
...
 import kotlinx.coroutines.channels.ReceiveChannel
+import kotlinx.coroutines.delay
+import java.util.concurrent.atomic.AtomicBoolean
 
 public class IntentionsRouter(
     private val intentionRoutings: MutableMap<String, Channel<Intention>> = mutableMapOf(),
     loggerFactory: LoggerFactory
 ) {
 
+    private val isObserving = AtomicBoolean(false)
```
dsvoronin: i have no idea what i'am doing here. do we even need atomic here? Are coroutines here produces actual threads?

**edit**: it is working fine even on simple boolean

subprojects/test-runner/service/src/main/kotlin/com/avito/runner/service/IntentionsRouter.kt
sergioboi: I think it's an expected behavior that sending happens before observing
The key problems there maybe in that `intentionRoutings` is simple mutable map.
Mutable map can't correctly work in multithreaded environment.

You could try to make `send` and `observe` synchronized

dsvoronin: Yes, you're right!
I used Mutex to synchronize access to map for all methods and it works fine that way

eugene-krivobokov (approved):

PR #1364, opened 2022-07-13
MBSA-636 Publish CdBuildConfig from Nupokati plugin

eugene-krivobokov:
Previously, clients used `Project.cdBuildConfig` to get a config. It was from legacy `upload-cd-build-result` module.

Not, the path to config is configured in nupokati plugin. 
Clients still need to read it to configure other tasks (e.g. qapps) and applications (e.g. version name).

### DIFF
--- a/subprojects/delivery/nupokati/src/main/kotlin/com/avito/android/NupokatiPlugin.kt
+++ b/subprojects/delivery/nupokati/src/main/kotlin/com/avito/android/NupokatiPlugin.kt
@@ -25,18 +25,20 @@ import org.gradle.kotlin.dsl.withType
 
 public class NupokatiPlugin : Plugin<Project> {
 
+    public lateinit var cdBuildConfig: Provider<CdBuildConfig>
+
     override fun apply(project: Project) {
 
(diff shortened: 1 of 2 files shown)


dsvoronin (approved):

subprojects/delivery/nupokati/src/main/kotlin/com/avito/android/NupokatiPlugin.kt
```diff
@@ -83,7 +85,7 @@ public class NupokatiPlugin : Plugin<Project> {
                     this.reportViewerUrl.set(extension.reportViewer.frontendUrl)
                     this.reportCoordinates.set(extension.reportViewer.reportCoordinates)
                     this.teamcityBuildUrl.set(extension.teamcityBuildUrl)
-                    this.cdBuildConfig.set(cdBuildConfig)
+                    this.cdBuildConfig.set(this@NupokatiPlugin.cdBuildConfig)
```
eugene-krivobokov, line 88: Again - https://github.com/gradle/gradle/issues/20932

PR #1365, opened 2022-07-13
MBSA-483 fix flaky httpserver test

dsvoronin:
We will discuss this server later, but for now i just fixing the flaky behavior of emcee http server test.
I don't know all of it requirements though, but seems like i didn't ruin it's previous behavior

### DIFF
--- a/subprojects/emcee/worker/src/main/kotlin/com/avito/emcee/worker/internal/rest/HttpServer.kt
+++ b/subprojects/emcee/worker/src/main/kotlin/com/avito/emcee/worker/internal/rest/HttpServer.kt
@@ -5,66 +5,81 @@ import java.io.BufferedReader
 import java.io.InputStreamReader
 import java.io.PrintWriter
 import java.net.ServerSocket
+import java.net.Socket
 
-internal class HttpServer(private val handlers: List<RequestHandler>, private val debug: Boolean) {
+internal class HttpServer(
+    private val handlers: List<RequestHandler>,
+    private val debug: Boolean
+) {
 
-    private var serverThread: Thread? = null
+    private var serverSocket: ServerSocket? = null
 
     @Volatile
-    private var isStopped = false
+    internal var isStarted: Boolean = false
 
     fun start(port: Int) {
-        serverThread = Thread {
-            val ss = ServerSocket(port)
-            if (debug) println("Starting REST server on $port port/")
-
-            ss.use {
-                while (!Thread.interrupted()) {
-                    val connection = ss.accept()
-                    val reader = BufferedReader(InputStreamReader(connection.getInputStream()))
-                    val writer = PrintWriter(connection.getOutputStream())
-
-                    try {
-                        val request = reader.readLine() ?: continue
-
-                        val body = StringBuilder()
-                        while (true) {
-                            val line = reader.readLine()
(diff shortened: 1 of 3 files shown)


subprojects/emcee/worker/src/main/kotlin/com/avito/emcee/worker/internal/rest/HttpServer.kt
```diff
@@ -5,66 +5,81 @@ import java.io.BufferedReader
...
+internal class HttpServer(
+    private val handlers: List<RequestHandler>,
+    private val debug: Boolean
+) {
 
-    private var serverThread: Thread? = null
+    private var serverSocket: ServerSocket? = null
 
-    @Volatile
-    private var isStopped = false
```
dsvoronin, line 14: I expected `isStopped` to be used in while loop instead of `!Thread.interrupted()`, but anyways it wasn't working as expected, because of blocking nature of `ServerSocket.accept()`

subprojects/emcee/worker/src/test/kotlin/com/avito/emcee/worker/rest/HttpServerTest.kt
```diff
@@ -6,47 +6,73 @@ import org.junit.jupiter.api.AfterEach
...
+    private var port: Int? = null
+    private val client = SimpleClient()
+    private val server = createServer()
 
     @BeforeEach
     fun before() {
-        server.start(defaultPort)
-        client.start("127.0.0.1", defaultPort)
+        port = findAvailablePort()
+        Executors.newSingleThreadExecutor().submit { server.start(port!!) }
```
dsvoronin, line 21: Correct me, but it looks like `HttpServer.start` should block the main thread and it's only bothers in tests.
That's my workaround

subprojects/emcee/worker/src/main/kotlin/com/avito/emcee/worker/internal/rest/HttpServer.kt
```diff
@@ -5,66 +5,81 @@ import java.io.BufferedReader
...
+    // used in tests only
+    internal var isStarted: Boolean = false
```
eugene-krivobokov, line 18: volatile?

subprojects/emcee/worker/src/test/kotlin/com/avito/emcee/worker/rest/HttpServerTest.kt
```diff
@@ -6,47 +6,73 @@ import org.junit.jupiter.api.AfterEach
...
     @AfterEach
     fun after() {
-        server.stop()
         client.stop()
+        server.stop()
     }
+
+    private fun waitServerStart() {
+        while (!server.isStarted) {
+            Thread.sleep(50)
```
eugene-krivobokov:
```suggestion
            Thread.onSpinWait()
```
?

eugene-krivobokov (approved):

subprojects/emcee/worker/src/test/kotlin/com/avito/emcee/worker/rest/HttpServerTest.kt
dsvoronin: Thanks! didn't know about that method

PR #1374, opened 2022-07-20
MBSA-234 build with java 17; upgrade to debian bullseye

dsvoronin:


eugene-krivobokov (approved):

subprojects/test-runner/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/internal/FileTypeAdapter.kt
```diff
@@ -0,0 +1,24 @@
...
+ * Manually serialize to avoid exception on later java runtimes
+ *
+ * Caused by: java.lang.reflect.InaccessibleObjectException:
+ * Unable to make field private final java.lang.String java.io.File.path accessible:
+ * module java.base does not "opens java.io" to unnamed module...
+ */
+internal class FileTypeAdapter : TypeAdapter<File>() {
+
+    override fun write(writer: JsonWriter, file: File) {
+        writer.value(file.absolutePath)
```
eugene-krivobokov:
```suggestion
        writer.value(file.cannonicalPath)
```
It would be a bit more stable, less relative.

subprojects/test-runner/instrumentation-tests/src/main/kotlin/com/avito/instrumentation/internal/DurationTypeAdapter.kt
```diff
@@ -0,0 +1,17 @@
...
+
+import com.google.gson.TypeAdapter
+import com.google.gson.stream.JsonReader
+import com.google.gson.stream.JsonWriter
+import java.time.Duration
+
+internal class DurationTypeAdapter : TypeAdapter<Duration>() {
+
+    override fun write(writer: JsonWriter, duration: Duration) {
+        writer.value(duration.toString())
```
eugene-krivobokov: Storing as a number would allow to avoid parsing ISO format on deserialization.

eugene-krivobokov (approved):

dsvoronin:
Merging without using actual java 17, only preparations, because we need to publish this first (problematic code used in instrumentation app)

PR #1376, opened 2022-07-26
MBSA-650 Worker returns a stub result for tests execution

RuslanMingaliev:
## Changes

- API models modified to match contract and avoid crashes while parsing
- Added new abstraction to store currently processing buckets
- Added new implementation of `TestJobConsumer` that returns a stub result to the `Queue` after a short delay.


subprojects/emcee/queue-api-models/src/main/kotlin/com/avito/emcee/queue/Bucket.kt
```diff
@@ -6,4 +6,6 @@ import com.squareup.moshi.JsonClass
 public data class Bucket(
     val bucketId: String,
     val payloadContainer: PayloadContainer,
+    val analyticsConfiguration: Any,
```
eugene-krivobokov, line 9: Why can't it be typed?

subprojects/emcee/queue-worker-api/src/main/kotlin/com/avito/emcee/worker/SendBucketResultBody.kt
```diff
@@ -21,13 +21,14 @@ public data class SendBucketResultBody(
...
             val testEntry: TestEntry,
-            val testRunResult: List<TestRunResult>
+            val testRunResults: List<TestRunResult>
         ) {
             @JsonClass(generateAdapter = true)
             public data class TestRunResult(
                 /**
                  * in sec
                  */
+                val udid: String,
```
eugene-krivobokov, line 28: user device identifier? Is this term from iOS/macOs world?

subprojects/emcee/worker/src/main/kotlin/com/avito/emcee/worker/internal/storage/SingleElementProcessingBucketsStorage.kt
```diff
@@ -0,0 +1,22 @@
...
+    override fun add(bucket: Bucket) {
+        require(this.bucket == null) { "There is a bucket with id=${bucket.bucketId} already" }
+        this.bucket = bucket
+    }
+
+    override fun remove(bucket: Bucket) {
+        require(this.bucket == bucket) { "There is no bucket with id=${bucket.bucketId}" }
+        this.bucket = null
+    }
+
```
eugene-krivobokov, line 18: Shouldn't it be thread safe?

eugene-krivobokov (approved):

subprojects/emcee/queue-api-models/src/main/kotlin/com/avito/emcee/queue/Bucket.kt
RuslanMingaliev: We don't use this field yet, but Moshi requires to declare it. It will be typed later.

subprojects/emcee/queue-worker-api/src/main/kotlin/com/avito/emcee/worker/SendBucketResultBody.kt
RuslanMingaliev: `unique device id`. This term is from Emcee: 
https://github.com/avito-tech/Emcee/blob/master/Sources/CommonTestModels/UDID.swift

It is used to distinguish the simulators or real devices each other.

subprojects/emcee/worker/src/main/kotlin/com/avito/emcee/worker/internal/storage/SingleElementProcessingBucketsStorage.kt
RuslanMingaliev: Definitely yes, it should be. I'll make it thread-safe in the next PR.

PR #1389, opened 2022-08-16
MBSA-486 Run fake queue in k8s

RuslanMingaliev:


subprojects/emcee/queue-fake-server/src/main/kotlin/com/avito/emcee/server/fake/Main.kt
```diff
@@ -11,15 +11,19 @@ import io.ktor.server.engine.embeddedServer
 import io.ktor.server.netty.Netty
 import io.ktor.server.plugins.contentnegotiation.ContentNegotiation
 import io.ktor.server.response.respond
+import io.ktor.server.routing.get
 import io.ktor.server.routing.post
 import io.ktor.server.routing.routing
 
 public fun main() {
-    embeddedServer(Netty, port = 8080) {
+    embeddedServer(Netty, port = 41000) {
```
RuslanMingaliev, line 19: It doesn't matter, but the real queue runs on the 41000 port

subprojects/emcee/k8s/emcee.yaml
```diff
@@ -0,0 +1,44 @@
...
+    replicas: 1
+    template:
+        metadata:
+            labels:
+                app: emcee
+                service: emcee
+        spec:
+            containers:
+                -   name: emcee
+                    image: emcee-queue
```
sergioboi, line 24: What is the image? How did you build it?

subprojects/emcee/k8s/emcee.yaml
```diff
@@ -0,0 +1,44 @@
...
+                    imagePullPolicy: IfNotPresent
+                    ports:
+                        -   containerPort: 41000
```
sergioboi, line 27: Why do we need that?

subprojects/emcee/queue-fake-server/build.gradle.kts
```diff
@@ -1,14 +1,27 @@
...
     mainClass.set("com.avito.emcee.server.fake.MainKt")
 }
 
+val registry = project.getMandatoryStringProperty("avito.docker.registry", allowBlank = false)
+
+docker {
+    javaApplication {
+        baseImage.set("$registry/android/openjdk:11")
+        ports.set(setOf(41000))
+        images.set(setOf("emcee-queue:latest"))
```
sergioboi, line 20: I'm against of using `latest` but in current place It doesn't matter

subprojects/emcee/k8s/emcee.yaml
```diff
@@ -0,0 +1,44 @@
+---
```
sergioboi, line 1: I don't remember all details but could you add what we try to do via this yaml file
Could yaml have a comment?
As I see we create a `deployment` related to `service`. But do we need a `service`? Or we could do the same without `service` only with `deployment`?

sergioboi (approved):

subprojects/emcee/k8s/emcee.yaml
```diff
@@ -0,0 +1,44 @@
...
+        name: emcee
+        service: emcee
+spec:
+    type: NodePort
+    selector:
+        app: emcee
+    ports:
+        -   protocol: TCP
+            port: 41000
+            targetPort: 41000
```
beefon, line 44: Emcee runs by default on default range of ports `[41000...41009]`, if it is important here.

subprojects/emcee/k8s/emcee.yaml
RuslanMingaliev: Add comments

subprojects/emcee/k8s/emcee.yaml
```diff
@@ -0,0 +1,44 @@
+---
+kind: Deployment
+apiVersion: apps/v1
+metadata:
+    name: emcee
+    namespace: default
+    labels:
```
RuslanMingaliev, line 7: Do we need these labels?

PR #1402, opened 2022-08-24
MBS-13487 Add swift-builder Docker image

RuslanMingaliev:


ci/docker/swift-builder/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM swift:5.6.2
```
sergioboi, line 1: I'm confused why we need this image at all

RuslanMingaliev: Emcee cannot be built without `libssl-dev` package which does not exist in `swift` image. In time there my be more dependencies like `libssl-dev`. For the same reason we use our own `android-builder` image except of official https://hub.docker.com/_/gradle or smth else.

PR #1407, opened 2022-09-07
MBSA-719 Check memory files existing to avoid errors

sergioboi:


subprojects/assemble/build-metrics/src/main/kotlin/com/avito/android/plugin/build_metrics/internal/runtime/Cgroup2.kt
```diff
@@ -40,12 +40,18 @@ internal sealed class Cgroup2 {
             val cgroupDir = File("/sys/fs/cgroup")
 
             return if (cgroupDir.exists()) {
-                Available(
-                    memory = Memory(
-                        memoryMax = File(cgroupDir, "memory.max"),
-                        memoryCurrent = File(cgroupDir, "memory.current")
+                val memoryMax = File(cgroupDir, "memory.max")
+                val memoryCurrent = File(cgroupDir, "memory.current")
+                if (memoryMax.exists() && memoryCurrent.exists()) {
```
sergioboi, line 45: As I see now we never find this files at CI build.
I check locally inside docker container that files exists.
So I have no idea why we can't find them at CI builds

sergioboi: I have only one idea is try to log files inside directory `/sys/fs/cgroup` at CI to see what are there

RuslanMingaliev (approved):

PR #1415, opened 2022-09-30
MBSA-758 use Ktor for worker's HttpServer

RuslanMingaliev:


subprojects/emcee/queue-api-models/src/main/kotlin/com/avito/emcee/queue/Bucket.kt
```diff
@@ -8,4 +8,6 @@ public data class Bucket(
     val payloadContainer: PayloadContainer,
     val analyticsConfiguration: Any,
     val workerCapabilityRequirements: List<String>
-)
+) {
+    public companion object
```
sergioboi, line 12: Tell me could we don't use that hack?

subprojects/emcee/worker/build.gradle.kts
```diff
@@ -12,11 +12,14 @@ dependencies {
     implementation(libs.okhttp)
     implementation(libs.okhttpLogging)
     implementation(libs.moshi)
+    implementation(libs.bundles.ktor)
     implementation("org.jetbrains.kotlinx:kotlinx-coroutines-core:1.5.2")
```
sergioboi, line 16: Do you plan to move versions to toml file?

subprojects/emcee/queue-worker-api/src/main/kotlin/com/avito/emcee/worker/RegisterWorkerResponse.kt
```diff
@@ -16,5 +16,6 @@ public data class RegisterWorkerResponse(
         val portRange: Any,
         val payloadSignature: String,
         val maximumCacheTTL: Long,
+        val trackSigkillIgnorance: Boolean
```
sergioboi, line 19: I prefer to see KDoc on this property.
Its' semantics is not clear

subprojects/emcee/worker/src/main/kotlin/com/avito/emcee/worker/internal/rest/HttpServer.kt
```diff
@@ -1,93 +1,55 @@
...
-                if (debug) println("<-- $request\n$body")
-
-                val handler = findHandler(request)
-                val response = if (handler != null) {
-                    handler.response().wrapToSuccessfulResponse()
-                } else {
-                    badRequest()
+        server = embeddedServer(Netty, port = port) {
+            install(ContentNegotiation) {
+                gson() // TODO: use Moshi instead when it will be available
```
sergioboi, line 26: Maybe let's migrato to KotlinX serialization if we use Ktor?

subprojects/emcee/worker/src/main/kotlin/com/avito/emcee/worker/internal/rest/HttpServer.kt
```diff
@@ -1,93 +1,55 @@
...
+            }
+            routing {
+                handlers.forEach { handler: RequestHandler<*> ->
+                    route(handler.path, handler.method) {
+                        handle {
+                            call.respond(handler.response())
```
sergioboi, line 32: Could we describe routes explicitly there without RequestHandler abstraction?

subprojects/emcee/worker/src/test/kotlin/com/avito/emcee/worker/rest/stub/StubBucket.kt
```diff
@@ -0,0 +1,42 @@
...
+import com.avito.emcee.queue.ApkLocation
+import com.avito.emcee.queue.Bucket
+import com.avito.emcee.queue.BuildArtifacts
+import com.avito.emcee.queue.Payload
+import com.avito.emcee.queue.PayloadContainer
+import com.avito.emcee.queue.RemoteApk
+import com.avito.emcee.queue.TestConfiguration
+import com.avito.emcee.queue.TestExecutionBehavior
+
+fun Bucket.Companion.stub(bucketId: String) = Bucket(
```
sergioboi, line 12: Personally Prefer to see fun `StubBucket`

subprojects/emcee/queue-api-models/src/main/kotlin/com/avito/emcee/queue/Bucket.kt
RuslanMingaliev: We use it throughout the project. We can use factory functions or something else, but I don't see anything wrong with exposing the companion objects.

subprojects/emcee/worker/build.gradle.kts
RuslanMingaliev: There is an old version already. But it has compatibility problems with current Kotlin version. We can discuss it at the next sync.

subprojects/emcee/queue-worker-api/src/main/kotlin/com/avito/emcee/worker/RegisterWorkerResponse.kt
RuslanMingaliev: It's not clear to me too. We don't use this property now, but can't skip it's declaration here, because we use `failOnUnknown` setting on Moshi configuration.
I'll plan to turn off this later and clean up on the models in a separate task later

subprojects/emcee/worker/src/main/kotlin/com/avito/emcee/worker/internal/rest/HttpServer.kt
RuslanMingaliev: Yes, we can. But I'd suggest to think about it after creating MVP

subprojects/emcee/worker/src/main/kotlin/com/avito/emcee/worker/internal/rest/HttpServer.kt
RuslanMingaliev: Yes, we can. But then we'd have to provide all necessary dependencies to the `HttpServer` directly. It brakes the dependency inversion principle and increases the coupling degree.

subprojects/emcee/worker/src/test/kotlin/com/avito/emcee/worker/rest/stub/StubBucket.kt
RuslanMingaliev: Answered about it above

subprojects/emcee/queue-api-models/src/main/kotlin/com/avito/emcee/queue/Bucket.kt
sergioboi: What if we make it `internal`

subprojects/emcee/worker/build.gradle.kts
sergioboi: I think it's not a problem to have it in toml file

subprojects/emcee/queue-worker-api/src/main/kotlin/com/avito/emcee/worker/RegisterWorkerResponse.kt
sergioboi: Could you create a task and add it number here

subprojects/emcee/worker/src/main/kotlin/com/avito/emcee/worker/internal/rest/HttpServer.kt
sergioboi: I consider it only because of comment. I think the comment could became outdated

subprojects/emcee/worker/src/main/kotlin/com/avito/emcee/worker/internal/rest/HttpServer.kt
sergioboi: What do you think about declaring multiple methods in Retrofit service? Does that break DI principle?
As for me I want to see REST API of one service in one place

sergioboi: We could extract routing describing in separate file

PR #1418, opened 2022-10-05
MBSA-724 Extract di logic to DI

sergioboi:


subprojects/emcee/worker/src/main/kotlin/com/avito/emcee/worker/internal/rest/HttpServer.kt
```diff
@@ -14,12 +14,15 @@ import io.ktor.server.routing.routing
 
 internal class HttpServer(
     private val handlers: List<RequestHandler<*>>,
-    private val debug: Boolean
+    private val debug: Boolean,
+    private val port: Int,
+    private val gracePeriodMs: Long = 500,
```
RuslanMingaliev, line 19: we don't need that

subprojects/emcee/worker/src/test/kotlin/com/avito/emcee/worker/rest/HttpServerTest.kt
```diff
@@ -20,15 +20,24 @@ import java.net.ServerSocket
...
-    private var port: Int? = null
+    private companion object {
+        private fun findAvailablePort(): Int {
+            val s = ServerSocket(0)
+            val port = s.localPort
+            s.close()
+            return port
+        }
+
+        private val port: Int = findAvailablePort()
```
RuslanMingaliev: It should be found on each call, for each test. Not only once.

subprojects/emcee/worker/src/test/kotlin/com/avito/emcee/worker/rest/HttpServerTest.kt
```diff
@@ -56,15 +65,11 @@ class HttpServerTest {
         server.stop()
     }
 
-    private fun findAvailablePort(): Int {
-        val s = ServerSocket(0)
-        val port = s.localPort
-        s.close()
-        return port
+    private fun createServer(): HttpServer {
+        return HttpServer(
```
RuslanMingaliev, line 58: it looks more pretty with builder pattern for me

subprojects/emcee/worker/src/main/kotlin/com/avito/emcee/worker/internal/di/WorkerDI.kt
```diff
@@ -0,0 +1,68 @@
...
+import com.avito.emcee.worker.internal.rest.handler.ProcessingBucketsRequestHandler
+import com.avito.emcee.worker.internal.storage.ProcessingBucketsStorage
+import com.avito.emcee.worker.internal.storage.SingleElementProcessingBucketsStorage
+import okhttp3.OkHttpClient
+import okhttp3.logging.HttpLoggingInterceptor
+import retrofit2.Retrofit
+import kotlin.time.ExperimentalTime
+
+@ExperimentalTime
+internal class WorkerDI(
```
RuslanMingaliev, line 24: dagger?

subprojects/emcee/worker/src/main/kotlin/com/avito/emcee/worker/internal/di/WorkerDI.kt
```diff
@@ -0,0 +1,68 @@
...
+    private val config: Config,
+    private val debugMode: Boolean
+) {
+
+    private val workerIdProvider: WorkerIdProvider = HostnameWorkerIdProvider()
+    private val okHttpClient = OkHttpClient.Builder().apply {
+        if (debugMode) {
+            addInterceptor(HttpLoggingInterceptor { message ->
+                println(message)
```
RuslanMingaliev: Logger should be used here too

RuslanMingaliev (approved):

subprojects/emcee/worker/src/main/kotlin/com/avito/emcee/worker/internal/rest/HttpServer.kt
sergioboi: I noticed that after failure our worker works for minute more. So I decide to use this properties explicitly
But problem was not in stopping HttpServer

subprojects/emcee/worker/src/test/kotlin/com/avito/emcee/worker/rest/HttpServerTest.kt
sergioboi: Before in test was @BeforeAll annotation I think that means that we make it one time before tests
What the point each time find availablePorts

subprojects/emcee/worker/src/test/kotlin/com/avito/emcee/worker/rest/HttpServerTest.kt
sergioboi: I against the builder in current case. In my opinion builder needs when you have many parameters and many configurations of object and you make them dynamically in runtime. It's not about current case

subprojects/emcee/worker/src/main/kotlin/com/avito/emcee/worker/internal/di/WorkerDI.kt
sergioboi: Let's stay without dagger. Maybe in the future

PR #1421, opened 2022-10-11
MBSA-764 Add compatibility with targetSdk 31

sergioboi:


subprojects/gradle/deeplink-generator/src/main/kotlin/com/avito/deeplink_generator/internal/merge/MergeDeeplinkManifestTask.kt
```diff
@@ -131,6 +131,7 @@ internal abstract class MergeDeeplinkManifestTask : DefaultTask() {
             
                 <application>
                     <activity
+                        android:exported="true"
```
sergioboi, line 134: the change

Issue #1425, opened 2023-01-17
TypeText.writeText is crashing on Android 13


idyatlov:
**Describe the bug**
Due to the recent [renaming](https://android.googlesource.com/platform/frameworks/base/+/68d63ef153a69376afb2bd6ceb7b57a20fec9c83) of the `mIInputContext` field to `mFallbackInputConnection`, everything crashes when trying to write text on Android 13.

**How to reproduce**
Launch any test with `TextFieldElement.write(text)` on Android 13

**Expected behavior**
Everything should work as expected

**Screenshots**
No need

**Environment:**
Android 13 and later

**Starcktrace**
`Caused by: java.lang.NoSuchFieldException: No field mIInputContext in class Landroid/view/inputmethod/InputMethodManager; (declaration of ‘android.view.inputmethod.InputMethodManager’ appears in /system/framework/framework.jar!classes3.dex)
at java.lang.Class.getDeclaredField(Native Method)
at com.avito.android.test.util.ReflectionKt.getFieldByReflectionWithAnyField(Reflection.kt:10)
at com.avito.android.test.element.field.actions.TypeText.writeText(TypeText.kt:78)
at com.avito.android.test.element.field.actions.TypeText.perform(TypeText.kt:53)
at com.avito.android.test.interceptor.ActionInterceptor$Proxy.perform(Interceptor.kt:29)
at androidx.test.espresso.ViewInteraction$SingleExecutionViewAction.perform(ViewInteraction.java:2)
at androidx.test.espresso.ViewInteraction.doPerform(ViewInteraction.java:25)
at androidx.test.espresso.ViewInteraction.-$$Nest$mdoPerform(Unknown Source:0)
at androidx.test.espresso.ViewInteraction$1.call(ViewInteraction.java:7)
at androidx.test.espresso.ViewInteraction$1.call(ViewInteraction.java:1)`

PR #1427, opened 2023-06-27
Properly fix text typing on Android 13

idyatlov:
To properly fix issue #1425 we need to go deeper into the wrapper to obtain an input connection.
Also added a small fix to PhoneNumberFormattingTextWatcher - this will allow related test to pass on any device, not just that configured with a US locale.

Tested and confirmed on real Pixel 6a A13 device.


sergioboi:
I made a fix to develop. Today will release a new version

